PR #1, opened 2015-01-05
Encryption implemented

AdamISZ:
Test this before merging it!

Dependency: install libsodium from here: http://doc.libsodium.org/installation/README.html, use the tarball for the 1.0.1 version (latest).

The following elements:
libnacl is a Python binding of libsodium, a C library implementation of NaCl. NaCl includes public key authenticated encryption using Dan Bernstein's ECC curve25519 (https://en.wikipedia.org/wiki/Curve25519), salsa20 stream cipher and poly1305 for MAC/authentication. These seem to be widely respected and used in the cryptographic community (and not designed by NIST!). See http://nacl.cr.yp.to for overview and the [libsodium docs](http://doc.libsodium.org/public-key_cryptography/authenticated_encryption.html) for details of usage (section 8) and installation (section 1).

Private message chunking is implemented with trailer characters ';' and '~' . Current chunk size is hardcoded at 350 characters, well short of IRC limits, obviously should be made configurable.

Encryption keypairs are generated dynamically as necessary and persisted in files with the appropriate user permissions for private keys.

* Changes to the messaging protocol

See the file 'encryption-protocol.txt' for details of the steps of the handshake.

All encrypted messages are exchanged in base64 encoding (as well as the message chunking mentioned above). The encryption and encoding is done *before* chunking when sending.

Each agent (taker/maker) keeps a dict containing the current encryption status with its counterparties so it knows when to send encrypted and when unencrypted.

* TODO

Lots, perhaps firstly making sure that the encryption state is maintained properly at the end of one transaction and possibly starting another or doing other things.

Another: make sure that the btc pubkey used by the taker to sign the encryption pubkey is indeed the one that's used in the coinjoin transaction, on maker side.

This PR has been tested on a few basic transactions using the !%go command as seen in the TestTaker objection in taker.py. But it is far from fully tested or polished at the moment.


AdamISZ:
Found a good introduction to libsodium here: http://labs.opendns.com/2013/03/06/announcing-sodium-a-new-cryptographic-library/

chris-belcher:
From what I quickly looked over, it looks great.

What do you think of doing that restructuring first, where the IRC communication is one option and so other kinds (like electrum servers or a DHT) can also be built in? I feel like this encryption might naturally fit together with that.

AdamISZ:
Altered the description above. The most important part is in the encryption-protocol.txt file. This will probably have to be expanded considerably, to discuss attacks etc.

PR #2, opened 2015-01-17
add reset option to wallet tool

AdamISZ:
An option to put all coins back to a single address. Perhaps useful if you want to re-start testing some aspect (although the "used" flags on addresses won't change of course), or to send coins back to the faucet. Debatable I guess, but since it's only an option, I suppose it doesn't hurt.

### DIFF
--- a/common.py
+++ b/common.py
@@ -32,6 +32,13 @@ def get_addr_vbyte():
 	else:
 		return 0x00
 
+def get_signed_tx(wallet,ins,outs):
+	tx = btc.mktx(ins, outs)
+	for index, utxo in enumerate(ins):
+		addr = wallet.unspent[utxo['output']]['address']
+		tx = btc.sign(tx, index, wallet.get_key_from_addr(addr))
+	return tx
+
 class Wallet(object):
 	def __init__(self, seed, max_mix_depth=2):
 		self.max_mix_depth = max_mix_depth
--- a/common.py
+++ b/common.py
@@ -32,7 +32,7 @@ def get_addr_vbyte():
 	else:
 		return 0x00
 
-def get_signed_tx(wallet,ins,outs):
+def get_signed_tx(wallet, ins, outs):
 	tx = btc.mktx(ins, outs)
 	for index, utxo in enumerate(ins):
 		addr = wallet.unspent[utxo['output']]['address']
--- a/irclib.py
+++ b/irclib.py
@@ -83,6 +83,10 @@ class IRCClient(object):
 		self.cp_pubkeys[nick]=None
 		self.enc_boxes[nick]=None	
 	
+	def end_all_encryption(self):
+		for k,v in self.encrypting.iteritems():
+			if v: self.end_encryption(k)
+			
 	def encrypt_encode(self,msg,nick):
 		if not (nick in self.encrypting.keys()) or not (nick in self.enc_boxes.keys()):
 			raise Exception("Encryption is not switched on.")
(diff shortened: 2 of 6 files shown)


chris-belcher:
Fine but please add a space after commas.

For example
print get_signed_tx(wallet,ins,outs)
becomes
print get_signed_tx(wallet, ins, outs)

AdamISZ:
This PR has become a bit of a grab bag of changes. Apologies for the laziness, but if possible perhaps it can be merged all as one. If not, I'll have to figure it out.

Here is a list of changes from before:

Encryption is now ended on both sides (taker and maker) at the point where the communication for the tx ceases (thus, it doesn't pay attention to whether the tx went through on the network, and presumably we can leave it like this until and unless we change the model so that the two counterparties might continue to have a dialog in cases where a tx did not go through for some reason). See 76ff4be.

Spaces in function names corrected.

I added 3 characters from the encryption key storage filename into the bot IRC nick for disambiguation. Not sure that going the other way (giving the seed as the name of the encryption storage file) is a good idea. Not important at this stage, of course.

In d2c4f55 I added an address notifier thread for the CoinJoinTx object on the taker side so that the utxo list is updated with unconfirm and confirm callbacks, as it is on the maker side. This seemed to be necessary to allow TestTaker to function, although perhaps you intend that these callbacks occur in the "higher level" bot modules? Let me know.

Also a change to yield generator - I got a crash in a case where the on_tx_confirmed() callback referred to the on_tx_unconfirmed() and referred to self.orderlist[0] when the previous call to on_tx_unconfirmed() had already deleted the order so that self.orderlist was empty. So I put a check into the function for this. Let me know if there is some error I made here; it's quite possible (although unlikely that this code is harmful).

chris-belcher:
Pretty good.

Encryption added to both sides as in https://github.com/chris-belcher/joinmarket/commit/76ff4be2e7854863d554a80782eac43f2ef300c4 is good principle. The implement has a bug though in the taker code. It ends the encryption after a !sig message but the maker may send several !sig messages, for example if they sent many utxos so the signatures. Instead encryption should be closed in add_signature() in the CoinJoinTX class. Right where pushtx() is called, around where debug('the entire tx is signed, ready to pushtx()') is. There should be a for loop that ends encryption for every maker.

Using add_addr_notify() in the Taker class in https://github.com/chris-belcher/joinmarket/commit/d2c4f55fdd7eb87fb09521b11f0b5f3b8a401b39 is wrong. If you want to add that functionality in TestTaker, make TestTaker pass a finishcallback function to CoinJoinTX and inside there call add_addr_notify(). You may notice your code will fail unless the Taker object already has a wallet field.

The yield-generator.py bugfix seems good as far as I can see.

AdamISZ:
Thanks for the code checking.

In 33d4d90 , I've done pretty much what you said. First, the encryption to all counterparties is ended on the taker side in the CoinJoinTx object exactly when the transaction is pushed. Second the addr_notify function is removed from the CoinJoinTx object and a finish_callback is added to the TestTaker that removes the old and adds the new utxos.

It occurs to me that the encryption model (where takers end encryption with all counterparties, while maker ends encryption with taker (1 counterparty) might be somehow confused in cases of hybrid bots inheriting from maker and taker. It also occurs to me that there may be some exception cases where a tx fails in some way but the encryption state is not switched off. This can be addressed in further work.

chris-belcher:
Regarding the hybrid bots, i doubt even hybrid bots will be acting as makers and takers at the same time with the same counterparties.

Issue #10, opened 2015-01-30
Messaging protocol


AdamISZ:
Opened to start a discussion.

Here is a first attempt.

(redacted: moved to https://github.com/chris-belcher/joinmarket/wiki/Joinmarket-Messaging-Protocol)

AdamISZ:
Some immediate extra comments: 

* Add extra documentation of the exact message format of each type (fill, auth etc.)
* I agree we have a potentially very important issue with message length leaking info. Needs careful thought.
* Not sure about binary vs ascii format encoding. If we could stick to one encoding at JoinMarket's own messaging layer that would be nice. Then any messaging implementation module (IRC, p2p etc) could convert as required. 

Obviously very much a first draft. I was "inspired" by various thoughts that came up when refactoring the encryption.

haarts:
So what do the fields mean in this `relorder`?
```
!relorder 0 1594202 18352866 100 6.9e-05 ~
```

chris-belcher:
The subreddit, irc channel or bitcointalk thread are better places for questions like that.

The fields are 'oid', 'minsize', 'maxsize', 'txfee', 'cjfee'. The bitcoin amounts are in satoshi, oid is order id, cjfee is the coinjoin fee as a string

AdamISZ:
I created [this](https://github.com/chris-belcher/joinmarket/wiki/Joinmarket-Messaging-Protocol) which is a slightly improved (I hope) version of the above.

AdamISZ:
Docs moved long ago to [this repo](https://github.com/JoinMarket-Org/JoinMarket-Docs), although as it happens right now they badly need updating. Messaging protocol changes for 0.2 in [this gist](https://gist.github.com/AdamISZ/baf93ce2589854a7992383b3c69fae13).

PR #12, opened 2015-01-31
Encryption - making it a property of the CoinJoinTx and CoinJoinOrder objects

AdamISZ:
This makes the encryption between counterparties completely 'ephemeral' in that the encrypted channel is negotiated on a per transaction basis by making the crypto Box objects be members of the CoinJoinTx and CoinJoinOrder classes.

The messages are now following the message format protocol as described in #10 . In particular, the decision whether to pass/read a message encrypted or unencrypted is now a function of the command (tx, sig, auth etc) rather than a persistent state. 

All message fields in the message after the command field are encrypted as one chunk and then parsed after decryption.

Apologies for the slight mess of commits in this PR, but it's a function of having to do a manual merge and then having the bug crop up during that process (and so having to do it again).

Tested OK with sendpayment and yield-gen; tests ongoing.


AdamISZ:
I forgot to add the obvious - why this is needed is to that the previous model (having encryption state per counterparty stored within a specific agent/bot/maker/taker) could lead to inconsistent state when there is an unexpected interruption in the flow of a transaction.

chris-belcher:
I agree with the principle of having encryption state be decided by the specific command, instead of storing an internal boolean deciding whether to encrypt or not.

It looks like the encryption code was removed from irclib.py, but there are still fields like IRCClient.encrypting which don't seem to do anything, so should be removed.

Does the chunking in IRC still work? I couldn't understand from quickly skimming.

I imagined we would eventually abstract away the specific IRC format and have maker/taker only talk to an abstract class, like this one https://github.com/chris-belcher/joinmarket/blob/msgchan/message_channel.py
Having the encryption code inside maker and taker seems like it would be incompatible with this goal of abstracting away. This makes me think the encryption code should live in irclib.py instead, so say privmsg() detects which command is being sent and uses that to decide whether to encrypt or not, it also handles chunking inside. Then later we just change maker.privmsg() and taker.privmsg() to MessageClient.fill_order() .send_auth() .send_pubkey() or whatever it might be.

Also this way it looks like the send_priv() function is repeated in maker and taker.

chris-belcher:
Since we'll soon be heavily changing irclib.py, you could change privmsg() and pubmsg() to take as arguments the command seperately.
privmsg(self, nick, cmd, parameters):
send('!' + cmd + ' ' + parameters)

AdamISZ:
@chris-belcher 
> It looks like the encryption code was removed from irclib.py, but there are still fields like IRCClient.encrypting which don't seem to do anything, so should be removed.

Yes, sorry, mistake, fixed.

> Does the chunking in IRC still work? I couldn't understand from quickly skimming.

Yes it's working still, unaffected.

> I imagined we would eventually abstract away the specific IRC format and have maker/taker only talk to an abstract class, like this one https://github.com/chris-belcher/joinmarket/blob/msgchan/message_channel.py
Having the encryption code inside maker and taker seems like it would be incompatible with this goal of abstracting away. This makes me think the encryption code should live in irclib.py instead,

Yes, as we discussed on IRC today, that sounds  right, so I'll try to make that fix now; but I also think that until the base structure of the application is more settled, I'm kind of trying to hit a moving target with this.

>Also this way it looks like the send_priv() function is repeated in maker and taker.
Yes I did notice this myself, but it was a conscious decision as there was a slight difference (Order object only talks to one taker, but Tx object talks to many makers - see how the crypto_box objects are referred to).

AdamISZ:
7a0f00b means the encryption functions are back in irclib (i.e. no encrypt/decrypt parsing in the Maker, Taker classes). The encryption boxes are a property of the Order and Tx objects still, the decision as to whether a particular message is encrypted or not is controlled by the encrypted_commands and plaintext_commands lists in common.py. Hopefully this is a step forward.

chris-belcher:
Thanks, it looks great. I will look at it in more detail shortly and test.

Issue #37, opened 2015-02-03
Bug in wallet generation on startup


AdamISZ:
When the wallet synchronisation runs at startup, it picks up utxos for the wallet which are in existing *unconfirmed* transactions, which can of course lead to pushtx() failure.

I saw this happen for 1bb51f2af3e0a05c37151f337c4a58a8294cc7f78af4cdc06d645530cfd0b98b and aaaf6f710e73740f1e413111b338d4b53a8209d8def2010d09c4adda50bc9150.
aaa was still unconfirmed when I started another sendpayment and outpoint 1 from 1bb was recorded as part of the wallet, so the pushtx() for the new tx failed.

It seems that http://tbtc.blockr.io/api/v1/address/unspent/(address) does *not* show transactions in the list with zero confirmations (I just tried briefly). If it did, I guess it would explain why that utxo showed up in the wallet.

AdamISZ:
So, because I can't see why this should happen, here's a pastebin of the sendpayment run concerned: http://pastebin.com/inu8zKGJ

AdamISZ:
OK, I get it now. The 1bb ..:1 output is counted as unspent *because* the api address/ output does not show unconfirmed txes in the list. Sorry, I got confused.

So I remember Electrum reported unconfirmeds with some flag; not sure how to do it for blockr or bci.

AdamISZ:
Yes, double checked with another example. Once a transaction is on the network unconfirmed, you can check its inputs - those utxos are still shown in the unspent list at address/unspent. This is .. somewhat unhelpful.

AdamISZ:
I guess using /tx/info as in the example: 

http://btc.blockr.io/api/v1/tx/info/60c1f1a3160042152114e2bba45600a5045711c3a8a458016248acec59653471,4addbc5ec75e087a44a34e8c1c3bd05fd495771072879a89a8c9aaa356255cb2 

would give us what we want, as there's a "is_unconfirmed" field.

PR #38, opened 2015-02-04
dont use utxos in existing unconfirmed txs

AdamISZ:
Using the parameter `unconfirmed = 1` stops the output of /address/unspent from containing utxos which are already in a recently pushed (unconfirmed) transaction, so preventing them from appearing in the wallet and causing pushtx() failures. 

Tested briefly (and checked the API output on the web). Seems to work correctly and allow a second  sendpayment to operate while the first is not yet confirmed.

### DIFF
--- a/common.py
+++ b/common.py
@@ -229,13 +229,15 @@ class Wallet(object):
 				blockr_url = 'http://tbtc.blockr.io/api/v1/address/unspent/'
 			elif network == 'btc':
 				blockr_url = 'http://btc.blockr.io/api/v1/address/unspent/'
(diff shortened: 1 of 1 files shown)


AdamISZ:
Important extra note: doing it this way, the output json includes utxos with zero confirmations; these can be spent, but if we don't want to allow that (probably we don't), we should add a line to filter out the entries with `confirmations=0` in the dict.

AdamISZ:
Implemented "extra note" above: will not include utxos into the wallet in find_unspent_addresses that are currently with 0 confirmations.

This PR addresses #37

PR #40, opened 2015-02-05
complete authentication by checking btc pubkey used is valid, also optim...

AdamISZ:
...ise blockchain query code.

The main point here is that in the CoinJoinOrder code, when we verify_unsigned_tx, we check that at least one of the inputs to the transaction comes from the bitcoin pubkey that was used to sign the encryption pubkey for the counterparty (taker). This check was already implicit on the taker side, so no change needed (cj_pubkey).

Also encapsulated the queries to blockr.io in a function in common.py.

Let me know if anything else is needed or doesn't work.


AdamISZ:
Used regex to find all commas without spaces (I think I got them all) - note the pybitcointools module isn't respecting it, but left that alone.

PR #49, opened 2015-02-21
Regtest

AdamISZ:
Highlights:

To use this functionality, you need to:
* Install Bitcoin Core 0.10 (required for watch only address functions)
* Put the path to bitcoin-cli in your $PATH
* Run the daemon as `bitcoind -regtest -daemon`
* Move the chain forward at least 100 blocks, e.g. `bitcoin-cli -regtest setgenerate true 151`
* Set an appropriate IRC chan for your bots to run in, at the top of `common.py`. They don't mix with testnet bots :)
* Run python regtesttest.py from the joinmarket main directory.

This will run rudimentary tests on two party joins (one yieldgen,  one sendpayment) and three party (two yieldgen), all using your local bitcoin regtest network. You can change the three party to higher numbers in the setUp method of the JoinNParty test class. This is not designed to be command line configurable because we want these tests to be automatable (i.e. tests should be self-contained). We could of course loop through with random values for all the constants, including number of joining parties. Or just play with different values of the constants.

Lowlights!

A rudimentary blockchaininterface class is created, with currently only one implementation: RegTestImp(). A thorny issue was the format of responses to calls like get_balance_at_addr. For now, I took the simplest decision: replicate the exact json format provided by blockr.io. At least it's the case for now that one can run either in regtest mode or in 'real' testnet mode. But this needs a lot of improvement.

Regtest implementation ignores any issue with unconfirmed transactions for now. Blocks are automatically created immediately after all transactions, so nothing is ever unconfirmed. This will definitely need to be changed to give realistic simulations.

Catching errors: The bots are run as subprocesses. This in some way is the most realistic structure for testing, but effort will have to be made to ensure the debug logs are readable; this hasn't been done yet (it should be OK with pipes, but that can sometimes be tricky esp cross platform).

Dealing with start up time: Yieldgen bots take a long time to start up because of wallet sync (e.g. 20 seconds?). For now I have just hardcoded reasonable sleep values, but obviously this is not a long term solution if tests get more and more sophisticated.


AdamISZ:
Updated to print all debug messages to a log file for each bot. This will be a bit tidier once things are put into a config file, but is basically functional for now.

There seems to be some problem with getting exception traces to come out to the logfile; it should already be working, but I'm still investigating why not. Obviously that's pretty important.

AdamISZ:
Exception stacks should now show up in logs (at least the last exception).

AdamISZ:
Forgot to note: this has been tested using yieldgen, sendpayment and patientsendpayment (note that 'tested' here means 'make sure the regtest code doesn't stop the normal mode from functioning'). Only very limited tests of course.

I have since discovered a few mistakes in the blockchaininterface code; they don't affect the functioning of regtesttest.py but they've been addressed in the next batch of code for local testnet functioning.

AdamISZ:
Closing, see #53

PR #53, opened 2015-03-03
Testnetrpc

AdamISZ:
Testing ongoing.
As mentioned in previous (withdrawn) PR there was a problem with squashing the earlier commits together, so apologies for the large number of commits.


AdamISZ:
Tests:
wallet tool OK (blockr, local testnet rpc)
yield generator + sendpayment with blockr: OK (after 38b2010)

AdamISZ:
Tested 4 party join with local testnet rpc OK

chris-belcher:
As discussed, this has a couple of bugs and missing features, but still worth merging now so we can work on them.

PR #62, opened 2015-04-18
added ssl

fivepiece:
Quick and dirty, but it works

### DIFF
--- a/joinmarket.cfg
+++ b/joinmarket.cfg
@@ -5,7 +5,7 @@ network = testnet
 bitcoin_cli_cmd = bitcoin-cli
 
 [MESSAGING]
-host = irc.freenode.net
+host = chat.freenode.net
 channel = joinmarket-pit-test
-port = 6667
+port = 6697
 #More stuff to come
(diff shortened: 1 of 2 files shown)


chris-belcher:
Thanks for the PR.

I agree in principle with the idea. I will have to ask for some changes and explanation.

There should be a way to configure where to use SSL or non-SSL connections. It will be useful in case something is wrong with SSL to go back to the simpler plaintext. e.g. SSL over tor isnt needed. Best is probably to add a new field to joinmarket.cfg.

Line 148 in the new file, self.sslsock.send(line + '\r\n') should be done using sendall() to make sure all the data goes down the socket.

Please explain the new line 316, did you find a bug with the PING/PONG that means it should be changed? As it's nothing to do with SSL it should be left out, since sometimes in github you revert individual changes.

fivepiece:
Will change `send()` to `sendall()`
PING issue - You're right, some fragments from me debugging were left in there. I'll clean those up.

Next I'll see how to read from the .cfg file. If anyone is better versed in python, maybe they can help

chris-belcher:
Good.
Merging.

One attractive feature would be to print out details of SSL, for example the cert fingerprint so an interested user could verify it themselves.
See how this network publishes its fingerprint. https://snoonet.org/ircservers

Issue #63, opened 2015-05-06
Implement socks5 proxy 


chris-belcher:
This code is needed so that Tor or an ssh tunnel can be used.

The configuration should go in the joinmarket.cfg file.

Should be a fairly easy addition for keen contributers.

abitfan:
Unfortunately snoonet doesn't allow tor either:
:ny.snoonet.org NOTICE 1NYAADB2S :*** You're banned! For instructions on how to appeal this ban please visit snoonet.org/banappeal.
ERROR :Closing link: (unknown@xxx.xxx.xxx.xxx) [Z-Lined: Hello WitoldZak,you have a host listed in the Dan.me.uk Torexit database. For more information, visit https://www.dan.me.uk/torcheck with your IP: xxx.xxx.xxx.xxx (ID: L1G627MVX6)]
Connection closed by foreign host.

PR #65, opened 2015-05-07
Implemented socks5 proxy

abitfan:
Socks5 proxy support as per #63

### DIFF
--- a/joinmarket.cfg
+++ b/joinmarket.cfg
@@ -10,4 +10,7 @@ host = irc.snoonet.org
 channel = joinmarket-pit
 port = 6697
 usessl = true
+socks5 = true
+socks5_host = 127.0.0.1
+socks5_port = 1500
 #More stuff to come
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -3,7 +3,7 @@ from common import *
 from message_channel import MessageChannel
 from message_channel import CJPeerError
 
-import socket, threading, time, ssl
+import socket, threading, time, ssl, socks
 import base64, os, urllib2, re
 import enc_wrapper
 
@@ -431,6 +431,8 @@ class IRCMessageChannel(MessageChannel):
 		self.given_nick = given_nick
 		self.nick = given_nick
 		self.serverport = (config.get("MESSAGING","host"), int(config.get("MESSAGING","port")))
+		self.socks5_host = config.get("MESSAGING","socks5_host")
+		self.socks5_port = int(config.get("MESSAGING","socks5_port"))
 		self.channel = get_config_irc_channel()
 		self.userrealname = (username, realname)
 		if password and len(password) == 0:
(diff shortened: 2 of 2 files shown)


abitfan:
This doesn't support .onion sites yet

chris-belcher:
It actually does work with .onions

I just tried it with another IRC onion and it works. The problem is the snoonet onion is down, I confirmed it asking in #snoonet, they do plan to get it back up in time.

Tested it, seems to work. Will merge.

BitpopCoin:
Beautiful work

BitpopCoin:
Tor may cause more disconnects

https://github.com/chris-belcher/joinmarket/issues/34

https://github.com/chris-belcher/joinmarket/issues/21

chris-belcher:
@BippopCoin what do you mean by linking those other issues? Please explain why they're relevant, I don't understand.

BitpopCoin:
I think those issues relate to disconnects and issues that occur after

chris-belcher:
@abitfan please get in touch: belcher@riseup.net or the IRC channel

BitpopCoin:
Can this be added to blockr?

BitpopCoin:
Guys are you sure this works? For me it's working using any port which means it's not really using the proxy.

abitfan:
I get ProxyConnectionError(), as expected, when using a non socks port.
So far I'm connecting successfully with non-tor socks but with tor I still haven't managed to connect and its a pita to test due to servers being down, refusing tor connections or meaningless errors which are hard to hunt down. I will eventually setup a tor irc server to get this finished.
When you are enabling socks you need to disable ssl for it to properly work, as the ssl wrapping is causing some problems.

chris-belcher:
When I was testing I found a working irc onion on reddit.com/r/onions I think, or maybe it was the hidden wiki.

It worked for me. The socks protocol that tor uses is the same as any other socks protocol, no reason it shouldn't work for both.

abitfan:
True. I made the difference because with tor you have to be extra careful to redirect dns requests through tor as well. Initially when I first implemented it debugging this took me the most time.

chris-belcher:
Connecting to an onion proves the DNS redirection works, because regular DNS doesn't understand the .onion TLD

BitpopCoin:
Look at the port, this isn't possible unless I'm an idiot? 

13 10:39:12] starting irc                                         
[2015/05/13 10:39:12] connecting                                           
[2015/05/13 10:39:12] Using socks5 proxy 127.0.0.1:80085                   
[2015/05/13 10:39:14] >>pubmsg !orderbook                                  
[2015/05/13 10:39:14] Connected to IRC and joined channel                  
waiting for all orders to certainly arrive

chris-belcher:
Yes port numbers only go up to 65535 or so.

Thats from here https://github.com/chris-belcher/joinmarket/blob/master/lib/irc.py#L455

Looks like the format string uses %s:%s when it should be %s:%d, easy fix. And doesn't affect the actual proxying.

PR #66, opened 2015-05-07
Separate Log Directories, and active_orders Maintenance

adlai:


chris-belcher:
What is the reasoning behind this PR ? Why must statements be separated by nickname ?

adlai:
That was done for convenience, making it easier to locate the relevant logs when conducting postmortems on crashes.

chris-belcher:
This changes the yield generator spreadsheet statements, not the log file.

To change there where the log file goes, edit this https://github.com/chris-belcher/joinmarket/blob/master/lib/common.py#L54

adlai:
- Separate directories for both statements and logs
- Fixed a minor annoyance where the active_orders dict never had old keys removed, so after leaving a yigen running for a couple days, the debug log contained multiple screenfulls of RandNick=None

chris-belcher:
Statements separated by usernames would mean if a yield generator gets restarted [with the default random_nick()], the operator no longer has a single file but many files where the income is recorded. So it's much harder to open in a spreadsheet and calculate the earned sum, time taken and so on.

I agree with the other changes.

adlai:
- single income statement CSV, in the log directory.
- fix crash when removing nonexistent key

PR #68, opened 2015-05-12
Add an option to choose the cheapest or manually pick orders

anduck:


chris-belcher:
As discussed on IRC, I agree in principle.

Could you combine all those commits into one commit. It's quite easy to do on git, search for it.

Does this work correctly for sweep? It looks like it might choose the most expensive option instead of the cheapest.

anduck:
Done. Also added an option to manually pick orders (only works when not sweeping.)

anduck:
-C (--choose-cheapest) works for both non-sweeping and sweeping sends.

chris-belcher:
With the added pick order option, the code is a bit unclean with all those boolean arguments.

A better way would be to have choose_order() accept a function as a parameter. Then in the sendpayment.py file it checks all the booleans and sets which method of selecting orders from the sorted, filtered list.

The three functions weighted_order_choose(), cheapest_order_choose() and pick_order() could be made to all take the arguments (orders, n, feekey). In the future if someone wants to come up with a 4th way of selecting orders, they just need to write a new function.

anduck:
Good remark, fixed it to be cleaner now.

PR #72, opened 2015-05-14
option gaplimit type set to int

AdamISZ:
option --gap-limit N wasn't working due to gap limit being read as str instead of integer, fixed.


K1773R:
thanks!
@chris-belcher this solved my issues with the lost coins. they were at m/0/0/0/017

chris-belcher:
Hope you didnt have too much of a fright.

Eventually the last used address pointer will be stored in the wallet file as in https://github.com/chris-belcher/joinmarket/issues/58#issuecomment-101251395 so stuff like this can't happen so easily.

Issue #73, opened 2015-05-14
Negative values in order format is allowed


chris-belcher:
<Anduck> !relorder 0 0 -124949 -1 -10 ~
	<Anduck> !relorder -129 0 -124949 -1 -10 ~
	<Anduck> !relorder -1 0 -124949 -1 -10 ~
	<Anduck> !relorder -2 0 -124949 -1 -10 ~
	<Anduck> !relorder -1230213 0 -124949 -1 -10 ~
	<Anduck> !relorder -123021300000000000 0 -124949 -1 -10 ~

That kind of stuff appears in the orderbook. The most basic sanity checking should be added. Also check maxsize >= minsize

Silently drop offers which fail these checks

PR #75, opened 2015-05-15
Sanity check for relorder params

abitfan:
Initial sanity checks as per #73

### DIFF
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -179,9 +179,9 @@ class IRCMessageChannel(MessageChannel):
 				maxsize = chunks[3]
 				txfee = chunks[4]
 				cjfee = chunks[5]
-				if any(t < 0 for t in [oid, minsize, maxsize, txfee, cjfee]):
+				if any(float(t) < 0 for t in [oid, minsize, maxsize, txfee, cjfee]):
 					return
-				if minsize > maxsize:
+				if float(minsize) > float(maxsize):
 					return
 				if self.on_order_seen:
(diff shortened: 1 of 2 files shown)


chris-belcher:
Does anyone remember what was said about negative cj fee ? They'll probably never actually be made but it seems anti-competitive to have a price floor.

Did you test this? oid can be zero (as more are now) as well as negative or positive. It's just a marker for orders because a single maker can have many orders.

minsize, txfee and cjfee can be zero.

minsize could be equal to maxsize

This could should probably be in the OrderbookWatch class https://github.com/chris-belcher/joinmarket/blob/master/lib/taker.py#L201

Those fields are all strings by the way, so integer comparison wont work

abitfan:
It only skips the order if any value is less than 0, they can be zero and it will pass
minsize is checked if its greater than maxsize, so they can be equal
Missed the fact about strings, should be fixed now.

I think we need to do sanity checks before passing the args further, so this seems a good place to add them

chris-belcher:
Ah yes you're right about the zero.

If you add the checks here, they'll have to be copypasted into any new MessageChannel class that gets made. Which is repeating code, the understanding of orders belongs in the OrderbookWatch class not the IRCMessageChannel.

btw, using float() would cause a crash if the string is not parsable as a float. All the other fields are actually integers, cjfee is a float only for relorder, it is an integer for absorder.

abitfan:
Added checks as proposed. There is no restriction on cjfee implemented

PR #77, opened 2015-05-16
Rename joinmarket.cfg to joinmarket.cfg.sample

BitpopCoin:
Don't overwrite my config when updating from git


chris-belcher:
Disagree, that would make new users quite confused.

Why don't you do what bitcoin core does. Have no config file in the repo, when joinmarket is run and doesnt detect a config file have it create the default one. Also add joinmarket.cfg to .gitignore

New users get a default config file without any attention, old users dont get their modified config file overwritten.

K1773R:
a sample is still useful to list what options can be set.

chris-belcher:
The generated config file will do that job too, without clutter.

Also the wiki is probably better for documentation.

PR #81, opened 2015-05-19
Use itertools.combinations when sweeping

adlai:
```
>>> create_combination(['apple', 'orange', 'pear', 'orange'], 3)
  File "<stdin>", line 1, in <module>
  File "lib/common.py", line 358, in create_combination
    assert len(result) == nCk(len(li), n)
AssertionError
```

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -7,6 +7,7 @@ import numpy as np
 import blockchaininterface
 from ConfigParser import SafeConfigParser
 import os
+import itertools
 
 JM_VERSION = 1
 nickname = ''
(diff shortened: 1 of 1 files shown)


chris-belcher:
Did this cause a crash in a real life usage?

That function is due to be deleted anyway https://github.com/chris-belcher/joinmarket/issues/48

    >>> import itertools
    >>> [i for i in itertools.combinations(['apple', 'orange', 'pear', 'orange'], 3)]
    [('apple', 'orange', 'pear'), ('apple', 'orange', 'orange'), ('apple', 'pear', 'orange'), ('orange', 'pear', 'orange'

chris-belcher:
This doesn't actually work out the mathematical correct number for a list with repeated elements. If you're going to use less-than instead of equals you might as well remove the assert totally.

Issue #88, opened 2015-05-25
Maker required to know private key for coinjoin address


chris-belcher:
To authenticate the end-to-end encryption, the maker send the pubkey of the coinjoin address.

That's bad because the maker has to know the private key, so patientsendpayment.py cannot do that. It also means the coinjoin address cant be a p2sh address.

Maker should use one of the inputs, same as the taker does.

This is a protocol change so would be incompatible with the older bots.

AdamISZ:
Thought about it, and ACK.
Looking at the code now.

AdamISZ:
Working through it, I see the need to change the !ioauth: message from 

    !ioauth utxo coinjoin_pubkey change_addr btc_sig

to

    !ioauth utxo input_pubkey coinjoin_addr change_addr btc_sig

since the input_pubkey cannot be recovered otherwise. However, the effect will only be to add about 32 bytes to the message (note: addresses take about 40ish bytes over the wire b64encoded, whereas pubkeys take more like 90), which seems fine I guess.

AdamISZ:
Addressed by 0.2.0 release.

Issue #83, opened 2015-05-23
Sending from multisig / p2sh


chris-belcher:
Sending from a p2sh/multisig address is something I've been planning. Essentially that would work by a bot creating the coinjoin, signed by all the makers, and then giving to the user the unsigned transaction who then signs it and broadcasts.

Note that market makers do not have an obligation to not double-spend their inputs, they will remove their inputs only when they see them broadcast on the network. So the multisig user will have to be fairly quick.

This could be fairly easily added to the ob-watcher.py bot (which should be renamed back to gui-taker.py maybe) Although a pure command line version would be useful for scripting.

chris-belcher:
Problem is, the encryption between maker and taker assumes p2pubkey because it uses an EC signature to authenticate. The pubkey has to be in the actual transaction. If the inputs are only p2sh this can't be done anymore.

chris-belcher:
Same as https://github.com/chris-belcher/joinmarket/issues/54

Issue #29, opened 2015-02-01
Minimum other maker count


chris-belcher:
A field specifying the minimum number of other makers should be added to the order format. It will be used by patientsendpayment type bots. Yield-generator bots would of course set the field to zero.

adlai:
This issue can be closed, 80b9f9a adds [--minmakercount](https://github.com/chris-belcher/joinmarket/blob/master/tumbler.py#L276-L277) option.

chris-belcher:
That's just for the taker. The maker still has no way of controlling how many other makers must take part in its coinjoin.

PR #90, opened 2015-05-25
use maker input pubkey instead of coinjoin pubkey for authorization

AdamISZ:
Elementary testnet testing (1 maker) passed OK.
Obviously this PR, being a breaking change, will have to wait until an opportune moment.
Also, it will need more testing.

### DIFF
--- a/encryption_protocol.txt
+++ b/encryption_protocol.txt
@@ -14,9 +14,12 @@ transactions and not to Maker and Taker entities.
 Encrypted
 =========
 
-TAK: !auth <input utxo pubkey> <btc sig of taker encryption pubkey using input utxo pubkey>
+TAK: !auth <input utxo pubkey1[, input utxo pubkey2,..,M]> <btc sig of taker encryption pubkey using input utxo pubkey>
 (Maker verifies the btc sig; if not valid, connection is dropped - send REJECT message)
-MAK: !ioauth <utxo list> <coinjoin pubkey> <change address> <btc sig of maker encryption pubkey using coinjoin pubkey>
+If the input utxo pubkey field is a comma-separated list instead of a single value, it is interpreted as the data
+for a multisig address, with the last field being the M in M of N (and N being the number of pubkeys. The btc sig
+can then be using any of the given pubkeys.)
+MAK: !ioauth <utxo list> <input pubkey> <coinjoin address> <change address> <btc sig of maker encryption pubkey using input pubkey>
 (Taker verifies the btc sig; if not valid, as for previous)
 
 Because the !auth messages are under encryption, there is no privacy leak of bitcoin pubkeys or output addresses.
--- a/lib/common.py
+++ b/lib/common.py
@@ -97,6 +97,22 @@ def get_network():
 	'''Returns network name'''
 	return config.get("BLOCKCHAIN","network")
 
+def get_multisig_addr(pubkeys, M, N):
+	'''Note on usage: It is critical to remember
+	that the multisig address constructed DOES depend
+	on the order of the pubkeys! An attempt to sign off
+	a transaction must have the order of signatures correct.'''
+	mscript = btc.mk_multisig_script(pubkeys, M, N).decode('hex')
+	return btc.script_to_address(mscript, get_magic_byte())
+
+def get_magic_byte():
+	if get_addr_vbyte() == 0x00:
+		return 5
+	elif get_addr_vbyte() == 0x6f:
+		return 196
+	else:
+		raise Exception("Unrecognized network type")
+
 def get_addr_vbyte():
 	if get_network() == 'testnet':
 		return 0x6f
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -236,12 +236,12 @@ class IRCMessageChannel(MessageChannel):
 						self.on_order_fill(nick, oid, amount, taker_pk)
 				elif chunks[0] == 'auth':
 					try:
-						i_utxo_pubkey = chunks[1]
+						pubkeys = chunks[1].split(',')
 						btc_sig = chunks[2]
 					except (ValueError, IndexError) as e:
 						self.send_error(nick, str(e))
 					if self.on_seen_auth:
-						self.on_seen_auth(nick, i_utxo_pubkey, btc_sig)
+						self.on_seen_auth(nick, pubkeys, btc_sig)
 				elif chunks[0] == 'tx':
 					b64tx = chunks[1]
 					try:
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -129,9 +129,9 @@ class IRCMessageChannel(MessageChannel):
 	def send_pubkey(self, nick, pubkey):
 		self.__privmsg(nick, 'pubkey', pubkey)
 
-	def send_ioauth(self, nick, utxo_list, cj_pubkey, change_addr, sig):
-		authmsg = (str(','.join(utxo_list)) + ' ' + 
-	                cj_pubkey + ' ' + change_addr + ' ' + sig)
+	def send_ioauth(self, nick, utxo_list, input_pubkey, cj_addr, change_addr, sig):
+		authmsg = (str(','.join(utxo_list)) + ' ' + input_pubkey + ' ' +
+	                cj_addr + ' ' + change_addr + ' ' + sig)
 		self.__privmsg(nick, 'ioauth', authmsg)
 
 	def send_sigs(self, nick, sig_list):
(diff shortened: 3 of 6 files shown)


AdamISZ:
Addresses #88

chris-belcher:
Good job.

The abstract functions in message_channel.py should be changed too.

Now for actually merging this.
One way is to merge it now while the project is small so it's less hassle.
Another way is to wait until more features like https://github.com/chris-belcher/joinmarket/issues/45 especially minimum maker count, and then ask everyone to update.
Another way is to create another command different from !ioauth and code takers to understand both, this is a kind of soft fork option. It would take much more code for this.

Would appreciate opinions on the best way to do this. I'm leaning towards *code the minimum-maker-count feature and release ASAP*.

chris-belcher:
Further point, if we're changing the authentication it might be worth fixing this issue at the same time https://github.com/chris-belcher/joinmarket/issues/83

AdamISZ:
Thanks for the abstract method catch. Fixed.

Re: #83 , it seems like one solution might be: replace

    auth input_utxo_pubkey btc_sig

with

    auth [input_pubkey1,input_pubkey2,..,M] btc_sig

where it's understood that if there is more than 1 pubkey, it's a multisig, and the maker should read the last entry in the list as M in M of N. Then he can verify the multisig address from these pubkeys and also verify that btc_sig is signed by the (first?) of the pubkeys in the list. A list of only one input_pubkey is read as meaning an ordinary non-multisig.

chris-belcher:
That sounds good.

btc_sig should be checked against any pubkey in the list because the taker's pubkey may happen to not be the first one.

Only issue I can think is there may be other kinds of p2sh scripts that are not multisig. Like some exotic contract.
I can't think of any situation where the script doesn't contain some public key and signature, otherwise the miner could take the same input and modify the output to steal the money. So what if the auth command has the entire scriptPubKey in it, the maker parses it for any pubkeys and the signature has to verify with one of them.
We have to ask whether this is worth it, all p2sh usage today is some kind of multisig. Almost everything on the bitcoin wiki contracts page is based on multisig and for the stuff that isn't it's not clear whether it actually requires coinjoin privacy.

AdamISZ:
I can do it now but I think it's worth looking at support-multisig-input as a "soft fork", because when it's introduced it will not break older clients/bots (according to the syntax above, let's say). 

Whereas the changes currently in this PR are a "hard fork", breaking everyone's auth mechanism. So combine with #29 let's say (but even then, just make separate PRs but merge at the same time), plus any other breaking changes you want to do.

As for more exotic contracts, I don't think it's pressing. Re: check all pubkeys, definitely, yes, we already do something similar to that anyway.

chris-belcher:
I think any "hard fork" should happen sooner rather than later. It gets more disruptive as time goes on because more users need to update.

It's nowhere near as disruptive or dangerous as a bitcoin hard fork. No value is at risk, it just means the two kinds of bots cannot communicate. The liquidity is split in two but there's so little of that right now it's not too much of a loss. The community is small and interested. So I'd say do a hard fork soon, I'll work on the *minimum maker count* feature.

AdamISZ:
OK, seems like a good point of view. I can do the multisig thing and include it in this PR if you like.

chris-belcher:
Yes please.

chris-belcher:
At the same time as this change, we could add a field to later implement this idea by theymos

> In exchange for a (comparatively large) extra fee, takers could require that the unspent-outputs provided by makers be x blocks deep (I'm thinking ~1500). This reduces the Sybil risk because an attacker will only have so many bitcoins, and this requirement ties up a lot of their bitcoins for a while if a lot of takers are routinely requiring that at least some of their partners provide old bitcoins.

The extra field announces a block height of the UTXOs, with zero meaning don't know/don't care.

AdamISZ:
fd76aa6 allows the maker to receive multisig address inputs in the !auth message and verifies the signature provided matches one of the given pubkeys. It also sanity checks the multisig parameters (M and N) and checks (as before) that the claimed multisig appears in the transaction.

What I haven't done yet is address what is going on on the maker side (it seems still a bit of an open question how the taker is going to create the multisig signatures as you described in #83). I was toying with the idea of putting it in crudely, via entries on the command line, but it seemed there's still too many open questions there.

All I have tested is a normal transaction in joinmarket-pit-test to make sure that standard transactions still work correctly.

chris-belcher:
Whoops forgot to reply to this.

I'll make a git fork which will have all the commits of the new protocol. Ready to be merged into master one day after everything works.

For the taker side we'll have a new script called raw-tx.py or something, which returns a partially unsigned transaction hex that the user has to copypaste to whoever has the private keys to get them signed. It will also work for normal p2pubkeyhash, although the the user has to feed in a pubkey for the encryption

chris-belcher:
created branch newprotocol and merged this in

adlai:
How about using a different order type for the new protocol? That lets people "soft fork" while outdated clients will simply ignore the new type.

chris-belcher:
What are the advantages of that? With this way the clients also ignore each other.

chris-belcher:
I see what you mean.

Yeah I guess the main reason is I don't feel like programming that up. Seems easier to just ask everyone to update :|

chris-belcher:
Clarification:
https://github.com/chris-belcher/joinmarket/blob/newprotocol/lib/maker.py#L69 This should be range(2, 17) to allow M-of-2 multisig

Or even range(1, 17) to allow 1-of-1 multisig in a future where we all use p2sh for everything

AdamISZ:
No, len(pubkeys) is 1 longer due to format [pubkey1,pubkey2, .. M] as I wrote in the changes to the *txt file.

As for 1 of 1 I also briefly considered it but decided against it - but sure, OK.

chris-belcher:
I've become increasingly aware that spending from hardware wallets and cold storage, as in http://www.reddit.com/r/joinmarket/comments/36xyw5/any_way_of_doing_multisig_or_hw_wallets_like_the/ cannot be done with the current auth protocol because the private key is needed. People who keep private keys offline won't want to transfer them to an online computer only to use them to authenticate the encrypted stream.

What do you think of optionally allowing takers to not authenticate their side? What are the cryptographic implications of only the maker authenticating himself? Of course most of the time the taker will authenticate, only in the multisig or hardware wallet spending they wont.

Could be implemented by the taker sending Null for the signature, which means the maker skips the ecdsa verifying stuff.

AdamISZ:
Given this perspective, I am struggling to find a very good argument for keeping the authentication to btc signatures, at least on the taker side. One other possibility (apart from just removing it) is to replace the signature to the input_utxo_pubkey with a signature to some chosen as-persistent-as-you-like bitcoin address. In this I am remembering theymos' comment:

>- Instead of requiring NickServ registration, makers could generate an identity separately (maybe just a Bitcoin address) and communicate it on the IRC channel using public-key crypto. This is more convenient and will work across multiple IRC servers.

On the maker side, it feels like a stronger argument to insist on some sort of verification of holding the claimed funds (although, as currently coded, the maker only attests to one small chunk of coin). But I am struggling to come to any strong opinion about all of this.

chris-belcher:
If one side authenticates it makes a MITM much more difficult. Since by design the maker always has immediate access to his private key, it makes sense to use it for authenticating as well.
So from the point of view of takers, they are always talking to the owner of the UTXOs they are about to join with.

I think theymos' idea was meant to be related to https://github.com/chris-belcher/joinmarket/issues/57 Using an identity related to a bitcoin address is equivalent to trust-on-first-use, which is slightly worse than using the UTXO pubkey.

If there's no serious implications of taker authentication being optional, I'll code it up later.

AdamISZ:
But it's a bit more complicated than that because, don't forget, we are using libnacl for authenticated encryption to an ephemeral pubkey per transaction. So we do have MITM protection during any particular transaction with all counterparties. 

TOFU - well, not necessarily, in as much as bitcoin keys could be used in a WoT a la bitcoin-otc. Using just a random bitcoin key without any persistence would be a pointless add-on to the authentication we already have in libnacl.

chris-belcher:
Sorry, a bit more complicated than what?

So libnacl already has authentication? How can it do that without distributing keys somehow? I thought in TLS the certificate authorities are used for distributing the correct keys, and in JoinMarket so far we use the UTXO pubkey instead.

OK yes agreed, TOFU is not the only way to use bitcoin-address-as-identity.

AdamISZ:
We generate keypairs on the fly for each new counterparty, for each new transaction. Then a session key is generated using DH key exchange. Part of that session key is used to create MACs for authentication, using Poly1305. This verifies that only one entity is conducting the entire exchange, as well as ensures the integrity of the messages. That's why I qualified with "during any particular transaction with all counterparties. " Obviously it does nothing to address the question of who you're starting your conversation with, only that it stays the same. But it's still classed as "authenticated encryption".

Upshot of all this, I see no reason that optionally removing the btc sig on taker side would cause any problems right now. Maybe on IRC we can discuss and come to a decision about where we see the whole "persistent identity" thing going.

chris-belcher:
Okay, understood. Interesting subtle differences. :)

I'll come on IRC a little later tonight.

PR #93, opened 2015-05-28
Sort the orderbook

the9ull:
Hello,
I ordered the orderbook
Martino


chris-belcher:
I agree in principle. But what would happen to absolute orders? Could you make them appear at the top next to the cheapest orders, I think like this they will appear at the bottom after the most expensive.

Right now absorders are only made by patientsendpayment.py so they're likely to be cheap.

the9ull:
I tested it, now absorders are on the top

chris-belcher:
What do you think of extending this so the user can choose several ways of sorting. Cheapest first, largest first, group orders from the same counterparty, etc

PR #98, opened 2015-05-29
bugfix for pick_order (-P)

anduck:


chris-belcher:
I agree of course. I'll merge soon.

By the way, what happens if someone does -C and -P at the same time?

Do you think it might be bette to have --mode=weighted, -mode=cheapest -mode=pick
or something like that.

anduck:
-P overrides -C, except for sweep which doesn't have -P.

I think switching to --mode could be good if there's more possible modes.

PR #99, opened 2015-06-01
Multiply orders over mix depths

shialsuoto:
The modification gives ability for setting up different prices for different mix levels to a maker.
Taker can pay less for smaller order sizes.

### DIFF
--- a/yield-generator.py
+++ b/yield-generator.py
@@ -12,13 +12,15 @@ import common, blockchaininterface
 
 from socket import gethostname
 
-txfee = 1000
-# fees for available mix levels from max to min amounts. mix_levels-1
-cjfee = ['0.0008','0.0007', '0.0006', '0.0005']
+#txfee = 1000
+# tx fees for available mix levels from max to min amounts. mix_levels is max. 
+txfee = [1000,1000,100,100]
+# fees for available mix levels from max to min amounts. mix_levels is max
+cjfee = ['0.0009','0.0008', '0.0007','0.0006']
 nickname = random_nick()
 nickserv_password = ''
 # minsize is dynamic and defined during orders creation in create_my_orders
-#minsize = int(1.2 * txfee / float(cjfee)) #minimum size is such that you always net profit at least 20% of the miner fee 
+#minsize = int(1.2 * txfee / float(cjfee)) #minimum size is such that you always net profit at least 20% of the miner fee
 mix_levels = 5
 
 
@@ -63,10 +65,10 @@ class YieldGenerator(Maker):
                 oid=0
                 for cj in cjfee:
                           max_mix = max(mix_balance, key=mix_balance.get)
-                          if  mix_balance[max_mix] - common.DUST_THRESHOLD > 0:
-                                    minsize = int(1.2 * txfee / float(cj))
+			  minsize = int(1.2 * txfee[oid] / float(cj))
+                          if  mix_balance[max_mix] - common.DUST_THRESHOLD > 0 and minsize < mix_balance[max_mix] - common.DUST_THRESHOLD:
                                     order = {'oid': oid, 'ordertype': 'relorder', 'minsize': minsize,
-                                            'maxsize': mix_balance[max_mix] - common.DUST_THRESHOLD, 'txfee': txfee, 'cjfee': cj}
+                                            'maxsize': mix_balance[max_mix] - common.DUST_THRESHOLD, 'txfee': txfee[oid], 'cjfee': cj}
                                     orders.append(order)
                                     oid+=1
                                     del mix_balance[max_mix]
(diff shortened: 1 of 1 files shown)


chris-belcher:
I agree in principle. Creating more detailed sell-side algorithms is important to meet the requirements of the market.

I've not yet tested.

chris-belcher:
I haven't gotten round to testing this yet. But i found another money making opportunity for you if you want to code a tiny bit more.

https://github.com/chris-belcher/joinmarket/issues/23#issuecomment-108669871

K1773R:
why is it mix_levels-1 ? ie, why not creating orders for all mix-depths?

shialsuoto:
Current logic is that all mix levels are sorted in descending order before applying cjfee from configuration. In the case last sorted mix_level is smallest one and in most cases it's just dust changes. 

I agree with Chris that when we decrease cjfee for smaller amounts our minsize is increasing. One solution would be ability for setting txfee for each mix-depths. Smaller txfee for smaller amounts that is logical will stabilize minsize.

chris-belcher:
@shialsuoto suppose we have 1btc in total across all mix depths, and they are arranged like 0.2, 0.2, 0.2, 0.25, 0.15btc. Then the smallest balance is 0.15btc which is more than just dust in this case.

I think the maker would actually be happy if somebody coinjoined with that mix depth because it would move those coins to another depth and then the max balance would be 0.40btc or so. So it sounds like K1773R is right to say all mixing depths should be announced?

In your code, it should skip announcing the order if minsize > maxsize for that depth. So dust orders that are impossible to fill wont be announced.

shialsuoto:
Sorry, let me clarify main idea for the pull request. Probably I had to put it to pull request description first.

Initially let's remove mix depths from the picture.  I'm proposing ability for multi-levels cjfees for market makers (MM). In general it should give better MM's capital involvement. Because MM's competition is unavoidable  and we are increasing involved capital it would lead to better offers for takers.  From the point of view following MM's offers

cjfee	maxsize
0.02 	1
0.015	0.5

translates to taker that MM guarantees max size of taken fill order up to 1btc with fee 0.02% or up to 0.5btc with fee 0.015. Mixed depths are out of taker scope. MM just guarantees max order size from his side for described fee.

Now to MM's technicals details. In general cjfee list is exposure of MM's wallet to market. Technical limit is max mix depths in a wallet. Rest is matter of MM's willingness for the exposure. I agree the mix_levels-1 is just my point of view, because I saw only "change" on last level. 

Example how it should work for MM (assuming full exposure of 4 mix depth levels wallet)

cjfee	maxsize	mix_depth    oid
0.02 	1	        0                    0
0.015	0.5	        1                    1
0.005	0.3	        2                    2
0.002	0.1	        3                    3

MM has fill request for oid 1 for size 0.15. Taker has following mix depth available for the order due to used cjfee.

maxsize	mix_depth    oid
0.5	        1                    1
0.3	        2                    2
0.1	        3                    3

I didn't touch logic for choosing mix_depth for mix and  Taker order will be taken from mix_depth 2 (best fitted mix depths). After the order execution MM will offer:

cjfee	maxsize	mix_depth    oid
0.02 	1	        0                    0
0.015	0.5	        1                    1
0.005	0.25	        3                    2
0.002	0.15	        2                    3

As you can see maxsize paradigm is still in place, however it's more dynamic and MM has ability for setting up different price levels.

--- 
I will add the "minsize > maxsize" skip.

chris-belcher:
An interesting write-up. Thanks.

Could you spell out for me why the number of orders is at most mix_levels-1 instead of mix_levels? I'm still not getting it.

shialsuoto:
@chris-belcher I've corrected the comment is source code. Max technical orders number is equal mix_levels.
Personally I use mix_levels-1, because smallest mix depth in my wallet is on level of dust so far.

CohibAA:
I like having this option for makers, nice work on this request, imho.

chris-belcher:
In the past few days I've been working on editing this PR slightly to add an absorder between zero and minsize, to fix the *Not enough liquidity* error people get when running tumbler that picks very small coinjoin amounts.

chris-belcher:
I believe I have finished a new yield generator algorithm, which is heavily based on this PR.

It's on the branch https://github.com/chris-belcher/joinmarket/tree/new-yieldgen-algo Worth a try although beware it's somewhat complex so there may still be bugs around. Run wallet-tool summary and watch out if any amounts are announced by yield generator incorrectly.

Eventually this could have logic like in #129 added.

raedah:
This pull request is for work that was prior to the new-yieldgen-algo branch. It initially inspired direction. We are now many iterations beyond it. I suggest that this pull request is closed. Anyone looking for the functionality should try out yield-generator-deluxe.py.

AdamISZ:
Agreed, closing.

PR #101, opened 2015-06-13
Improved ordering

the9ull:
> What do you think of extending this so the user can choose several ways of sorting. Cheapest first, largest first, group orders from the same counterparty, etc

I was thinking the same thing.
This inplementation look like: http://the9ull.silix.org/imageupload/4bc8278c/ordering.png


chris-belcher:
Looks good. I'll try it soon and then commit.

chris-belcher:
Merged.
It should be noted this project uses currently tabs for indenting.

CohibAA:
Wow, I had literally just logged in to submit a request for this, using http://www.kryogenix.org/code/browser/sorttable/ instead - I have it implemented here http://bitcoin-p2pool.com/joinmarket_orderbook.html - glad I got the email before I spent much time on it, lol.

Anyway, great change for usability of the orderbook, thanks!

PR #107, opened 2015-06-26
Added minsize > maxsize error message - Fixes #106

### LINKED ISSUE #106: Checking if minsize is greater than maxsize
This is a really easy fix.

After experimenting for almost an hour, I learned that in my configuration, I had accidentally set it so that the minsize was greater than the maxsize (ob-watcher.py notifies you of that if you're running it).

It would be a super simple, five minute fix that involves including a check just after line 63:

    if order["minsize"] >= order["maxsize"]:
        print("Minimum order size is greater than maximum order size.  Please  adjust this in the configuration.")

btcspry:
Fixes #106 (issue #106)

Added log statement regarding the minsize being greater than the maxsize, and proves a warning.  The method of providing the output may require change, as it currently uses `self.log_statement([timestamp ... , "Big Message..."])`

### DIFF
--- a/yield-generator.py
+++ b/yield-generator.py
@@ -61,6 +61,9 @@ class YieldGenerator(Maker):
 		max_mix = max(mix_balance, key=mix_balance.get)
 		order = {'oid': 0, 'ordertype': 'relorder', 'minsize': minsize,
 			'maxsize': mix_balance[max_mix] - common.DUST_THRESHOLD, 'txfee': txfee, 'cjfee': cjfee}
+		if order['minsize'] > order['maxsize']:
+			timestamp = datetime.datetime.now().strftime("%Y/%m/%d %H:%M:%S")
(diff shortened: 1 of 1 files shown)


chris-belcher:
Printing out an error message doesn't do much, it can still be ignored by people.

That log statement is meant to be used for record keeping not for error messages, it can be opened with spreadsheet software and be manipulated.

This problem will be fixed for good with the new yield generator algorithm coming out shortly.

btcspry:
Printing out an error message notifies them that there's an error.  It's actually a really common way of provided feedback to the users [sarcasm]

If `self.log_statement` isn't the right option, then use a print statement or something else

Also, are you going to be porting to Python3, or are your plans to stick with the legacy version?

chris-belcher:
Eventually to python3, yes.

PR #109, opened 2015-06-29
Add warning for low maker count and related info to sendpayment.py.

CohibAA:
When starting sendpayment.py, check if makercount is below default (2), if so, confirm send request before connecting to irc.  If makercount set to default (2), or unset, display info about increasing for increased privacy (and fees).  If makercount is set above default (2), display the count and info before proceeding.

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -116,14 +116,15 @@ def main():
 			'wallet. Requires blockchain_source=json-rpc')
 	(options, args) = parser.parse_args()
 
-        if options.makercount < 2:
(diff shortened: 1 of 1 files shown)


chris-belcher:
Yeah these could be merged I guess. A warning is better than no warning.

chris-belcher:
Could you make these check options.answeryes so these files can be scripted without user input

i.e. running it with --yes skips the prompt

CohibAA:
@chris-belcher updated to support answeryes

PR #111, opened 2015-06-29
Update sendpayment.py

CohibAA:
When starting sendpayment.py, check if makercount is below default (2), if so, confirm send request before connecting to irc. If makercount set to default (2), or unset, display info about increasing for increased privacy (and fees). If makercount is set above default (2), display the count and info before proceeding.


CohibAA:
duplicate, sorry

PR #113, opened 2015-06-29
Torssl

abitfan:
fixed tor/ssl connection


CohibAA:
Confirmed, tor is working for me now, thanks.  I am connected using joinmarket.cfg show below - take note of port, usessl must be true, socks5 must be true, socks5_port must match torrc config.

[MESSAGING]
#host = irc.cyberguerrilla.info
channel = joinmarket-pit
#for clearnet
port = 6697
#port = 6667
usessl = true
#socks5 = false
socks5_host = 127.0.0.1
socks5_port = 9050
#for tor
#host = 6dvj6v5imhny3anf.onion
host = a2jutl5hpza43yog.onion
#usessl = false
socks5 = true

chris-belcher:
Thanks for testing @CohibAA 

Does it also work with all combinations of true/false. socks5 with no ssl, ssl with no socks5, false for both.

CohibAA:
@chris-belcher - the only combination I could make connect over tor is above.  It would not connect to tor without ssl enabled or without socks5 proxy.

chris-belcher:
OK I'll try with a socks5 ssh tunnel when I get the chance.

abitfan:
@CohibAA you always have to use socks5 for tor connections and cyberguerrilla enforces ssl so you need to have both on.
@chris-belcher tested all combos

CohibAA:
@abitfan thanks again.

PR #119, opened 2015-07-02
Fix BitcoinCoreWallet's UTXO selection.

domob1812:
The UTXO selection of BitcoinCoreWallet with fromaccount set did accept also UTXO's without an account.  This is in direct contradiction to the user's wish for a specific account.  With this patch, we really only accept UTXO's that have the specified account set explicitly.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -323,7 +323,7 @@ class BitcoinCoreWallet(AbstractWallet):
 		for u in unspent_list:
 			if not u['spendable']:
 				continue
-			if self.fromaccount and 'account' in u and u['account'] != self.fromaccount:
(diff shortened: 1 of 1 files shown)


chris-belcher:
Good fix.

haarts:
I have a hard time following what is being checked for. Is it an idea to extract the right hand side of the if statement into a predicate method?
```python
if self.fromaccount and isSomeSaneName(u):
    continue
...
```

chris-belcher:
This code is meant to allow the user to choose which bitcoin core account to spend from, see https://bitcoin.org/en/developer-reference#listunspent

PR #121, opened 2015-07-03
Change how random irc nicknames are generated.

CohibAA:
I believe this is a more anonymous and random method and possibly more efficient way to generate random_nick for ircnick.  Thoughts?

### DIFF
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -3,10 +3,9 @@ from common import *
 from message_channel import MessageChannel
 from message_channel import CJPeerError
 
-import string
-import random
+import string, random
 import socket, threading, time, ssl, socks
-import base64, os
+import base64, os, re
 import enc_wrapper
 
 MAX_PRIVMSG_LEN = 400
@@ -18,6 +17,8 @@ plaintext_commands = ["fill", "error", "pubkey", "orderbook", "relorder", "absor
 
 def random_nick(size=9, chars=string.ascii_uppercase + string.ascii_lowercase + string.digits):
         ircnick = ''.join(random.choice(chars) for _ in range(size))
+        if re.match('\\d', ircnick[0-9]):
+	        ircnick = '_' + ircnick
         print 'Generated random nickname: ' + ircnick #not using debug because it might not know the logfile name at this point
         return ircnick
 
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -3,8 +3,10 @@ from common import *
 from message_channel import MessageChannel
 from message_channel import CJPeerError
 
+import string
+import random
 import socket, threading, time, ssl, socks
-import base64, os, urllib2, re
+import base64, os
 import enc_wrapper
 
 MAX_PRIVMSG_LEN = 400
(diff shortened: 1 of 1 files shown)


CohibAA:
Forgot to mention, I believe urllib2 may have also been causing an error in tor logs when generating the nickname...

    [warn] Socks version 80 not recognized. (Tor is not an http proxy.)

but I am not 100% sure of the cause of the tor log error, so feedback is appreciated.

CohibAA:
I am not too familiar with restrictions on irc naming, so maybe this still needs to check for naming rules and try again if any violations occur.  What restrictions exist for irc nicknames? Can any digits or letters be the first character?.  I didn't have any issues while testing, but maybe the re.match check should not be removed, or should be modified somehow?

chris-belcher:
I used wikipedia because I thought the english-language words would make it less clear which IRC nick was a bot and which was an actual human. It might be important for a tumbler bot for example which idles in the channel for a long time and never publicly says anything.

CohibAA:
That makes sense, regarding why to use Wikipedia for names.  My primary concern was the call to get the wiki page and urllib2's impact on tor integrity, hence this workaround method.

Regarding urllib2 and tor, [this](https://stackoverflow.com/questions/5148589/python-urllib-over-tor) is likely what caused the tor log warnings.

    The problem is that httplib.HTTPConnection uses the socket module's create_connection helper function which does the DNS request via the usual getaddrinfo method before connecting the socket.

I have not seen the warnings while not using the urllib2 import.  This is the solution suggested on stackflow.

    The solution is to make your own create_connection function and monkey-patch it into the socket module before importing urllib2, just like we do with the socket class.

Suggested code example:

    import socks
    import socket
    def create_connection(address, timeout=None, source_address=None):
        sock = socks.socksocket()
        sock.connect(address)
        return sock
    
    socks.setdefaultproxy(socks.PROXY_TYPE_SOCKS5, "127.0.0.1", 9050)

    # patch the socket module
    socket.socket = socks.socksocket
    socket.create_connection = create_connection
    
    import urllib2

I haven't researched completely nor tested at all, but this looks to fit at first glance.  I will revert to the Wikipedia names method and try this solution to see if the tor warnings stay away.

WyseNynja:
I like this PR.  If someone wants to "hide" in the channel by pretending to be a bot, they can pick some random numbers or letters just as easily as they can pick a random page off wikipedia.  There's no need to make an http request.

chris-belcher:
The http request really kills it actually.
Has anyone tried this yet? I'll merge after trying it.

kristovatlas:
There's python code for connecting to Tor in this script that works, if anyone is interested: https://github.com/kristovatlas/onion-uptime/blob/master/check-onion-status.py

tailsjoin:
I have tried this code and it works fine. It should be merged immediately imo because this unproxied http call is messing things up for Tails while using Bitcoin core. Torify is not a possible solution for my situation either as it will torify rpc calls as well.

CohibAA:
my watcher bot is also using this, no problems

chris-belcher:
Hope not to offend @CohibAA by throwing away his code, but I couldn't resist coming up with a way to produce more pronounceable names. As in https://gist.github.com/chris-belcher/cd51b9edf3465c19c8f1

CohibAA:
No offense!  My main concern was getting away from the http call, however it is accomplished.

PR #125, opened 2015-07-04
Prevents startup spam to log.

### LINKED ISSUE #122: Yield generator spamming up the income statement with headings
Every time the irc bot reconnects it will add a line of headings to the income statement csv file. https://github.com/chris-belcher/joinmarket/blob/master/yield-generator.py#L48

Instead it should check whether the file exists and only add the headings the first time it's created.

Also, if the bot is running under testnet it SHOULD NOT log coinjoins to file.

haarts:
Fixes #122 except for the testnet stuff.

### DIFF
--- a/ob-watcher.py
+++ b/ob-watcher.py
@@ -1,4 +1,3 @@
-
 import BaseHTTPServer, SimpleHTTPServer, threading
 from decimal import Decimal
 import urllib2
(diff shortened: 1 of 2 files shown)


chris-belcher:
You can check for testnet using

    from common import get_network
    if get_network() == 'testnet':

haarts:
Since `common` was already imported and I don't like to add stuff to the global namespace I did a `if common.get_network() ...` call.

chris-belcher:
on_tx_confirmed() should also check it's not testnet

in fact it's probably cleanest to put the testnet check in log_statement()

haarts:
Done.

PR #128, opened 2015-07-04
Change install instructions for LibSodium

btcspry:
When I ran `sudo apt-get install libsodium-dev` it didn't find it, so I changed the instructions to build the program from scratch in step 1 of installing the dependencies.  It works great, although it would be good for someone else just to make sure it's good.  It will help the less tech-savy people install this program from source.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -28,7 +28,7 @@ REQUIRED INSTALLATION DEPENDENCIES
 
 GET DEBIAN / UBUNTU INSTALL DEPENDENCIES:
 0. sudo apt-get update -y && sudo apt-get upgrade -y && sudo apt-get install python 
-1. sudo apt-get install libsodium-dev -y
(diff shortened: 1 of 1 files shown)


chris-belcher:
Could someone on ubuntu/debian actually try these? I merged the other one without really testing

haarts:
I've installed it a couple of days ago without doing this.
Which version (of Ubuntu/Debian) are you using?
I also think installing from source is something you want to avoid.

CohibAA:
I wrote it based on my install experience on a debian install, but maybe it already had a dependency for dev packages that I missed.  Sorry I didn't test on multiple systems.

btcspry:
Xubuntu (which should be the exact same as Ubuntu for packages) nor Raspberry Pi Debain knew what I was talking about when I tried `sudo apt-get install libsodium-dev`, so I figured source was my only option.  At the very least, include both so that people can try the source method if their package manger doesn't help.

AdamISZ:
This is addressed in the new install process for secp256k1 (builds libsodium from source), which will be merged in the near future.

Issue #127, opened 2015-07-04
Add a total for the table in ob-watcher.py


btcspry:
It would be great for stat keeping if there was either a JS function or part of the Python code that totaled the table so we could see:
- How many people there are
- The total value of all possible bitcoin joins

chris-belcher:
The ob-watcher page already shows you the number of counterparties and orders.

Yes the sum of all maxsizes would be interesting too.

PR #130, opened 2015-07-05
Bootstrap styling fix and added total of MaxSize in ob-watcher.py table

btcspry:
- Removed active class from navbar item so that it doesn't appear you're on the same page when you click the links in the navbar
- Added a total of the maxsize column in the `ob-watcher.py` code, and then added it to the table footer

### DIFF
--- a/ob-watcher.py
+++ b/ob-watcher.py
@@ -105,6 +105,7 @@ class OrderbookPageRequestHeader(SimpleHTTPServer.SimpleHTTPRequestHandler):
 
 	def create_orderbook_table(self, orderby, desc):
 		result = ''
+		totalMaxSize = 0.0
 		rows = self.taker.db.execute('SELECT * FROM orderbook;').fetchall()
 		if not rows:
 			return 0, result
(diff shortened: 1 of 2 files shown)


btcspry:
Fixes #127

chris-belcher:
Did you try using the sql SUM function
http://www.w3schools.com/sql/sql_func_sum.asp

I think it's better to display the sum of offered coins in the SECONDHEADING header.

btcspry:
I think that displaying it in the table is more a concise solution because it localizes that information.

CohibAA:
@btcspry Nice work, added to my [jm orderbook](http://bitcoin-p2pool.com/joinmarket_orderbook.html), thanks.

@chris-belcher That seems more consistent, but would it be better to maybe add a new header for TOTALMAXSIZE rather than including in SECONDHEADING?  Might be more flexible for different uses that way.

chris-belcher:
@CohibAA in that case what happens for the other ob-watcher http pages, like /depth and /ordersize

CohibAA:
Right, I started looking closer and there would be more changes involved than I thought.  I only use SECONDHEADING and MAINBODY for my public orderbook.html page, and not directly, so I was thinking it would be easier (for me) to just be able to call TOTALMAXSIZE somewhere in orderbook.html if I want to change the layout.

btcspry:
@CohibAA looks great, thanks!  Could you submit a pull request for the information at the bottom of your page.  I think it could be useful, and therefore it would get merged up to JoinMarket.io and really help people understand the project a little better.

CohibAA:
@btcspry - [133](https://github.com/chris-belcher/joinmarket/pull/133) submitted

PR #134, opened 2015-07-06
coin unit choices

CohibAA:
for [114](https://github.com/chris-belcher/joinmarket/issues/114) - added [unit choices](https://github.com/chris-belcher/joinmarket/issues/114#issue-92050486) and cleaned BTC decimal precision displayed, did not include [bp or ppm for cjfee](https://github.com/chris-belcher/joinmarket/issues/114#issuecomment-118338157), but can easily add if there is demand

### DIFF
--- a/ob-watcher.py
+++ b/ob-watcher.py
@@ -11,6 +11,16 @@ from irc import IRCMessageChannel, random_nick
 from common import *
 import common
 
+#  BTC (1  BTC = 1.00000000 BTC) # bitcoin
+# mBTC (1 mBTC = 0.00100000 BTC) # millibitcoin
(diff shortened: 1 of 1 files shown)


chris-belcher:
Thank you.

This probably needs a way for the user to change the unit through the GUI. There is already (unused) code here for getting input from the url which might be useful https://github.com/CohibAA/joinmarket-dev-cohibaa/blob/patch-10/ob-watcher.py#L157

Also `UnitDisplayType` is bad naming convention, it makes it look like a class. Should probably be `unit_display_type`.

CohibAA:
fixed naming

PR #136, opened 2015-07-06
Display tx fee contribution when picking orders.

domob1812:
In the 'pick orders' dialog, display the tx fee contribution in addition to the resulting CJ fee for each order.

I also changed the display formatting a bit, IMHO the new display is more readable as it is "tabular".  If this change is not desired, let me know and I'll revert it and just append the tx fee contribution at the end.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -384,7 +384,7 @@ def pick_order(orders, n, feekey):
 	print("Considered orders:");
 	for o in orders:
 		i+=1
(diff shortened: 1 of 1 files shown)


chris-belcher:
I agree in principle, and that implemention looks pretty good too.

CohibAA:
I agree that this is important information to display for takers, thanks.

adlai:
:+1:

Issue #139, opened 2015-07-06
NameError: global name 'statement_file' is not defined - Running Yield Generator


btcspry:
When I run the yield-generator it runs for about two minutes, and then tells me `NameError: global name 'statement_file' is not defined`.  I think this is a basic bug that needs to be fixed!

Mikadily:
Same here

PR #140, opened 2015-07-06
bug fix: 'statement_file' is not defined

Mikadily:
fix for issue #139


chris-belcher:
Note to self: actually test pull requests yourself before merging.

haarts:
It is I who should apologize. I relied to much on my experiences with compilers.

btcspry:
Thanks for fixing it!

PR #141, opened 2015-07-07
Implement JSON-RPC interface instead of CLI.

domob1812:
Replace the connection to Bitcoin Core via bitcoin-cli by a connection through its JSON-RPC interface.  This is cleaner, as it does not require to convert everything to/from strings.  It is also able to handle error codes.

See https://github.com/chris-belcher/joinmarket/issues/137.

I have tested that it works successfully for a `sendpayment.py` order.  More testing would probably be good, though.

### DIFF
--- a/lib/blockchaininterface.py
+++ b/lib/blockchaininterface.py
@@ -8,18 +8,61 @@ import bitcoin as btc
 import common
 import jsonrpc
 
+# This can be removed once CliJsonRpc is gone.
+import subprocess
+
+class CliJsonRpc(object):
+    """
+    Fake JsonRpc class that uses the Bitcoin CLI executable.  This is used
+    as temporary fall back before we switch completely (and exclusively)
+    to the real JSON-RPC interface.
+    """
+
+    def __init__(self, cli, testnet):
+	self.cli = cli
+	if testnet:
+	    self.cli.append("-testnet")
+
+    def call(self, method, params):
+	fullCall = []
+	fullCall.extend (self.cli)
+	fullCall.append (method)
+	for p in params:
+	    if isinstance(p, basestring):
+		fullCall.append(p)
+	    else:
+		fullCall.append(json.dumps(p))
+
+	res = subprocess.check_output(fullCall)
+
+	if res == '':
+	    return None
+
+	try:
+	    return json.loads (res)
+	except ValueError:
+	    return res
+
 def get_blockchain_interface_instance(config):
 	source = config.get("BLOCKCHAIN", "blockchain_source")
-	rpc_host = config.get("BLOCKCHAIN", "rpc_host")
-	rpc_port = config.get("BLOCKCHAIN", "rpc_port")
-	rpc_user = config.get("BLOCKCHAIN", "rpc_user")
-	rpc_password = config.get("BLOCKCHAIN", "rpc_password")
 	network = common.get_network()
 	testnet = network=='testnet'
-	if source == 'json-rpc':
+	if source == 'json-rpc-socket':
+		rpc_host = config.get("BLOCKCHAIN", "rpc_host")
+		rpc_port = config.get("BLOCKCHAIN", "rpc_port")
+		rpc_user = config.get("BLOCKCHAIN", "rpc_user")
+		rpc_password = config.get("BLOCKCHAIN", "rpc_password")
 		rpc = jsonrpc.JsonRpc(rpc_host, rpc_port, rpc_user, rpc_password)
 		bc_interface = BitcoinCoreInterface(rpc, network)
+	elif source == 'json-rpc':
+		bitcoin_cli_cmd = config.get("BLOCKCHAIN", "bitcoin_cli_cmd").split(' ')
+		rpc = CliJsonRpc(bitcoin_cli_cmd, testnet)
+		bc_interface = BitcoinCoreInterface(rpc, network)
 	elif source == 'regtest':
+		rpc_host = config.get("BLOCKCHAIN", "rpc_host")
+		rpc_port = config.get("BLOCKCHAIN", "rpc_port")
+		rpc_user = config.get("BLOCKCHAIN", "rpc_user")
+		rpc_password = config.get("BLOCKCHAIN", "rpc_password")
 		rpc = jsonrpc.JsonRpc(rpc_host, rpc_port, rpc_user, rpc_password)
 		bc_interface = RegtestBitcoinCoreInterface(rpc)
 	elif source == 'blockr':
@@ -411,14 +454,12 @@ class BitcoinCoreInterface(BlockchainInterface):
 			self.add_watchonly_addresses(wallet_addr_list, wallet_name)
 			return
 
-		ret = self.rpc('listtransactions', [wallet_name, 1000, 0, True])
-		buf = json.loads(ret)
+		buf = self.rpc('listtransactions', [wallet_name, 1000, 0, True])
 		txs = buf
 		# If the buffer's full, check for more, until it ain't
 		while len(buf) == 1000:
-			ret = self.rpc('listtransactions', [wallet_name, 1000,
+			buf = self.rpc('listtransactions', [wallet_name, 1000,
 					len(txs), True])
-			buf = json.loads(ret)
 			txs += buf
 		used_addr_list = [tx['address'] for tx in txs if tx['category'] == 'receive']
 		too_few_addr_mix_change = []
--- a/lib/blockchaininterface.py
+++ b/lib/blockchaininterface.py
@@ -1,5 +1,4 @@
 #from joinmarket import *
-import subprocess
 import unittest
 import json, threading, abc, pprint, time, random, sys, os, re
 import BaseHTTPServer, urllib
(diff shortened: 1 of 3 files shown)


fivepiece:
Does joinmarket need that kind of elevated access to Core?
Maybe it'll be better to use the REST interface?

chris-belcher:
Read the code, looks pretty good.

I will try to test soon. Eventually run a long-running yield generator with it.

CohibAA:
ACK - FYI, no problems here during testing of wallet-tool, ob-watcher, sendpayment, patientsendpayment (see [147](https://github.com/chris-belcher/joinmarket/pull/147)), and yield-generator. I did not test tumbler, yet.

chris-belcher:
Tumbler will almost certainly work if the others do.

Did you try getting a coinjoin as well with yield generator?

CohibAA:
I've received successful orders as a maker using yield-generator with this commit, but I have seen this now several times in the console (doesn't show in logs):

    [2015/07/11 xx:xx:xx] rpc: getrawtransaction ['redacted (transaction is valid and confirmed on blockchain']
    ----------------------------------------
    Exception happened during processing of request from ('127.0.0.1', 43359)
      File "/usr/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
        self.process_request(request, client_address)
      File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
        self.finish_request(request, client_address)
      File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
        self.RequestHandlerClass(request, client_address, self)
      File "/home/redacted/joinmarket/lib/blockchaininterface.py", line 283, in __init__
        SimpleHTTPServer.SimpleHTTPRequestHandler.__init__(self, request, client_address, base_server)
      File "/usr/lib/python2.7/SocketServer.py", line 649, in __init__
    self.handle()
      File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
      File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
      File "/home/redactedjoinmarket/lib/blockchaininterface.py", line 290, in do_HEAD
    txd = btc.deserialize(self.btcinterface.rpc('getrawtransaction', [txid]))
      File "/home/redacted/joinmarket/lib/bitcoin/transaction.py", line 76, in deserialize
    obj["version"] = read_as_int(4)
      File "/home/redacted/joinmarket/lib/bitcoin/transaction.py", line 57, in read_as_int
    return decode(tx[pos[0]-bytez:pos[0]][::-1], 256)
      File "/home/redacted/joinmarket/lib/bitcoin/py2specials.py", line 89, in decode
    result += code_string.find(string[0])
    UnicodeDecodeError: 'ascii' codec can't decode byte 0x80 in position 128: ordinal not in range(128)
    ----------------------------------------

This seems to result in the balances for available mix levels not being updated on the orderbook until another fill request is made.

chris-belcher:
Good find. That's definitely a bug with this.
Looks like getrawtransaction returns something that causes btc.deserialize() to crash. Looks like an encoding issue.

CohibAA:
Yes, the jsonrpc call returns something like:

    {"result":"RAWTRANSACTIONVALUEHERE","error":null,"id":"getrawtransactiontest"}

while the command line simply returns the bare string (result value from above).  I'm sure it's probably a relatively easy fix, but I don't know exactly how to do it myself for certain.  I think it might just need strip to remove the double quotes from result...

domob1812:
@CohibAA Thanks for testing!  This result of the JSON-RPC call is expected, that's how JSON-RPC works.  It should be handled correctly by my code (picking out the `result` part).  Not sure where the error you are seeing comes from.  Seems to be related to character encodings (possibly).  I can not imagine how, though, as the string is hex encoded.

Is there a chance you can send me the full log without censored tx privately?  E. g., by email with GPG or private message on Bitcointalk?

CohibAA:
@domob1812 sure, whats your id on bct, send me a pm?  [me@bct](https://bitcointalk.org/index.php?action=profile;u=329589;sa=summary)

domob1812:
Sent you a PM; I'm "domob" there.  I've now also started my own yield-generator with the patch applied, maybe I can see the problem live for myself.

domob1812:
Was the decoding error fixed by https://github.com/chris-belcher/joinmarket/commit/ad749d7ab97633e5c9d51eeab7bb036bb6b5236b?  I've rebased my patch now.

I've been running a yield generator for some time now with JSON-RPC, and it seems to work fine.  The only thing I see from time to time as the "utxo already spent" issue https://github.com/chris-belcher/joinmarket/commit/ad749d7ab97633e5c9d51eeab7bb036bb6b5236b.  This may occur more often now due to the longer confirmation times in the Bitcoin network recently.

chris-belcher:
I'm going to start running my yield generator with this.

chris-belcher:
I had to edit some stuff to make it work for me. I think we might be using different python versions that handle encoding slightly differently. I made it check whether the string is unicode and cast it back to an ascii string.
Also fixed the conflicts between some other commits done since then.

My yield generator is running with this now, I'll check for that issue of balances not being updated.

	diff --git a/lib/blockchaininterface.py b/lib/blockchaininterface.py
	index dd3ae45..e5c5e1e 100644
	--- a/lib/blockchaininterface.py
	+++ b/lib/blockchaininterface.py
	@@ -306,13 +306,12 @@ class NotifyRequestHeader(SimpleHTTPServer.SimpleHTTPRequestHandler):
	 			if unconfirmfun == None:
	 				common.debug('txid=' + txid + ' not being listened for')
	 			else:
	-				jsonstr = None #on rare occasions people spend their output without waiting for a confirm
	+				txdata = None #on rare occasions people spend their output without waiting for a confirm
	 				for n in range(len(txd['outs'])):
	-					txout = self.btcinterface.rpc('gettxout', [txid, n, True])
	-					if txout is not None:
	+					txdata = self.btcinterface.rpc('gettxout', [txid, n, True])
	+					if txdata is not None:
	 						break
	-				assert jsonstr != ''
	-				txdata = json.loads(jsonstr)
	+				assert txdata != None
	 				if txdata['confirmations'] == 0:
	 					unconfirmfun(txd, txid)
	 					#TODO pass the total transfered amount value here somehow
	@@ -385,6 +384,9 @@ class BitcoinCoreInterface(BlockchainInterface):
	 		if method != 'importaddress':
	 			common.debug('rpc: ' + method + " " + str(args))
	 		res = self.jsonRpc.call(method, args)
	+		if isinstance(res, unicode):
	+			res = str(res)
	 		return res
	 
	 	def add_watchonly_addresses(self, addr_list, wallet_name):
	@@ -410,13 +412,11 @@ class BitcoinCoreInterface(BlockchainInterface):
	 		if not set(wallet_addr_list).issubset(set(imported_addr_list)):
	 			self.add_watchonly_addresses(wallet_addr_list, wallet_name)
	 			return
	-		ret = self.rpc('listtransactions', [wallet_name, 1000, 0, True])
	-		buf = json.loads(ret)
	-		txs = buf
	+		txs = self.rpc('listtransactions', [wallet_name, 1000, 0, True])
	+		buf = txs
	 		# If the buffer's full, check for more, until it ain't
	 		while len(buf) == 1000:
	-			txs = self.rpc('listtransactions', [wallet_name, 1000, str(len(txs)), True])
	-			buf = json.loads(ret)
	+			buf = self.rpc('listtransactions', [wallet_name, 1000, str(len(txs)), True])
	 			txs += buf
	 		used_addr_list = [tx['address'] for tx in txs if tx['category'] == 'receive']
	 		too_few_addr_mix_change = []
	@@ -484,6 +484,7 @@ class BitcoinCoreInterface(BlockchainInterface):
	 				one_addr_imported = True
	 				break
	 		if not one_addr_imported:
	+			debug('importing the address ' + notifyaddr + ' so that walletnotify will fire')
	 			self.rpc('importaddress', [notifyaddr, 'joinmarket-notify', False])
	 		tx_output_set = set([(sv['script'], sv['value']) for sv in txd['outs']])
	 		self.txnotify_fun.append((tx_output_set, unconfirmfun, confirmfun))

chris-belcher:
I've been using this the last few days and it seems to work really well. Anyone else apart from @domob1812 tried it?

One issue is people have to update their joinmarket.cfg, but hopefully there's not too many lazy people using JoinMarket right now.

I'm thinking of merging this soon.

chris-belcher:
I've been thinking, when people upgrade to this it will break their configuration. I don't want that to happen because it discourages regular updating by people. Everyone should have to update very soon because of the bip66 signature fix (I will send an alert about this.) And I wouldn't want people to update then find they have to mess around with configuration before it works again.

So a way around this problem is to make it configurable which kind of json-rpc to use. In the .cfg file add an extra option

    blockchain_source = json-rpc
    blockchain_source = json-rpc-socket

We'll update the wiki so new people use json-rpc-socket
Then some time in the future, remove all the bitcoin-cli code and make json-rpc and json-rpc-socket mean the same thing. Before that we'll add a warning that prints for people still using bitcoin-cli

So this PR would require a small rewrite. And ideally another PR created when bitcoin-cli is phased out so it can be readily merged.

chris-belcher:
My blockchain is corrupted so I have to redownload. I'll test this.

CohibAA:
I have not seen the encoding issue since [ad749d7](https://github.com/chris-belcher/joinmarket/commit/ad749d7ab97633e5c9d51eeab7bb036bb6b5236b).

tailsjoin:
I have been successfully using this with no issues on yield-generator and sendpayment for 3 days.

tailsjoin:
Got this with a sendpayment today:

    Exception in thread Thread-2:
      File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
        self.run()
      File "sendpayment.py", line 66, in run
        self.taker.makercount, self.taker.chooseOrdersFunc)
      File "domobjoinmarket/lib/common.py", line 410, in choose_order
        for o in sqlorders if cj_amount >= o['minsize'] and cj_amount <= o['maxsize']]
      File "domobjoinmarket/lib/common.py", line 343, in calc_cj_fee
        real_cjfee = int((Decimal(cjfee) * Decimal(cj_amount)).quantize(Decimal(1)))
      File "/usr/lib/python2.7/decimal.py", line 547, in __new__
        "Invalid literal for Decimal: %r" % value)
      File "/usr/lib/python2.7/decimal.py", line 3873, in _raise_error
        raise error(explanation)
    InvalidOperation: Invalid literal for Decimal: u'Dugulexer'

Did the exact same sendpayment with belchers master and it worked fine just seconds later.

CohibAA:
@tailsjoin That was caused by [this](https://github.com/chris-belcher/joinmarket/pull/178), I was testing my own sendpayment.py using the fixed code, and it likely blocked your sendpayment.py bot when I had the badmakerbot activated for the test.  That was one of the random names my badmakerbot generated, sorry for the inconvenience.

PR #147, opened 2015-07-11
implement userpcwallet for patientsendpayment.py

CohibAA:
Was there a reason this has been left disabled?  I was able to initiate a patientsend via rpc (although never got a taker).  Everything seemed okay.  Curious if something else needs to be done to finalize this.  Thanks.

### DIFF
--- a/patientsendpayment.py
+++ b/patientsendpayment.py
@@ -150,8 +150,6 @@ def main():
 	if not options.userpcwallet:
 		wallet = Wallet(wallet_name, options.mixdepth + 1)
 	else:
-		print 'not implemented yet'
-		sys.exit(0)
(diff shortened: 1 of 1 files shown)


chris-belcher:
patientsendpayment is broken right now. The underlying protocol requires makers to know the private key of the coinjoin address they are sending to.

In actuality the sys.exit(0) should be for all everything, however thinking back to when I wrote it. Someone had asked me to implement --rpcwallet and I wanted to specifically make the point it didn't work for patientsendpayment.

The branch called newprotocol is working on fixing this amongst other stuff.

CohibAA:
Great, thank you, closing.

PR #150, opened 2015-07-12
use curl instead of wget for maximal portability

adlai:
see http://daniel.haxx.se/docs/curl-vs-wget.html for details


tailsjoin:
Agreed.

chris-belcher:
untested ACK

That commit code probably isnt portable to windows which doesnt have /dev/null

adlai:
-s silences the progress updates, but curl still outputs the headers (from -I) to stdout. If portability trumps spam, we could leave them, and let each person customize complete silencing for their platform if necessary. I've also added the same timeout as the wget code.

chris-belcher:
curl: option --connect-timeout: expected a proper numerical parameter

`--connect-timeout 1` seems to work. Trying this tonight.

chris-belcher:
Seems to work well.

chris-belcher:
edited the wiki to change wget to curl

PR #157, opened 2015-07-21
IOError when using wallets/wallet.json argument. Looks for wallets/wa…

lacksfish:
Just a small fix.

running

$ python wallet-tool.py wallets/wallet.json
(where wallet is saved after generation)

returned

  File "wallet-tool.py", line 52, in <module>
    wallet = Wallet(seed, options.maxmixdepth, options.gaplimit)
  File "/storage/Coinstuff/joinmarket/lib/common.py", line 172, in __init__
    self.seed = self.get_seed(seedarg)
  File "/storage/Coinstuff/joinmarket/lib/common.py", line 200, in get_seed
    raise IOError('wallet file not found')
IOError: wallet file not found

The wallet is checked for existence, which fails because the path is actually wallets/wallets/wallet.json

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -191,7 +191,9 @@ class Wallet(AbstractWallet):
 	def get_seed(self, seedarg):
 		self.path = None
 		self.index_cache = [[0, 0]]*self.max_mix_depth
-		path = os.path.join('wallets', seedarg)
+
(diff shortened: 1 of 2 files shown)


lacksfish:
I made a mistake, closing this.

PR #158, opened 2015-07-21
Wallet saving/creating security and filepath error fix

lacksfish:


chris-belcher:
1. Why [this](https://github.com/chris-belcher/joinmarket/pull/158/files#diff-062cede901db6c97c3ac1bf03a3865b5R195) ?

2. Its probably cleaner to simply exit with an error message if the wallet file already exists, rather than going through the motions of asking the user for a new file name.

3. please combine all those commits into one before submitting, as in http://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git It makes following development history easier.

By the way, for future reference you can use raw_input('message here') to give a slightly cleaner interface. Also it's technically more correct to say that private keys in the wallet will be lost not bitcoins themselves, because everyone should have a backup in the form of the 12 word mnemonic seed. (You all wrote that down right? right?!)

lacksfish:
1. The seedarg is the argument for the filepath as used in [this line](https://github.com/chris-belcher/joinmarket/pull/158/files#diff-062cede901db6c97c3ac1bf03a3865b5L194), but appending 'wallets/' to the seedarg results in the file path 'wallets/wallets/wallet.json', for example.

2. good point.

3. will do.

chris-belcher:
1. That is intended that way. You're meant to just write the wallet file name.
All over in the wiki it shows `wallet.json` being used.

PR #160, opened 2015-07-22
bugfix: load transactions past the 1Kth

adlai:
You can check the fix by deleting a zero or two (or three!), and verifying that all your transaction history is loaded in whichever chunk size you choose.


chris-belcher:
Won't this infinitely loop if your wallet has less than 1000 transactions?

So it sounds like this crashed your bot when you reached 1k

adlai:
No...? This worked perfectly for me, tested with amounts below and above my transaction count.

chris-belcher:
Yes you're right.

PR #163, opened 2015-07-25
Add travis for testing, testing files in test directory

AdamISZ:
Details:

* Test subdirectory created, with supporting files for regression tests:
* regtest_joinmarket.cfg - special config for regtest running, bitcoin.conf - for same, regtest.py same as previous but modified and put under /test
* Travis automated build test support: added .travis.yml to main directory, contains script to build joinmarket on Ubuntu1204 VM and then to run enc_wrapper and regtest scripts (just to start of course).

I've connected Travis to my AdamISZ/joinmarket repo (it uses githooks) and it'll automatically test all pushes to my master branch. The tests are currently passing (You (@chris-belcher) can obviously do the same if and when it seems like a good idea.

Future plans in this regard: not sure what to do first, but probably add tests involving a larger number of bots and/or more and randomised transaction sizes. One stumbling block for testing other things (wallets, tumbler etc.) is have to create non-interactive versions (like -yes for sendpayment) which sits awkwardly with wallet creation for example (current bots cheat by using the seed option on test network).

Created an explanatory post here (since it can help with install): https://www.reddit.com/r/joinmarket/comments/3ek6qf/testing_and_setup_automation/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ before_install:
  - sudo apt-get install -y automake
 install:
  - pip install numpy
+ - pip install pexpect
 script:
  - git clone git://github.com/jedisct1/libsodium.git
  - cd libsodium
@@ -31,6 +32,7 @@ script:
  - bitcoin-cli -regtest setgenerate true 101
  - cd test
  - python regtest.py
+ - python wallet-test.py
 branches:
  only:
   - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+language: python
+python:
+ - "2.7_with_system_site_packages"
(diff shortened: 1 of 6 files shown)


chris-belcher:
Looking good.

Your reddit thread appears to be deleted (?)

AdamISZ:
As commit notes say, this is now testing basic wallet function (only generate and recover for now) also. Build is still passing. I can give you a url for an image to show passing/failing status if you like. As it turns out there is a cool Python module 'pexpect' that can handle shell interactivity very well (just using pipes isn't good enough for things like getpass), so that problem is solved. At some point I'll add some tests for wallet display and test out deposits etc.

PR #167, opened 2015-07-29
fix bip66 issue with signatures

AdamISZ:
Tested 10000 transactions on regtest with old version: got 44/10000 failures; see http://pastebin.com/u6imE6MC for txhex of each failed tx. A quick way to find the source of the error is to search for the string '021f' (although it could be lower than 1f, but the likelihood of that is near zero).

Tested 10000 with this bip66 correction: got 0/10000 failures (of any type).

The value 44/10000 is closely in line with expectations: 1/(256*2) is the expected probability of this failure mode for one DER encoded value (half of the cases where the first byte of 32 is zero). However, each of these transactions had 1 input, meaning 2 DER encoded values, leading to a 1/256 or ~40/10000 expected number of failures.

Note that all failures were associated with the same error string: "error: {"code":-26,"message":"64: non-mandatory-script-verify-flag (No error)" , as expected.

Since a real joinmarket transaction might have ~6 inputs, we could well see 1/20 to 1/50 transactions failing this way.

It may be worth updating to https://github.com/simcity4242/pybitcointools/ , although I know we are considering using a different codebase entirely. simcity4242 seems to have identified a few other issues, although I can't say how important they are.

### DIFF
--- a/lib/bitcoin/transaction.py
+++ b/lib/bitcoin/transaction.py
@@ -151,15 +151,18 @@ def signature_form(tx, i, script, hashcode=SIGHASH_ALL):
 
 
 def der_encode_sig(v, r, s):
-    b1, b2 = safe_hexlify(encode(r, 256)), safe_hexlify(encode(s, 256))
-    if r >= 2**255:
-        b1 = '00' + b1
-    if s >= 2**255:
-        b2 = '00' + b2
-    left = '02'+encode(len(b1)//2, 16, 2)+b1
-    right = '02'+encode(len(b2)//2, 16, 2)+b2
-    return '30'+encode(len(left+right)//2, 16, 2)+left+right
-
+    """Takes (vbyte, r, s) as ints and returns hex der encode sig"""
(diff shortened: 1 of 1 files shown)


chris-belcher:
Good work!

After this gets merged I'll send an alert asking yield generators to update.

And look a bit at the other stuff in simcity4242's repos

benma:
This patch was applied to the bitcoin subdirectory, which is a copy of pybitcointools (https://github.com/vbuterin/pybitcointools/). This patch should be applied upstream. The bitcoin subdir in joinmarket should ideally only be a git submodule to pybitcointools or a joinmarket fork of it, so diverging code can be tracked.

AdamISZ:
@benma yes, but the issue is that fix the was applied by user @simcity4242, who created his own fork because the repo owner @vbuterin is not responding to PRs (at least it seems that way). See the notes above re: @simcity4242 's fork, and how this interacts with the other issue here about moving away from pybitcointools.

Issue #104, opened 2015-06-18
Move query_utxo_set() outside the loop


chris-belcher:
query_utxo_set() can query several UTXOs at once, it makes no difference when using Bitcoin Core, but it reduces the number of http requests when using blockr.io
So move it outside the loop and query all UTXOs at once, skipping over UTXOs that belong to the taker.

https://github.com/chris-belcher/joinmarket/blob/61e7bcfa357dd9648c8edaa68fe5ceb685d85755/lib/taker.py#L128

PR #172, opened 2015-08-03
move query_utxo_set out of the loop

AdamISZ:
Intended to address #104 .

Tested it in a few configurations, specifically against cases where the yield gen has 3-6 utxos and verified in the logs that the utxos were bundled and sent to query_utxo_set only once. Also as requested made sure that our own utxos aren't included.

Unfortunately regtest testing doesn't support blockr of course.


chris-belcher:
I've read the code. It looks good.

I'll test soon.

CohibAA:
I've tested this PR with both sendpayment and yield-gen bot via blockr source and bitcoin-rpc.  Successful transactions on all, no issues found.

Issue #91, opened 2015-05-26
Recode UTXO selection algorithm


chris-belcher:
Right now the UTXO selection algorithm is taken straight from pybitcointools

It could use a recode to more closely follow what bitcoin core does, as in http://bitcoin.stackexchange.com/questions/1077/what-is-the-coin-selection-algorithm

chris-belcher:
Another compromise between number of UTXOs and privacy/linkage is that the user sets a max number of UTXOs they want to have at one time, the wallet merges them to stay below that number and splits them to go higher.

AdamISZ:
Was this addressed by #173 or is there more to do?

chris-belcher:
There may be more to do. In the link in OP I was thinking of this comment http://bitcoin.stackexchange.com/a/29962 which describes how bitcoin core chooses UTXOs, which tries to optimize for privacy. #173 is a good step but it only helps for not making transactions too large with too many inputs.

chris-belcher:
Regarding #173 we could do this

    if len(wallet_utxos) < 10:
        select_algo = select_default
    elif 20 <= len(wallet_utxos) < 30:
        select_algo = select_greedy
    elif 30 <= len(wallet_utxos):
        select_algo = select_greediest

With configurable threshold numbers instead of hardcoded like that.

PR #173, opened 2015-08-05
combat dust thru utxo merge policies

adlai:
cf #91

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -5,7 +5,7 @@ from math import factorial, exp
 import sys, datetime, json, time, pprint, threading, getpass
 import random
 import blockchaininterface, slowaes
-from ConfigParser import SafeConfigParser
+from ConfigParser import SafeConfigParser, NoSectionError
 import os, io, itertools
 
 JM_VERSION = 2
@@ -50,6 +50,10 @@ socks5_port = 9050
 #port = 6697
 #usessl = true
 #socks5 = true
+
+[POLICY]
+#for dust sweeping, try merge_algorithm = gradual
+merge_algorithm = default
 """
 
 def load_program_config():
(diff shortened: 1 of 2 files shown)


CohibAA:
Nice PR.  Testing...

tailsjoin:
I will also test. Hopefully it can deprecate my pycoin reallocation tool.

chris-belcher:
I realise there's a few different meaning to dust in bitcoin.

1. Outputs too small to be relayed by the network, as in common.DUST_THRESHOLD
2. Outputs that cost more to spend than they're worth. Depends on the fee/kB you're using.
3. Outputs that are small enough to be annoying but otherwise spendable as normal.

Unless I'm wrong, these new utxo selection algos solve the problem of 3. because 1. is already solved and 2. cant be solved yet as joinmarket miner fees are broken.
To me 3. doesn't seem like much of a problem.

CohibAA:
@chris-belcher That is correct, currently.  I, for one, would vote to include this code though, as I am one that find those small outputs annoying.  If I am not mistaken, having more and smaller outputs eventually contributes to more bandwidth usage and larger blockchain size over time, no?  I understand that is not an issue to be solved within scope of joinmarket, but we may as well not contribute to the issue.

@adlai Code is good so far, no issues with it over the last couple days.

chris-belcher:
OK yes @CohibAA it's true that combining together lots of small outputs will reduce the size of the UTXO set, which cant be pruned without some new fancy crypto. On the downside, combining together more outputs links them and provides evidence of common ownership to anyone looking at the blockchain. So I don't think it's a good idea to encourage it.

For dust of type 2. I've written about a possible solution here https://gist.github.com/chris-belcher/27b1743ba63e34736863

@tailsjoin whats your pycoin reallocation tool? Maybe tell me by email or start an issue about dust.

tailsjoin:
https://github.com/tailsjoin/tailsjoin/blob/master/extras/pycoin-reallocate.sh

adlai:
Outputs are only ever linked from within a single mixdepth; the current privacy model assumes that these are already linkable, so no privacy is lost by signing them into the same transaction. Dust must eventually be swept, and this approach allows some granularity over the manner in which this sweeping is done, although doing a single ```sendpayment``` with ```merge_algorithm = greedy``` replicates the "one-shot"sweeping behavior provided by all other tools.

Most importantly, default behavior is unchanged.

adlai:
Rebased onto latest master.

I still don't see why this shouldn't be merged: it doesn't change default behavior, and only merges within a single depth, leaving the privacy model intact.

the9ull:
I'd add that the default behaviour it's a factory of dust (when it is possible it always use 1 input and the output are 2). Why do not change the default behaviour?

adlai:
@the9ull I believe it's fairer towards users that new behavior is enabled through [opt-in rather than opt-out](https://shkspr.mobi/blog/2014/06/dark-patterns-opt-out-opt-in/).

MMMonsterrr:
If I understand the privacy model based on mixing depth correctly, then a higher mixing depth usually means a higher level of uncertainty regarding the origin of the utxos, right?

Wouldn't it then make sense to have different merge policies for different mixing depths? 

For example:
STRICT policy for depth 0     >    never merge outputs
DEFAULT        for depth 1-2  >    try not to merge, but merge if necessery
GRADUAL       for depth 3-4  >    gradually reduce utxos
GREEDY         for depth 5+   >    assume that there's enough uncertainty now to force merge of utxos

chris-belcher:
@MMMonsterrr that's somewhat true for the tumbler, depending on in which mixing level the user first deposited their coins. But for the yield generator which is what this thread is mostly about, the mixing levels are circular, after level 4 it goes back to level 0. So it probably makes no difference.

MMMonsterrr:
Oh okay, I didn't know that.

But even then, the yield generator would still work, even with these depth dependent policies. STRICT might be a bit annoying in that case though, so maybe the circular jump should go from 4 to 1 instead?

Anyway, was just a thought :)

PR #174, opened 2015-08-06
add ?scale=log to /ordersize

adlai:
Sneak Preview:
log-bucketed order size histogram


chris-belcher:
Looking good. I tested it. It's pretty simple so merging straight away.

PR #178, opened 2015-08-08
prevent sendpayment.py dos

CohibAA:
A poorly coded or malicious maker bot script may be able to cause all potential takers to hang when seeking available orders, with something like

    ValueError: invalid literal for int() with base 10: '0.1337'

This is done by making certain changes to how the maker order is announced to potential takers in IRC.

Change int to float instead possibly to avoid hanging?  Not sure what else this may affect or if this is the best solution, in any case, but this is a potential way to deny service to takers, so please review ASAP.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -335,7 +335,7 @@ class BitcoinCoreWallet(AbstractWallet):
 def calc_cj_fee(ordertype, cjfee, cj_amount):
 	real_cjfee = None
 	if ordertype == 'absorder':
-		real_cjfee = int(cjfee)
+		real_cjfee = float(cjfee)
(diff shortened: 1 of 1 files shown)


AdamISZ:
I think that's right, but it should be worth adding a try/except block around that code in case you receive something not parsable as a number at all, i.e. 

try:
   ...
except ValueError as e:
    print e
    do something to give up without crashing

etc.

CohibAA:
@AdamISZ You are right.  If anything other than a number is used by a yield-gen bot, it also seems to hang watchers along with takers.

CohibAA:
I imagine there is a better way to do this, but I have tested this and it works for sendpayment.py

Issue #179, opened 2015-08-09
ob-watcher hangs from bad maker data


CohibAA:
It is trivial to hang the ob-watcher.py script using a malicious or poorly coded maker bot, preventing the orderbook.html page from displaying until the bad order information is no longer showing on the !orderbook data.  Similar to https://github.com/chris-belcher/joinmarket/pull/178, probably another quick fix on ob-watcher.py exception handling will prevent this.

CohibAA:
closing, fixed by https://github.com/chris-belcher/joinmarket/pull/180

PR #180, opened 2015-08-09
validate cjfee in OrderbookWatch.on_order_seen

adlai:
resolves #178 and #179


ob-watcher.py
```diff
@@ -1,5 +1,4 @@
 import BaseHTTPServer, SimpleHTTPServer, threading
-from decimal import Decimal
```
adlai: In common.py we have ```from decimal import Decimal```, and a few lines down we have ```from common import *```, making the above import redundant.

PR #183, opened 2015-08-09
standardize satoshi_to_unit output

CohibAA:
cleans up the display of units returned for the orderbook.  only affects decimal output of fees and min/max size display.

see [this display with clean satoshi display](http://bitcoin-p2pool.com/joinmarket_orderbook.html) vs [this current style of display](http://joinmarket.io/)

edit, both of the above are now using (variations of) this PR.

### DIFF
--- a/ob-watcher.py
+++ b/ob-watcher.py
@@ -99,7 +99,7 @@ def cjfee_display(cjfee, order):
 		return str(float(cjfee) * 100) + '%'
 
 def satoshi_to_unit(sat, order):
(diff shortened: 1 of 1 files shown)


CohibAA:
related to [#134](https://github.com/chris-belcher/joinmarket/pull/134/files)

chris-belcher:
Looks good. I will test.

tailsjoin:
I'm using it here: http://ruc47yiosooolrzw.onion:62601/

PR #186, opened 2015-08-10
numpy removal

AdamISZ:
Removed usage of numpy module.

There are a set of of rand_* functions in common which replace the necessary functions. The only slightly complex ones are: first, power law distribution, achieved using the inverse of the cdf, see the wolfram link for the idea. I checked this by looking at sample arrays to see that the number of entries in different bins was approximately the same as that produced by the numpy version. Second, the random-choice-weighted-by-probabilities function, which is done by setting boundaries according to the cumulative array of probabilities and then choosing using a random sampled from a uniform dist. There are floating point inaccuracies here, but as noted in comments that's expected and considered unimportant. If later it's felt to be necessary to remove that, it can be done.

Testing: I'm going to produce a separate PR for the tumbler tests, which are now working, at least up to 6 yield generators against one tumbler. I've run this test many times and am not finding errors with this commit, but there's an important caveat: I had to make the change on line 151 of tumbler.py to refer to total_amount instead of amount, as otherwise I was occasionally getting negative change amounts.

In order to run those tests I had to make some small but fiddly configuration changes, but I'll explain that in the separate testing PR. Edit: see #187 . Added instructions in the wiki page 'Testing 2'.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -1,9 +1,9 @@
 
 import bitcoin as btc
 from decimal import Decimal, InvalidOperation
-from math import factorial
+from math import factorial, exp
 import sys, datetime, json, time, pprint, threading, getpass
-import numpy as np
+import random
 import blockchaininterface, slowaes
 from ConfigParser import SafeConfigParser
 import os, io, itertools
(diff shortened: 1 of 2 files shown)


AdamISZ:
Forgot to add: I didn't edit the README, seems that some edits will be needed in a few places to remove the reference to numpy.

chris-belcher:
Created tests for the new random functions. The power law deviates slightly from the numpy power law, but it doesn't really matter. It's probably just a vs a-1

choice_test
power_test

Issue #190, opened 2015-08-14
Wallet file hash changes.


tailsjoin:
I found something interesting today. I don't know if you remember, but one day in the irc I said that the hash on my wallet had changed out of nowhere, and it was asking me to rescan. Well I did rescan that day. For some reason, after the rescan the wallet hash went back to it's original and I moved on with my life. Well since then this same thing has randomly happened to me again and again, but I never rescanned. I just try three, four times until it gets right or I recover from seed.

Here's where things get interesting. I got a walletnotify twice today that my maker was not waiting on. The first came immediately after a JM tx so I ignored it, but the second came when all was quiet and it was for 1.5btc! So I opened up qt and saw that it was from this wallet that I accidentally created that day in the irc . So someone else is using this wallet and I'm watching all of their transactions currently.

![](https://infotomb.com/5tshc.png)

First I checked that it wasn't watching only funds I had forgotten about in an old wallet, but I have never deleted one and I have all the seeds I've ever created with JM. I painstakingly went through all of my wallet files to make sure  I'm not sharing a wallet with someone, and I definitely do not have that wallet hash. Maybe someone has a log from irc to check. Somewhere around the middle of July I think.

In playing with the wallets I found that if you enter an incorrect password for your wallet file sometimes it will accept the incorrect (or even blank on some wallet files) password and create a wallet using that as a seed. You can recreate it with the same incorrect password that worked and it will make the same wallet again. It will not make the original wallet unless you provide the correct passphrase, but sometimes even with the correct passphrase a new wallet is generated. I'm very sure of this because I use a password manager, and I don't even know what my passphrases are. I just copy/paste.

AdamISZ:
It seems like there's a lot going on here. With regard to:

> In playing with the wallets I found that if you enter an incorrect password for your wallet file sometimes it will accept the incorrect (or even blank on some wallet files) password and create a wallet using that as a seed. 

Do you ever use testnet? (I only ask because that's expected behaviour there). More importantly, which command are you running here? yieldgen, sendpayment, wallet-tool or tumbler?

AdamISZ:
@tailsjoin  Please see #191 . I believe this is probably a big contributory factor to what you saw, although there's a good chance it's not the total explanation.

tailsjoin:
I have seen the hash change on yield-generator and wallet-tool. Yesterday was wallet-tool. 

[...]

PR #191, opened 2015-08-14
ensure correct password

AdamISZ:
PKCS7 padding has the form: 01 or 0202 or 030303 etc. For a correctly entered password and a correct decryption, there will be a full block of padding at the end of the data: 1010... (16 times, hex). This obviously cannot happen by chance, which is why entering an incorrect password throws a ValueError which the current code then interprets as decryption failure and requests the user to try again.

However, in the case where an entered password is incorrect, it is possible that the decrypted data (*before* stripping the padding) ends with ..01. This would be interpreted as correctly pkcs7 padded, the 01 would be stripped, and a decrypted seed of length 32+15 = 47 bytes would be returned. No ValueError would be thrown and then a bip 32 master seed would be correctly (but randomly) generated. Note that the bip32 master key generation uses an hmac of the seed data, and so the seed data length is ignored.

So in summary we have a slightly bigger than 1 in 256 chance of a wrong password being accepted here, so this PR insists that the decrypted data has a length of 32 bytes, making this error impossible.

I don't consider this a big deal really, but it's as well to fix it of course. I was 'inspired' by #190 , but there seems to be no chance that this was the cause of the strange behaviour described there.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -218,7 +218,12 @@ class Wallet(AbstractWallet):
 			try:
 				decrypted_seed = slowaes.decryptData(password_key, encrypted_seed
 					.decode('hex')).encode('hex')
-				decrypted = True
+				#there is a small probability of getting a valid PKCS7 padding
+				#by chance from a wrong password; sanity check the seed length
+				if len(decrypted_seed) == 32:
+					decrypted = True
+				else:
+					raise ValueError
 			except ValueError:
 				print 'Incorrect password'
 				decrypted = False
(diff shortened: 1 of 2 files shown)


AdamISZ:
TLDR: The real situation is that there is a **1 in 8** chance of this happening, not 1 in 256!

The reason is that the slowaes module **does not properly verify the padding**, incredibly! It checks only the final byte, not the whole set of bytes. E.g.:

If the last byte is 04, it will immediately count it as valid. It will not check if the three preceding bytes are 04 04 04.

I have confirmed this (and confirmed my fix) by making a wallet with password 'abcd', and found that it will also accept 'fg' and 'tt' and so on. Creating, of course, a new random wallet with a different hash.

I have used the same code as I did in tlsnotary to check the pkcs7 padding is properly formatted. It now works correctly. Thus the previous comments no longer apply; this **is** fairly important.

I now believe this is probably *very* relevant to #190.

PR #193, opened 2015-08-15
default tor socks5_port consistency

CohibAA:
See https://github.com/chris-belcher/joinmarket/pull/182

### DIFF
--- a/ob-watcher.py
+++ b/ob-watcher.py
@@ -99,7 +99,7 @@ def cjfee_display(cjfee, order):
 		return str(float(cjfee) * 100) + '%'
 
 def satoshi_to_unit(sat, order):
-	return str(Decimal(sat) / Decimal(1e8))
+	return "%.8f" % float(Decimal(sat) / Decimal(1e8))
 
 def order_str(s, order):
 	return str(s)
--- a/test/commontest.py
+++ b/test/commontest.py
@@ -1,0 +1,72 @@
+import sys
+import os, time
+data_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+sys.path.insert(0, os.path.join(data_dir, 'lib'))
+import subprocess
+import unittest
+import common
+from blockchaininterface import *
+import bitcoin as btc
+import binascii
+import pexpect
+import random
+
+'''Some helper functions for testing'''
+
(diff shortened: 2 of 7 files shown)


CohibAA:
Also, do/will test scripts use freenode (without ssl)?

AdamISZ:
@CohibAA in the test/ directory there's a sample joinmarket.cfg, but you can edit the IRC settings to anything you like. If that's what you were asking?

CohibAA:
@AdamISZ No, I mean that in the current master test directory, the regtest_joinmarket.cfg (https://github.com/chris-belcher/joinmarket/blob/master/test/regtest_joinmarket.cfg) shows irc.freenode.net on a #joinmarket-pit channel.  I am curious if this should be updated to match the current IRC channel, or if the tests have some dependency on freenode IRC instead.

AdamISZ:
I guess the best I can say is - for ssl, sure, it should probably be set as default, including in the basic regtest. I will try that now (perhaps you already have).

Re; the channel, yes, as the wiki says, set it to anything you like, but remember about the -test suffix.

I don't think you can run the tests on cyberguerrilla because you're only allowed 2 (or 3?) connections. And I don't know if you can use a tor connection on freenode(?) I didn't pay attention to those aspects, but for sure I'd be interested to know if you can do it. It would clearly be a good thing if we can run tests via tor as well as clearnet.

AdamISZ:
@CohibAA I just ran a regtest.py with port 6697 and use_ssl=True on freenode and it was OK. So maybe add this to the PR, as well as whatever else you discover. No reason not to use ssl by default I think.

tailsjoin:
https://freenode.net/irc_servers.shtml#tor

CohibAA:
Freenode default SSL settings updated per https://freenode.net/irc_servers.shtml#ssl

FYI regrading tor on freenode:
> https://freenode.net/irc_servers.shtml#tor
> Accessing freenode Via Tor
> *** PLEASE NOTE THAT THE tor HIDDEN SERVICE IS CURRENTLY UNAVAILABLE AND THERE IS NO SET DATE FOR ITS RETURN ***

chris-belcher:
Needs a rebase

PR #205, opened 2015-08-29
"(potentially) earned" phrasing change

CohibAA:
if tx is not broadcast and confirmed on the network, no earnings are made.  "earned = " is more accurate after the tx is in a block.


adlai:
In an ideal patch, there'd be less overlap between the messages (to be precise: unique text in the confirmation, absent from the potentiation), saving the need for ```|grep -v pot``` to isolate confirmed earnings.

PR #207, opened 2015-08-30
Timing fixes for large/multiparty transactions

adlai:
Tested with 16 parties.

### DIFF
--- a/lib/irc.py
+++ b/lib/irc.py
@@ -10,8 +10,8 @@ import enc_wrapper
 
 MAX_PRIVMSG_LEN = 400
 COMMAND_PREFIX = '!'
-PING_INTERVAL = 180
-PING_TIMEOUT = 30
+PING_INTERVAL = 50              # lowered since we compete with data chunks
+PING_TIMEOUT = 340              # raised so we don't time out while sending data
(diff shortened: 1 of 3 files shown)


chris-belcher:
Looks good.

Is the line in taker.py fixing a race condition?

adlai:
Yes, there appears to be a race to build the transaction before keys are received.

chris-belcher:
I see, so it may happen that the taker sends !fill to several makers inside CoinJoinTX. The first maker responds with !pubkey before the CoinJoinTX method has returned and thus self.cjtx is null.

adlai:
Rebased, added comments

CohibAA:
I was recently kicked from the IRC channel (Orly? Whore!) as a taker using sendpayment while sending auths.  I was able to resolve by adding 

    time.sleep(1) #HACK! really there should be rate limiting, see issue#31

under [this line](https://github.com/chris-belcher/joinmarket/blob/master/lib/irc.py#L105).

Also, my maker was later kicked from IRC (Orly?) also while sending auths, so I added the same thing under [this line](https://github.com/chris-belcher/joinmarket/blob/master/lib/irc.py#L144) which seems to have resolved the kicks, afaict.

Issue #166, opened 2015-07-27
Crude taker recovery from unresponsive makers


chris-belcher:
A quicker and easier way of fixing the unresponsive maker problem than a fully-fledged internal reputation system like issue #57 

Could be a simple idea that if a maker doesnt respond the taker will message other makers to try to make the coinjoin happen.

If a maker doesn't respond before !tx has been sent, the taker can just !fill a new maker. If !tx has already been sent it has to start again.

A new thread is started for each CoinJoinTX object. There will have to be a callback function passed with can generate new makers to try, maybe just the chooseOrdersBy thing.
An issue might be that the required coinjoin fee has now changed.

PR #212, opened 2015-09-01
Crude maker-timeout recovery by taker

chris-belcher:
Implemented changes suggested in #166

A new thread is started for each CoinJoinTX object, there is a threading condition lock which passes notifications between the two threads. Currently the timeout for maker non-reply is 10 seconds.

If a maker doesn't reply, it will be ignored for the rest of lifetime of the Taker instance.

`sendpayment.py` and `tumbler.py` have been made to work with this.

Testing welcome. I've tested this on regtest with tumbler, it was able to complete it's run even though one maker never responded to either !fill, !auth or !tx

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -401,10 +401,11 @@ def pick_order(orders, n, feekey):
 		pickedOrderIndex = -1
 		
 
-def choose_order(db, cj_amount, n, chooseOrdersBy):
+def choose_orders(db, cj_amount, n, chooseOrdersBy, ignored_makers=[]):
 	sqlorders = db.execute('SELECT * FROM orderbook;').fetchall()
 	orders = [(o['counterparty'], o['oid'],	calc_cj_fee(o['ordertype'], o['cjfee'], cj_amount), o['txfee'])
-		for o in sqlorders if cj_amount >= o['minsize'] and cj_amount <= o['maxsize']]
+		for o in sqlorders if cj_amount >= o['minsize'] and cj_amount <= o['maxsize'] and o['counterparty']
+		not in ignored_makers]
 	counterparties = set([o[0] for o in orders])
 	if n > len(counterparties):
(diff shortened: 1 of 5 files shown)


chris-belcher:
Also, why does the *merged latest master....* commit seem to have a ton of unrelated changes that were going on in parallel to this pull request. I assume a rebase would fix this?

chris-belcher:
Best to revert the merge, a pull request is a merge after all

Also, the lock.acquire() lock.notify() lock.release() pattern can be simplied using the python keyword with.

adlai:
0f4b954 is a spurious merge, it adds no new functionality but confuses history by causing Github to display all the history that wasn't in 84a0859 as its own diff. Merging a "clean" (ie, fast-forward) pull request through Github's web interface will create a "merge commit" that doesn't do this.

tl;dr: Please rebase :smile:

MechanicalMink:
I was told on Reddit to try to rebase this manually, since I have no experience with version control software. I added and removed the code as per these changes to the latest available build, and every time I tried to run the tumbler it didn't seem to like something to do with this line in taker.py

self.utxos = {None: self.input_utxos.keys()} #None means they belong to me

Evidently we're trying to access the keys of a NoneType object somewhere..? This would always result in a semi-crashy-loopy-thing when the output hit the following:

started timeout thread for coinjoin of amount xxxxxxx to addr xxxxxxx
[2015/09/06 04:43:24] waiting for all replies..

Python would then repeatedly try to run a traceback, throwing an attribute error that said something along the lines of "NoneType objects don't have keys". I hope this is helpful, as I only have experience in Java at the moment and I really don't know this codebase or what I'm talking about in the least...

chris-belcher:
The git commit tree is broken on it. I stupidly merged it into master and then created a PR. Fixing this git problem is my first priority.

chris-belcher:
It was easier to create a new branch rather than fix this one. Apparently this kind of screw up is hard to recover from.

PR #221, opened 2015-09-05
Bugfix: socks5 negotiation failure reporting

### LINKED ISSUE #220: crash when irc hidden tor service unavailable
When the IRC's hidden service (6dvj6v5imhny3anf.onion in this event) becomes unavailable for whatever reason (tor log below)

>    [notice] Closing stream for '[scrubbed].onion': hidden service is unavailable (try again later).

the IRC will cause a crash.  This crashes yield-gen bots, watcher bots, and likely tumbler bots (but I have not confirmed tumbler personally).

This is the crash shown in the yield-gen console:

>    line was zero length
    disconnected irc
    connecting
    Using socks5 proxy 127.0.0.1:9050
    CRASHING, DUMPING EVERYTHING
    ...
      File "ygen-flat.py", line 213, in main
        irc.run()
      File "/home/cohibaa/joinmarket/lib/irc.py", line 476, in run
        self\.sock.connect(self.serverport)
      File "/home/cohibaa/joinmarket/lib/socks.py", line 369, in connect
        self\.\__negotiatesocks5(destpair[0],destpair[1])
      File "/home/cohibaa/joinmarket/lib/socks.py", line 236, in \__negotiatesocks5
        raise Socks5Error(ord(resp[1]),\_generalerrors[ord(resp[1])])
    TypeError: \__init__() takes exactly 2 arguments (3 given)

and the same event from a watcher bot:

>    line was zero length
    disconnected irc
    connecting
    Using socks5 proxy 127.0.0.1:9050

Perhaps something around [this line in socks.py](https://github.com/chris-belcher/joinmarket/blob/master/lib/socks.py#L236) or maybe something else in irc.py can be changed to have the bots wait for a period of time then re-attempt the connection if the connection attempt fails at this point.

adlai:
Fixes #220 (thank you @CohibAA for reporting it!)

Appears to have been an upstream bug, which each upstream fork has fixed differently; this fix is ```min```imal.

### DIFF
--- a/lib/socks.py
+++ b/lib/socks.py
@@ -232,10 +232,7 @@ class socksocket(socket.socket):
 		elif resp[1] != "\x00":
 			# Connection failed
 			self.close()
-			if ord(resp[1])<=8:
-				raise Socks5Error(ord(resp[1]),_generalerrors[ord(resp[1])])
-			else:
-				raise Socks5Error(9,_generalerrors[9])
+			raise Socks5Error(_socks5errors[min(9,ord(resp[1]))])
 		# Get the bound address/port
 		elif resp[3] == "\x01":
 			boundaddr = self.__recvall(4)


chris-belcher:
This fixed a bug that happens in Section 6. Replies

https://www.ietf.org/rfc/rfc1928.txt

It's an odd bug, the constructor only takes one argument but they gave it two.

Has someone tested this just in case? Try it on one of the broken .onion links on the hidden wiki.

PR #222, opened 2015-09-05
Delete leftover print statement in select_greedy

adlai:
oops :disappointed:


CohibAA:
Maybe [this](https://github.com/chris-belcher/joinmarket/blob/master/lib/common.py#L236) should also be updated with this PR?

https://github.com/chris-belcher/joinmarket/commit/e12a70e58d7aa0c35c1a3e028cfe5170274f116d#commitcomment-13083058

chris-belcher:
Agree with @CohibAA, someone on reddit appeared already to be mislead by the "needs a Policy" message.

Issue #213, opened 2015-09-02
yield-generator fails if unconfirmed transaction isn't seen


dooglus:
commit 850e8194ecc88a60bb4288ec3704678eb5172cc7
Author: chris-belcher <chris-belcher@users.noreply.github.com>
Date:   Mon Aug 31 19:39:36 2015 +0100

My wallet never saw the unconfirmed join transaction before it saw it in a block.

on_tx_unconfirmed() didn't run, but on_tx_confirmed() did.

on_tx_confirmed() tried to access self.tx_unconfirm_timestamp[cjorder.cj_addr] but it hadn't been set, and so an exception occurred:

    [2015/09/01 20:50:19] <<pubmsg nick=Meyofawic message=!absorder 0 10000 223725421 10000 10000 ~
    [2015/09/01 20:54:27] rpc: getrawtransaction ['4f74d2667542988c1e5cf2535d503b185124c1bfdd4c717421e92c7c5dfd9f98']
    [2015/09/01 20:54:27] rpc: gettxout ['4f74d2667542988c1e5cf2535d503b185124c1bfdd4c717421e92c7c5dfd9f98', 0, True]
    [2015/09/01 20:54:27] rpc: listunspent []
    [2015/09/01 20:54:27] bitcoind sync_unspent took 0.0365281105042sec
    [2015/09/01 20:54:27] tx in a block
    [2015/09/01 20:54:27] earned = 0
    ----------------------------------------
    Exception happened during processing of request from ('127.0.0.1', 57880)
    Traceback (most recent call last):
      File "/usr/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
        self.process_request(request, client_address)
      File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
        self.finish_request(request, client_address)
      File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
        self.RequestHandlerClass(request, client_address, self)
      File "/home/user/Programs/joinmarket/lib/blockchaininterface.py", line 326, in __init__
        BaseHTTPServer.BaseHTTPRequestHandler.__init__(self, request, client_address, base_server)
      File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
        self.handle()
      File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
        self.handle_one_request()
      File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
        method()
[...]

PR #226, opened 2015-09-08
a solution for #213: ygen unseen pending tx crash

CohibAA:
possible solution for https://github.com/chris-belcher/joinmarket/issues/213

I've tested as working and preventing the crash, but I'm not sure how sound this logic is, please review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 *.swp
 joinmarket.cfg
+blockchain.cache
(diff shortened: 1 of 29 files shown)


chris-belcher:
Catching an Exception to detect this error isn't a good idea. Exception is a very general class and can happen for many reasons.

Why not just use a new boolean flag, on_tx_unconfirm_called and then just before you call on_tx_confirmed() check it first

    if not on_tx_unconfirm_called:
        on_tx_unconfirm()
    on_tx_confirm()

adlai:
same issue is being addressed in #438

PR #232, opened 2015-09-14
Update readme and change .txt to .md

tailsjoin:
- Copy/paste brief explanation of what JoinMarket is from wiki. Should be quickly explained on the landing page.

- Change install dependencies from numbered list to bullets. With two numbered lists the steps look a bit daunting at first glance.

- Added 4th step since we've had a few new people recently not realize that the config is populated on first run, causing some confusion. I believe this also eliminates the need for the note near the end about putting mainnet in the config.

- Reformatted already existing information.


tailsjoin:
Consolidated `python` and `libsodium-dev` apt-get to same step as well.

dcousens:
LGTM

PR #241, opened 2015-09-24
Implemented importing into bitcoin core of newly generated addresses

chris-belcher:
Completely untested by me.

Bitcoin-Qt's option of exporting all receive addresses will be useful for testing.

To test, import a joinmarket wallet into bitcoin core, export the addresses, run yield-generator or tumbler until the HD wallet increments enough that you need to import a new address. Might be helpful to reduce the number of addresses imported down from 50.

### DIFF
--- a/lib/blockchaininterface.py
+++ b/lib/blockchaininterface.py
@@ -422,6 +422,7 @@ class BitcoinCoreInterface(BlockchainInterface):
 
 		self.notifythread = None
 		self.txnotify_fun = []
+		self.wallet_synced = False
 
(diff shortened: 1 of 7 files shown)


chris-belcher:
Tested a bit. Yield generator with this correctly imports addresses and uses them.

Running wallet-tool still imports a large amount of new addresses because it checks 50 addresses at once and yield generator only imports one at a time. Need to edit the message to explain that you only now need to -rescan when recovering an old wallet from seed, all other times just restart the joinmarket script.

PR #245, opened 2015-09-28
orderbook.html: select specific orders with a single click

the9ull:
I added JS code to highlight lines of the orderbook table. This gives a quicker overview over the interested orders (e.g. your own yield generator).
The selection it's persistent.


chris-belcher:
Looks nice.

PR #246, opened 2015-09-30
Move yieldgen configuration to joinmarket.cfg

abitfan:


AdamISZ:
It seems like nobody's looked at this?

It seems like an entirely sensible idea, and simple. But: did you notice that we have a required_options dict at the top of common.py, so you could enter these fields into the joinmarket.cfg file, and put it in the default version of course. Then you wouldn't have to load the fields individually. Of course, these fields would just be ignored by the taker scripts.

Also, squash commits into 1 please.

raedah:
We have 3 yeildgens in develop now. Organizing the config still needs to be done, but will need to be done in a different way which takes into account options which are shared across all yieldgens and options that are specific that that yieldgen. I suggest this pull request is closed.

PR #262, opened 2015-10-08
tumbler option to start from lowest used mix depth

weatherhead99:
got hit by this a couple of times when using tumbler myself.
Naive and simple implementation, but seems to work here.


chris-belcher:
Thanks for implementing this.

It needs a small fix, search for `options.mixdepthsrc` in the entire tumbler.py file and you'll see its used in other places. I think it's best to just reassign it as in `options.mixdepthsrc = min_depth`.

Also, I'm thinking it might be better to not even bother asking the user for confirmation(?) It adds more mental space requirement in their heads, and I can't think of any downside except maybe that if they're running with bitcoin core it will say "imported addresses, use -rescan if you're recovering from seed".

Finally, please use git rebase so that this PR has only one commit. It makes it easier to follow development.

weatherhead99:
The only downside I can think of is if the user is perhaps re-starting a multi-stage coinjoin that got some partway through. e.g. if I want to tumble 5 times, I start tumbler.py asking for 5 tumbles. Then if it dies or gets killed after 3 tumbles, I actually want to start again with only 2 tumbles from the last level. The current implementation will try to do another 5.

Sorry about rebase, I pushed to my repo by accident before doing it!

chris-belcher:
Ah that's a different feature.
I think the best way to do that is to save the 'transaction list' and other information to a file, and allow someone to give that file to tumbler.py on the command line when they want to resume.

weatherhead99:
agree completely that it's a separate feature in principle, just that at the moment, if my tumbler is killed, when I restart it, it will stop with "no coins remaining at level 0", and that will remind me to change the level. Wheras if we don't ask with this fix, then a tumbler that needs to be restarted often might waste a lot of time.

chris-belcher:
The drawback you mention here https://github.com/chris-belcher/joinmarket/pull/262#issuecomment-146582441 already exists in your PR.
So maybe the best solution is to simply `sys.exit(0)` with the message like "no coins in mixing depth XXX, use -m to choose another mixing depth to start from"

weatherhead99:
I guess as a minimum fix to both problems, it might be useful for tumbler to drop a short text file on startup, and delete on successful exit. Then we can also check on startup (in a fairly hacky way) whether the last run went to completion.

chris-belcher:
Yes thats a good idea. It's probably better to call it a "tumbler progress file", ask the user for a name and allow the user to pass that name to tumbler.py if they want to restart where they left off.
It's much less hacky than the script checking if the file exists on its own without the user's knowledge.

Issue #197, opened 2015-08-18
Creation of multiple copies of orders can skew probabilities


AdamISZ:
It's a well understood issue with the system that offers placed in the orderbook by makers are not enforced to be real. This is connected with the oft-discussed related issues of whether we want to use persistent identities, and whether snooping to gain info on utxos might link join outputs( #156 ). The connection between the two being that utxos have to be *announced*; if they're announced "early" we lose privacy, if they're announced "late" (as they are today) we regain privacy but have no certainty that offers represent real coins.

I'm not attempting to answer all these questions here, but that just gives context to the following:

The [choose order code](https://github.com/chris-belcher/joinmarket/blob/master/lib/common.py#L409-L429) leaves open a possibility of skewing the probability of selection in favour of a particular maker. It ensures that each chosen order is from a separate counterparty, but only after  having chosen the order from the full list via the chooseOrdersFunc. Thus, if a maker just announces multiple copies of each of his orders, any probabilistic choice will favour him for doing so (e.g. weighted_orders_choose).

If the code were altered to prevent this, there are a few decisions that need to be made about how to do that. Rejecting identical orders is a trivial one, but equally trivial coding by the antisocial maker could sidestep it. A hard limit on the number of orders per maker is another possibility.

However, even if this change was made, the basic problem remains: it only requires a little ingenuity on the part of an antisocial maker to run multiple bots off the same wallet, perhaps disguising the orders to make them look different. It doesn't work "out of the box" for sure, but it equally clearly can be done. I don't think we can "fix" this without revisiting the issues mentioned in the first paragraph.

chris-belcher:
Yet another way to fix this would be to only consider one order per maker in the weighted probability chooser.

This is kind of how it works now (though voluntarily, in a way that can be gamed) where the publicly released yield generator algorithms have min/max sizes that dont overlap, so the weighted_choose_order code only ends up with one-order-per-maker for any cj_amount.

AdamISZ:
Rather late revisit, but anyway..

> Yet another way to fix this would be to only consider one order per maker in the weighted probability chooser.

Well, yes, I said that "A hard limit on the number of orders per maker is another possibility." And I agree that this is the most practical first cut decision. But the other concern, the deeper one, remains (see last paragraph and, I guess, first).

I'm revisiting it because looking at the orderbook I see at least one (and prob. more) makers gaming the system this way. I might try to put together a PR to enforce one order consideration per maker.

[...]

PR #266, opened 2015-10-10
consider only 1 order from each counterparty before choosing

AdamISZ:
I spent some time thinking about the right way to handle the issue in #197.

From one angle it looks like a very small problem that's being solved here, because the code already enforces one order being accepted from each counterparty. But as noted, the probabilities are skewed in case the choice algorithm chosen is probabilistic, because the weighting is done before that enforcement. This code uses a Python one liner to take only one order from each counterparty - the one with the lowest calculated coinjoin fee - from each counterparty before doing any selection algorithm. I put an if/else in case someone wants to make a manual choice.

(Btw in case you're wondering, reverse=True *is* correct here - the reason is a bit obscure).

I've done some test runs on a modified branch using new_yieldgen style yieldgen bots, further modified to artificially create multiple orders per mixdepth. I'm pretty sure it's functionally correct.

However, I can well imagine there may be further discussion about the *right* way to do this. Right now, I believe this would be a meaningful improvement since there are already bots running in the pit that duplicate orders over the same amount range.

 And of course it doesn't address the deeper problem discussed in #197.

### DIFF
--- a/lib/common.py
+++ b/lib/common.py
@@ -532,7 +532,14 @@ def choose_orders(db, cj_amount, n, chooseOrdersBy, ignored_makers=[]):
 		debug('ERROR not enough liquidity in the orderbook n=%d suitable-counterparties=%d amount=%d totalorders=%d'
 			% (n, len(counterparties), cj_amount, len(orders)))
 		return None, 0 #TODO handle not enough liquidity better, maybe an Exception
-	orders = sorted(orders, key=lambda k: k[2]) #sort from smallest to biggest cj fee
+	#restrict to one order per counterparty, choose the one with the lowest cjfee
+	#this is done in advance of the order selection algo, so applies to all of them.
+	#however, if orders are picked manually, allow duplicates.
+	if chooseOrdersBy != pick_order:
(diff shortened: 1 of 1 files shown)


AdamISZ:
Follow up from IRC:

Applying a similar logic to choose_sweep_order is problematic. Since there's no change, and we don't know in advance the coinjoin amount, we can't discard all-but-one of the orders from a single counterparty before going through the choice algorithm. I'm not sure there's a way to enforce this without something kludgy that will fail in edge cases (i.e. occasionally result in a 'insufficient liquidity' error where it needn't). I would argue we should apply this PR as is, and then address the sweep case to whatever extent possible later.

chris-belcher:
What if all the orders from one counterparty were bundled into one object that had it's own probability weighting. Then if it's chosen pick an order from the bundle.

raedah:
So if an offer A has a cjfee that is 1 satoshi cheaper then offer B, but offer B has a txfee of 20,000 satoshis, offer A is what will be passed to the taker as the best offer.

Issue #263, opened 2015-10-08
yield-generator crash if blockr returns vout without is_spent


ande2101:
[2015/10/08 06:00:21] Traceback (most recent call last):
      File "yield-generator.py", line 147, in main
        irc.run()
      File "/home/user/joinmarket/lib/irc.py", line 502, in run
        self.__handle_line(line)
      File "/home/user/joinmarket/lib/irc.py", line 405, in __handle_line
        self.__handle_privmsg(chunks[0], chunks[2], get_irc_text(line))
      File "/home/user/joinmarket/lib/irc.py", line 358, in __handle_privmsg
        self.__on_privmsg(nick, parsed)
      File "/home/user/joinmarket/lib/irc.py", line 262, in __on_privmsg
        self.on_seen_tx(nick, txhex)
      File "/home/user/joinmarket/lib/maker.py", line 213, in on_seen_tx
        self.active_orders[nick].recv_tx(nick, txhex)
      File "/home/user/joinmarket/lib/maker.py", line 81, in recv_tx
        goodtx, errmsg = self.verify_unsigned_tx(self.tx)
      File "/home/user/joinmarket/lib/maker.py", line 128, in verify_unsigned_tx
        input_utxo_data = common.bc_interface.query_utxo_set(list(tx_utxo_set))
      File "/home/user/joinmarket/lib/blockchaininterface.py", line 314, in query_utxo_set
        if vout['is_spent'] == 1:
    KeyError: 'is_spent'

ande2101:
Update: This could have been because I was using it over Tor and it was using HTTP rather than HTTPS. I've put additional logging in and will see if it goes away.

tailsjoin:
FWIW I've used blockr with Tor extensively over the past months and have never seen this issue.

ande2101:
Interesting. I haven't restarted in ages, perhaps I've just got a bad exit node?

I had a similar issue yesterday before I put the HTTPS changes in:

    Traceback (most recent call last):
      File "yield-generator.py", line 147, in main
        irc.run()
      File "/home/anonymous/Documents/cash/joinmarket/lib/irc.py", line 502, in run
        self.__handle_line(line)
      File "/home/anonymous/Documents/cash/joinmarket/lib/irc.py", line 405, in __handle_line
        self.__handle_privmsg(chunks[0], chunks[2], get_irc_text(line))
[...]

PR #267, opened 2015-10-10
Use HTTPS rather than HTTP

### LINKED ISSUE #264: Use HTTPS for API calls
Use HTTPS calls for blockchian, blockr.io, etc. inside:

> lib/blockchaininterface.py
> lib/bitcoin/bci.py

ande2101:
Use HTTPS rather than HTTP for API calls to blockr.io and blockchain.info

Fixes #264

### DIFF
--- a/lib/bitcoin/bci.py
+++ b/lib/bitcoin/bci.py
@@ -46,7 +46,7 @@ def bci_unspent(*args):
     u = []
     for a in addrs:
         try:
-            data = make_request('http://blockchain.info/unspent?address='+a)
+            data = make_request('https://blockchain.info/unspent?address='+a)
         except Exception as e:
             if str(e) == 'No free outputs to spend':
                 continue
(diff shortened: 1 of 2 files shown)


chris-belcher:
So has anyone tried this yet?

I remember doing something like this many months ago but seem to remember it didn't work. Also you should all know the python urllib library doesn't check the SSL certificate.

ande2101:
I've been running yield-generator for a few days now and that seems okay. My failures are now related to DNS and connection issues to IRC, I haven't seen issue #263 since so it must have been a naughty exit node. Also the SSL certs are checked as of [Python 2.7.9](https://www.python.org/downloads/release/python-279/)

I haven't done a shared send or run the tumble script, mostly because I'm cheap. I did check that the HTTPS addresses worked though.

chris-belcher:
The IRC network has an onion so you don't have to go through exit nodes.

Has your yield generator done any successful joins?

ande2101:
Yep I've successfully done joins using this, also I've removed the other commits.

Cheers

Ande

PR #272, opened 2015-10-13
Specify default donate percent, tumber.py --help

tailsjoin:


tailsjoin:
I think users deserve to know the default amount of their funds being diverted. If not, should we remove all information about default settings then? 

There are only two options in `tumbler --help` without defaults specified. One is mixdepthcount (I'm assuming because it's all of them and that's common sense?) and the other is the donation amount. In contrast, there are 13 options with their defaults specified.

chris-belcher:
It's not clear to me what users deserve. They already get this software for free and derive a large amount of utility from it.

How to funding shared infrastructure is an unsolved problem. Hive wallet is one of the latest to halt development https://www.reddit.com/r/Bitcoin/comments/3llr1a/hive_wallet_no_longer_being_developedtime_to_bail/cv7mpbu?context=1

tailsjoin:
This is starting to feel scammy.

adlai:
@tailsjoin There's nothing wrong with developers asking for their time to be compensated. If you think the default policy here is unfair, put a different one in your fork.

AdamISZ:
@adlai but that's not what's at issue here, as far as I understand. It's about defaults and visibility. A default to donate is absolutely fine, but it really needs clear communication. I guess the only thing left to argue is what constitutes "clear communication".

tailsjoin:
@adlai This PR does not try to remove donations, does it?

AdamISZ:
@chris-belcher why shouldn't the donation default amount be shown here? It certainly should. I see no reason not to merge this.

The other stuff we've been arguing about is a separate matter (including what the default should be), but I can't see any possible controversy here.

adlai:
ACK

lacksfish:
As the default donation is now 0 percent, maybe we can close this pull request and get along again.

On a sidenote, I also didn't like how this issue was handled and, because of using and testing the tumbler, I donated more to @chris-belcher than I would've liked..

tailsjoin:
I just assumed that this was closed already.

Closing now as fixed elsewhere.

PR #289, opened 2015-10-22
Dust threshold for Bitcoin v0.11.1 new default

tailsjoin:


chris-belcher:
Merged now to stop bots breaking, in the medium term we should understand the limit better as I described in the issue about this.

Issue #291, opened 2015-10-24
secp256k1


AdamISZ:
I'm opening this issue to keep a log/update of progress on libsecp256k1 integration.

The branch is at https://github.com/AdamISZ/joinmarket/tree/secp256k1 .

An overview: 
First, the motivation: we need to use the best secure, tested, maintained, and conforming-to-standards code for Bitcoin signing and key management; in particular, the code for **elliptic curve calculations and ECDSA operations**. Hence the desire to use this [library](https://github.com/bitcoin/secp256k1). However, this code does only that, it does not replace the code for transaction creation/manipulation, or for BIP32 (the HD style of wallets used by joinmarket).

So in this branch, all EC and ECDSA operations are moved to the module ec_ecdsa.py, which is a wrapper for the [this](https://github.com/ludbb/secp256k1-py) python binding of the libsecp256k1 shared library. I've submitted a PR to add exposure of needed functionality [here](https://github.com/ludbb/secp256k1-py/pull/3).

This will allow us, hopefully, to no longer use snapshots of nacl or secp256k1 (the latter is currently added into this branch) and use pip install instead. For pybitcointools, this will no longer be relevant since the library has been cannibalised.

Further in line with improving security/testing, I've added the bip32 test vectors to the library [here](https://github.com/AdamISZ/joinmarket/commit/dab7c3fa3ff8d445cec45cf7d1bdb9ae3f1c24f0) (they passed).

There is other irrelevant code that could be moved from this version of bitcoin/ , for example blocks.py and composite.py (?). Opinions welcome, but not important.

I'm going to log the results of testing here. I plan to merge current master of joinmarket into this branch today.

AdamISZ:
Tests so far:

Basic test/regtest.py : passing. ("Basic" means the version currently in master; I have a more sophisticated version with more coins in more coin depths, randomised, and more bot participants which can get merged sometime).

test/wallet-test.py : passing. (These tests for now are mostly about password and seed recovery. I want to add more, especially e.g. private key dump).

test/tumbler-test.py : passing.

The above 3 tests are of course running of a regtest blockchain.

I have also just tried a 3 bot test (2yg, 1sp) on testnet using blockr. Also passed.

I'll add to this when I have more. There are quite a lot of 'corner use cases' nowadays like using multisig, using Core wallet features. If you can add to this or help me test, please do.

AdamISZ:
A note about installation of secp256k1 before I forget:
[...]

PR #302, opened 2015-11-02
Secp256k1

AdamISZ:
See #291 . 
Squashed most of the commits down.


AdamISZ:
Ugh, sorry I thought this would only create 6dd408a and 4a344f5 and the final merge 4882165

AdamISZ:
Will try again.

PR #303, opened 2015-11-03
Issue291

AdamISZ:
#291
Slightly cleaned up the git commits, hopefully.
Also, noticed another fix needed for legacy message signatures: new version only verifies low S, so convert all high S to low S in the ecdsa_verify function.

As I may have mentioned before, the install.sh in this is rudimentary, a much more complete version is being prepared and can be added any time.


AdamISZ:
e0a1f08 is a bugfix which addresses the rare case where the mess. sig has r or s values < 32 bytes. Tested OK on regtest and blockr.

AdamISZ:
arubi has done the install scripts for Linux, we are testing on Ubuntu, Debian and Tails. Obviously there is likely more to be done there, but no need for it to be a barrier to a future merge.

AdamISZ:
The above commit removes repeated calls to secp256k1_context_create in the constructor, see the notes in this comment: https://github.com/chris-belcher/joinmarket/issues/316#issuecomment-156867422 . This finally creates the performance boost I was expecting from the start.

PR #309, opened 2015-11-06
Don't print IRC password to the log when doing a debug_dump_object

MagikSquirrel:
It's a bad practice to print the password in plaintext to the log.


lib/common.py
```diff
@@ -177,6 +177,8 @@ def debug_dump_object(obj, skip_fields=[]):
 	for k, v in obj.__dict__.iteritems():
 		if k in skip_fields:
 			continue
+		if k == 'password' or k == 'given_password':
+			continue
```
dcousens: why not just add this to `skip_fields`?

PR #311, opened 2015-11-07
unmask sybil counterparties by hostmask

adlai:


adlai:
Inexplicable weirdness is happening with `IRCMessageChannel.__privmsg`. Closing the public PR until this actually works properly.

chris-belcher:
What kind of weirdness ?

Issue #254, opened 2015-10-02
Changing tabs for spaces and structural updates


chris-belcher:
@ghtdak has used a tool to change tab indentation for spaces without breaking git blame. As well as some other stuff, like changing lib/ into joinmarket/ and making it a python package instead of it being added to path.

https://github.com/ghtdak/joinmarket/commits/development

Apparently this gets combined into the repos with `git branch -f master master-yapf`

ghtdak:
I'm writing up what I did in the project which really doesn't exist. I had to name it though, so it now has a wiki

https://github.com/ghtdak/gitreformat/wiki

ghtdak:
The code at 
https://github.com/ghtdak/joinmarket/commits/development 
is both a history / blame preserving yapf reformat and a fix of the imports including the creation of the joinmarket module, replacing "lib"

All the path manipulating python is replaced in favor of the usual standard module inclusion.

AdamISZ:
This has been addressed by #340 and #345.

Issue #287, opened 2015-10-21
Takers should choose orders by another metric not just price


chris-belcher:
To reduce possible sybil attacks, incentivize new coins being brought into the anonymity set, and increasing the chances of new makers being chosen for each coinjoin, I propose that takers somehow choose orders based on height, bitcoin-days-destroyed and maybe size in bytes.

There's a few ways this might work in practice.

One is the makers announce a list of bitcoin amounts, heights/confirmations, bytes, bitcoin-days-destroyed, priority, etc

So the yield generator algorithm could have a range (in log space) of bitcoin-days-destroyed or heights and announce the prices for each one. The taker checks at some point whether the send UTXOs match the maker's claims.

In practice the protocol would be modified with a 'soft fork'. Creating a new kind of order.

Another way is that the taker sends a !fill order to every single maker in the orderbook and simply throws away the unsuitable UTXOs. I think this is a bad idea for many many reasons.

adlai:
We should adopt a flexible approach, letting takers each configure their own prioritization weightings, while provide sane (or even beneficial) defaults (similar to what we've done with [merge policy](https://github.com/chris-belcher/joinmarket/wiki/Configuring-with-joinmarket.cfg#input-merging-policy)). Prioritizing height (eg, *[volume-weighted? [geometric] mean? ...]* average number of confirmations on contributed inputs) or "Bitcoin days destroyed" (total of each input's size multiplied by its confirmations) incentivises adding new coins to, and thus increasing, the anonymity set, aligning incentives with the common good rather than the current perversion of shuffling about the money already invested in a slightly more "fungible" manner.

We've already seen JoinMarket's incentive start to twist: before #266, the simplest way to boost your volume was redundant orders; today, it is running multiple yield generators. This either fragments your wallet (reducing the maximum output size, eroding the market's usefulness to transaction initiators), or requires you to operate multiple bots off the same wallet, running a risk of duplicated UTXOs in a single transaction (causing failed transactions, and alerting The Customer to twisted game afoot).

>                             (for readability, I've converted these values to BTC, satoshi and percent)
>     [2015/10/22 07:11:49] <<pubmsg nick=Tedokadix message=!relorder 0 100000  45.24945649 0 0.04999%
>     [2015/10/22 07:11:50] <<pubmsg nick=Topihuruv message=!relorder 0 100000  51.38726145 0 0.00799%
[...]

PR #313, opened 2015-11-08
unmask sybil counterparties by hostmask

adlai:
Working (tested) version of #311.

I'm not certain what caused the heisenbug with `IRCMessageChannel.__privmsg()`; the symptom was that it didn't strip hostmask, and didn't execute debug print statements when I tried to figure out why; yet print statements did indicate that it was running changed code. Makes me itch for modularization à la #254... any update from @ghtdak on regarding https://github.com/ghtdak/joinmarket/tree/master-refactor_v1 ?

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -22,13 +22,14 @@ import old_mnemonic, slowaes
 
 parser = OptionParser(usage='usage: %prog [options] [wallet file] [method]',
 	description='Does useful little tasks involving your bip32 wallet. The'
-	+ ' method is one of the following: display- shows addresses and balances.'
+	+ ' method is one of the following: display - shows addresses and balances.'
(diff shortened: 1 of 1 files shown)


tailsjoin:
Maybe I don't understand this PR, but just on a quick check of the room `joinmarket-pit`:

    81 total
    22 users with the hostmask: CgAn-7g5.05e.aubm9g.IP

So it seems this is easier to fake, just use Tor?

adlai:
@tailsjoin The ideal solution is being fleshed out in #287, this is a short-term fix which gives sybils an incentive to get new IPs. If the easiest way for them to do so is Tor, then that's what they'll do.

tailsjoin:
Understood, thanks.

adlai:
* commit message summarizes rationale
* fixed bug which was due to omission of stripping from `announce_orders`

Issue #156, opened 2015-07-20
Bad faith taker spy not filling orders so that it learns which UTXOs belong to which maker, allowing future unmixing


chris-belcher:
Here is a way a malicious taker could unmix every coinjoin.

- Sit in the marketplace and partially-fill every order with the maximum amount. Stop responding once it learns the UTXOs that belong to each maker.
- Every time a broadcast tx containing a UTXO is seen, repeat the process to get everyone's UTXOs again.
- Use the knowledge to unmix every coinjoin.

Essentially this works by breaking the assumption that the taker cares about privacy.

Firstly this does nothing for passive observers of the blockchain.

It would be quite easy to detect this and raise the alarm. Look out for repeated unfilled orders for coinjoin amounts equal to your maxsize.

It could be defeated by radially changing the protocol to involve something like coinshuffle which would result in no single party knowing the mapping between inputs and outputs.

A third way is to write the yield generator algorithm so that it's very hard to find all the UTXOs.

AdamISZ:
Yes. 

Does Coinshuffle address it? I think not; if you're able to figure out that all-bar-1 of the join outputs are from makers, it doesn't matter what order they're in.

Just dropping the note here now, but continuing to think about it.

AdamISZ:
Forgive me in advance if this is a bit rambling, but I want to record what I understand about it right now.

The fundamental issue is that makers **announce** their utxos to anyone who (pretends to) start a transaction with them; whether this can be avoided without creating hideous DOS issues I don't know.

Coinshuffle does not help, since its only purpose is to make the ownership of all the coinjoin outputs hidden from *all* participants (thus avoiding the current situation where there is a single coordinator, namely the taker, knowing all the linkages). This attack, to the extent that it works, sidesteps that.

Chaumian blinding does not help either. You can think of it as a lower-tech version of Coinshuffle (because, to achieve the same result, it requires the additional step of breaking the network linkage between input and output, i.e. you have to anonymously reconnect during the protocol, which is .. icky).

Possible Mitigations:
1. Punishment for non-completion: a taker who does not respond after receiving !ioauth has his btc utxo blacklisted (and remember he had to prove ownership of it in the !auth message before getting the information). This is quite harsh considering people can have connectivity problems, and also not very effective considering how easy it is to create a new utxo with a trivial transaction. But at least, it is not totally costless.
[...]

PR #328, opened 2015-11-18
initial blacklist edits

AdamISZ:
DO NOT MERGE.

(needs rebasing anyway :) )

Idea: every new btc utxo used for signing by taker in !auth is appended to a 'blacklist' file with a counter. the counter is incremented when the utxo is reused. If the counter exceeds the value in "LIMIT","taker_utxo_retries", the !auth is rejected (so no maker utxos are sent across).

There are as I see it two big issues with this (at least).

The first, most important: we currently only send input btc *pubkey* in the !auth message, not a utxo. So, the taker can send any old pubkey, not necessarily corresponding to a real utxo. In the current model (see the last paragraph of encryption_protocol.txt), the validity of this pubkey is checked only before sending across our transaction signatures, so if a fake pubkey is used, we don't send signatures, but we do send utxos (in !iouath) from maker to taker.

This PR changes that, which means a version change in Joinmarket, so is very disruptive. The taker sends also a utxo along with the pubkey (the latter could be ditched). the maker calls query_utxo_set on it, and ensures that it's a valid utxo. It could easily (I would definitely add this) check the coin age/confirmations of that utxo also. This allows what's described in the first paragraph above.

The other big issue with this is that it makes takers very fragile to disruptive makers. If *any* maker decides not to cooperate (or just fails), this "uses up" that utxo and the taker cannot use it more than 3 times. *If* we were to apply this model I think we would also want to add code to allow the taker to choose a different one of his utxos in the next attempt.

The last point to make, I guess, is that this is fairly weak, depending on if we decided to include coin age restriction and coin amount restriction.

### DIFF
--- a/lib/bitcoin/__init__.py
+++ b/lib/bitcoin/__init__.py
@@ -4,4 +4,5 @@ from bitcoin.main import *
 from bitcoin.transaction import *
 from bitcoin.deterministic import *
 from bitcoin.bci import *
+from bitcoin.podle import *
 
--- a/lib/bitcoin/podle.py
+++ b/lib/bitcoin/podle.py
@@ -1,0 +1,100 @@
+#Proof Of Discrete Logarithm Equivalence
+#For algorithm steps, see https://gist.github.com/AdamISZ/9cbba5e9408d23813ca8
+import secp256k1
+import os
+from py2specials import *
+from py3specials import *
+
+N = 115792089237316195423570985008687907852837564279074904382605163141518161494337
+dummy_pub = secp256k1.PublicKey()
+
+'''NUMS - an alternate basepoint on the secp256k1 curve
+For background (taken from https://github.com/AdamISZ/ConfidentialTransactionsDoc/blob/master/essayonCT.pdf)
+>>> import bitcoin as btc
+>>> import os
+>>>H_x = int(sha256(btc.encode_pubkey(btc.G,'hex').decode('hex')).hexdigest(),16)
+>>> H_x
+	36444060476547731421425013472121489344383018981262552973668657287772036414144L 
+>>> H_y = pow(int(H_x*H_x*H_x + 7), int((btc.P+1)//4), int(btc.P))
+>>> H_y
+	93254584761608041185240733468443117438813272608612929589951789286136240436011L
+>>> H = (H_x, H_y)
+'''
+J_raw = '0350929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0'
+J = secp256k1.PublicKey(safe_from_hex(J_raw), raw=True)
+        
+def getP2(priv):
+    priv_raw = priv.private_key
+    return J.tweak_mul(priv_raw)
+       
+def generate_podle(priv):
+    '''Given a raw private key, in hex format,
+    construct a commitment sha256(P2), which is
+    the hash of the value x*J, where x is the private
+    key as a raw scalar, and J is a NUMS alternative
+    basepoint on the Elliptic Curve. Also construct
+    a signature (s,e) of Schnorr type, which will serve
+    as a zero knowledge proof that the private key of P2
+    is the same as the private key of P (=x*G).
+    Signature is constructed as:
+    s = k + x*e
+    where k is a standard 32 byte nonce and:
+    e = sha256(k*G || k*J || P || P2)
+    '''
+    if len(priv)==66 and priv[-2:]=='01':
+        priv = priv[:-2]
+    priv = secp256k1.PrivateKey(safe_from_hex(priv))
+    P = priv.pubkey
+    k = os.urandom(32)
+    KG = secp256k1.PrivateKey(k).pubkey
+    KJ = J.tweak_mul(k)
+    P2 = getP2(priv)
+    commitment = hashlib.sha256(P2.serialize()).digest()
+    e = hashlib.sha256(''.join([x.serialize() for x in [KG, KJ, P, P2]])).digest()
+    k_int = decode(k, 256)
+    priv_int = decode(priv.private_key, 256)
+    e_int = decode(e, 256)
+    sig_int = (k_int + priv_int*e_int) % N
+    sig = encode(sig_int, 256, minlen=32)
+    P2hex, chex, shex, ehex = [safe_hexlify(x) for x in [P2.serialize(),commitment, sig, e]]
+    return {'P2':P2hex, 'commit': chex, 'sig': shex, 'e':ehex}
+
+def verify_podle(Pser, P2ser, sig, e, commitment):
+    Pser, P2ser, sig, e, commitment = [safe_from_hex(x) for x in [Pser, P2ser, sig, e, commitment]]
+    #check 1: Hash(P2ser) =?= commitment
+    if not hashlib.sha256(P2ser).digest() == commitment:
+        return False
+    sig_priv = secp256k1.PrivateKey(sig,raw=True)
+    sG = sig_priv.pubkey
+    sJ = J.tweak_mul(sig)
+    P = secp256k1.PublicKey(Pser, raw=True)
+    P2 = secp256k1.PublicKey(P2ser, raw=True)
+    e_int = decode(e, 256)
+    minus_e = encode(-e_int % N, 256, minlen=32)
+    minus_e_P = P.tweak_mul(minus_e)
+    minus_e_P2 = P2.tweak_mul(minus_e)
+    KG = dummy_pub.combine([sG.public_key, minus_e_P.public_key])
+    KJ = dummy_pub.combine([sJ.public_key, minus_e_P2.public_key])
+    KGser = secp256k1.PublicKey(KG).serialize()
+    KJser = secp256k1.PublicKey(KJ).serialize()
+    #check 2: e =?= H(K_G || K_J || P || P2)
+    e_check = hashlib.sha256(KGser + KJser + Pser + P2ser).digest()
+    if not e_check == e:
+        return False
+    return True
+
+if __name__ == '__main__':
+    
+    for i in range(10000):
+        priv = os.urandom(32)
+        Priv = secp256k1.PrivateKey(priv)
+        Pser = safe_hexlify(Priv.pubkey.serialize())
+        podle_sig = generate_podle(safe_hexlify(priv))
+        P2ser, s, e, commitment = (podle_sig['P2'], podle_sig['sig'], 
+                                   podle_sig['e'], podle_sig['commit'])
+        if not verify_podle(Pser, P2ser, s, e, commitment):
+            print 'failed to verify'
+    
+    
+    
+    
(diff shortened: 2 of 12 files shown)


AdamISZ:
Fixed up the commit so you can see the code (got rid of the junk changes).

AdamISZ:
Amended to include a proof of discrete log equivalence - style commitment. See Defence 2 [here](https://gist.github.com/AdamISZ/9cbba5e9408d23813ca8).

The handshake is now:

    !fill oid amount encryption_pubkey commitment=H(P2)
    !pubkey encryption_pubkey
    !auth btc_pub btc_sig input_utxo P2 s e

Since only messages from taker to maker are affected, the backward compatibility concern only operates in one direction. This code allows the maker to accept old-style takers (if commitment not provided, operates as before).

I believe it operates correctly with `create-unsigned-tx` (picks up the authorising utxo from there).

The "blacklist" aspect is unchanged from the previous version; the commitments (H(P2) values) are added to a blacklist file and requests are rejected when exceeding the value in "LIMITS", "taker_utxo_retries" in the config.

Giszmo:
Just to reflect what was already said on IRC: Please consider wallet providers. This is maybe not an exhaustive perspective but best think of hardware wallets that under no circumstances would share private keys but that might be convinced to blindly sign transactions that make them richer (maker's yielding) if the rules to do so can be embedded in them in a fool proof way.

I'm trying to integrate JM in Mycelium without re-implementing JM in Java but even if somebody provided a Java library for that, I would hesitate to put it anywhere near the wallet code and watch over all changes to that library like a hawk, to keep the wallet safe.

My current approach is to run joinmarket in an http proxy but for sake of argument, a proxy could also run on the wallet's phone itself, in its own sandbox, which wouldn't change the API requirements although it would make things decentralized and put joinmarket under the sole control of the user.

For all the above it would be desirable to require little to no extra round trips and especially little to no extra signing other than the final transaction. The current requirement to sign a NACL key with a private key involved in the UTXO set is already removing hardware wallets from the picture for now, as this would mean yet another step, where the user would have to approve the signing of some "random" string just to kick off the joinmarket "negotiations", to then, seconds later approve of a transaction.

AdamISZ:
Well, it's only fair to include the detailed responses I gave on IRC:

waxwing> i appreciate Giszmo's concern and would say 2 things: 1/ merging #328 as current would not enforce the PoDLE check, in that code makers will simply continue as before if it's not provided. Enforcing it would be a huge step, so..
 waxwing> and 2/ I'm thinking what we need is an abstraction in which a maker has some kind of policy file, where he sets flags for which checks he enforces. Other possible policy checks might be a PoW, a signature from a proxy (gulp .. maybe?), other things we haven't thought of.
 waxwing> there could even be intelligence in it, like 'request rate is going up, switch to LOCKDOWN mode, allow only with strong checks now'. This might be a terrible idea, but just throwing it out there.
waxwing> i guess the most important technical point would be: if we allowed multiple different styles of commitments, they would have to all fit in the same field (although not necessarily same byte length), so as to keep the messaging protocol clean.
waxwing> probably !fill oid amount <commitment> . but, guess that's pretty easy unless there were some form of commitment with a ridiculously unwieldy size.

I've tried to explain that we cannot simply remove the nacl key signing; it is the only guarantee that the counterparty has that he isn't being MITMed. It's intended (caveat: see above) to ensure that the entire conversation is with one entity - the owner of those bitcoins/utxo. 

At the same time, I fully understand that PoDLE is not realistic (at least not currently) for a hardware wallet, it requires EC operations with the private key, no less. Hence my thoughts listed above.

If an implementation of JM was indeed *integrated* into a wallet, this might not be an issue. Not to downplay how difficult that would be. If you try to separate JM entirely from the wallet, a desire I totally understand, then this extra signing operation problem occurs, but **for good reason**.

Giszmo:
@AdamISZ it's not that I don't get what the goal is here but there is more things to consider:

* I heard third hand that big companies would raise red flags and refuse to take coinjoin transactions, so the window of opportunity to roll them out as the standard might be closing. If it's standard to block them, users will be annoyed at the wallet tricking them into using this evil stuff. If it's standard to use them, they will be annoyed at the business that freaks out over their needs for privacy.
* Anything that looks like CJ but will only be somewhat private once implemented better later raises awareness for CJ. And I believe that an explicit MITM, like a "benevolent" Mycelium server can be a very good first step to improve privacy. We promise to not log your transactions but sure we could but with CJ at least only we could for now.
* The ability to delegate is not the same as being open to arbitrary MITM and not being able to delegate doesn't absolutely protect against MITM. In fact I would be surprised if companies were not able to convince people to share this harmless `xpriv` of theirs, in order to provide them with ultra secure and private transactions. After all it's a reputable company that would never steal their funds.
* With explicit delegation via an individually signed nacl key, the client gains security and doesn't loose it. It can't be in the interest of the joinmarket project to bloat wallet code. In fact to have something less worrying to sign, I wish it was not "[random string]" but rather some clear ... "joinmarket auth key [random string]".

AdamISZ:
@Giszmo 

>I  heard third hand that big companies would raise red flags and refuse to take coinjoin transactions, so the window of opportunity to roll them out as the standard might be closing. If it's standard to block them, users will be annoyed at the wallet tricking them into using this evil stuff. If it's standard to use them, they will be annoyed at the business that freaks out over their needs for privacy.

Well, that's certainly interesting! I haven't heard such myself (and indeed have made bitpay payments with coinjoins with no problem), but it raises interesting questions if it could happen. As far as I'm aware, people have been using both DarkWallet and SharedCoin with no problems so far? 

It would be most disappointing if people actively try to degrade Bitcoin's fungibility like that. It is effectively the same as whitelisting (we only accept utxos conforming to X), which is pretty much the worst scenario. But I suspect a long discussion of this veers too far off topic.

> not being able to delegate doesn't absolutely protect against MITM.

I don't understand what you mean here. It's intended to do exactly that.

> With explicit delegation via an individually signed nacl key, the client gains security and doesn't loose it.

I agree, I don't see anything wrong with this model if people opt in to it. But it's this model that you're unhappy about, no? Because you don't like needing the extra signature.

> In fact to have something less worrying to sign, I wish it was not "[random string]" but rather some clear ... "joinmarket auth key [random string]".

That's a very good technical point which I think we overlooked (to be fair to us, it's because we didn't anticipate delegation, instead we thought only of integration).

>  It can't be in the interest of the joinmarket project to bloat wallet code.

"Bloat" just seems pejorative here. We created a bare-bones wallet in joinmarket because we understood it would not be easy to integrate this functionality into existing wallets (although on the plus side, you don't need mixing depths if only doing taker side), but however difficult that might be, it is the natural way to expose it to wallet users.

Giszmo:
> > not being able to delegate doesn't absolutely protect against MITM.

> I don't understand what you mean here. It's intended to do exactly that.

I gave the example of users exposing their keys to third party banks in order to do JM, if delegation is prevented, putting that bank into the position of an ultimate MITM

> I agree, I don't see anything wrong with this model if people opt in to it. But it's this model that you're unhappy about, no? Because you don't like needing the extra signature.

Ideally I would pay for the maker revealing their order books and UTXOs factions of a cent, and then pay for signing my tx and then pay for not double spending on me per second. I doubt that the proof-of-good-intentions by requiring the involved UTXO makes any sense as a commitment and for a MITM protection I agree that it might provide some benefit although I'm not convinced. Naively I would rather have some makers with reputation linked to a nacl key that people trust, which allows takers to contact those. If the up-front commitment is an UTXO's signature, then a committed attacker can always query all the makers at no costs and the taker role can't be moved into a module away from the private keys.

Of course we also want to do the maker side at some point or in other words, users will want to put their money into wallets, not into mixing scripts. The interface that approves of mixing transactions should be decoupled from all the coin selection logic, communication protocol, order book management etc. I want joinmarket to only do join market and wallets to do transaction approval. At mycelium we are thinking about putting extra stuff into modules but we should actually do the private keys containing stuff into a module with a hardware-wallet like interface. It's pretty insane to have a 10MB app to keep secure against all the attacks.

chris-belcher:
Regarding the taker-side-auth-being-optional, this was discussed once before when planning for a protocol update: https://github.com/chris-belcher/joinmarket/pull/90#issuecomment-108648823

There we both seemed to agree that making taker-side-authentication optional was okay. I don't know if your opinion has changed since then (I'm not sure myself)

In Mycelium's case (and also Electrum and any other lightweight wallets), the server already knows all the addresses of the user's wallet and that they're linked together, with this information they can already unmix all your coinjoins.

However, this PR is about solving issue #156, allowing takers to opt-out from providing a UTXO would stop this PR from working, which could ruin the privacy for every JoinMarket user, not just users of lightweight wallets.


Large companies blocking coinjoin is interesting, I hoped JoinMarket would stay under the radar a little longer but seems not. CoinJoins can be quite hard to spot, many large bitcoin institutions that use Bitcoin Core sendmany can sometimes create coinjoin-like transactions. (Look at Adlai's cjhunt to see how many false positives there are out there)

Also some wallets like Samourai wallet and Bitcoin Core plan to create coinjoin-like transactions by default (making the change be two outputs where one of them is the same value as the payment output) which might also gunk up the analysis.

Finally, coinswaps as in #335 are another way to improve privacy that can be indistinguishable from normal bitcoin payments.

AdamISZ:
@chris-belcher 

Re #90 : good point! I don't know whether I was thinking more carefully there, or just glossing over details, but it's true: if we only want to prevent a MITM attack then **authentication on one side only is sufficient**. I had completely forgotten this rather crucial subtlety.

Using a bitcoin key is us trying to get the weakest viable form of "attestation to an identity". Without even that (on either side), then MITM is fully possible in as much as anyone can carry out the protocol with you, and replay your side of the protocol to anyone else. There can even be a huge chain of such MITMs.

A one sided authentication is enough to stop that. It doesn't mean that the authenticating side isn't a snooper, but it at least means you're talking to someone who owns that private key. Weak, but not nothing. And that single authentication prevents passing the dialogue on because a MITM needs to be able to fake both sides.

Then, there's wanting to have both sides make some kind of commitment/attestation, to prevent partial snooping: learning utxos, principally. For that, a double-sided authentication with btc signatures is a logical choice, but it might not be necessary and probably isn't sufficient (at least, if you allow infinite repeats). 

So there are the other possibilities: PoDLE, P2CH, perhaps PoW or even what I mentioned on IRC: 

> . Other possible policy checks might be ..., a signature from a proxy (gulp .. maybe?), ...

i.e. you could replace the authentication-via-bitcoin-signature with trust. And on that topic:

>In Mycelium's case (and also Electrum and any other lightweight wallets), the server already knows all the addresses of the user's wallet and that they're linked together, with this information they can already unmix all your coinjoins.

This is a good point which I had also overlooked (although it might not be 100% true that they know all addresses, but let's say, for the sake of argument). It doesn't mean delegation is great, but it strengthens the case that using trust here doesn't lose much.

> Look at Adlai's cjhunt to see how many false positives there are out there

Minor point, but I wouldn't overplay this; all the false positives seemed to have basically the same pattern. I don't think there are a huge number of txs at the moment which are a good match to our JM txs.

For hardware wallets: a "signature from a delegating proxy" doesn't work there. What else is left I wonder, to prevent spam/snooping? If PoW is not practical, I wonder. I can't see how we can have takers propose transactions without either a btc signature or *any* form of commitment.

PR #340, opened 2015-12-01
Blame preserving reformat - simple merge

ghtdak:
This PEP-8 / Google Yapf reformat preserves the blame history of the project and is a straight merge with the Joinmarket master  (In contrast with the tree replacement strategy discussed previously)

This pull-request contains no code changes other than Yapf reformats though is an important starting point to begin the process of refactoring joinmarket.

ghtdak/master contains the squashed rebased merge of the parallel (universe?) reformatted project history exploiting some git-fu: 

```git merge --squash -X theirs master-yapf```

This reformat fixes many of the problems of the unformatted and very "tool unfriendly" aspects of the codebase (most notably tabs... ugh...:-)

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -22,13 +22,14 @@ import old_mnemonic, slowaes
 
 parser = OptionParser(usage='usage: %prog [options] [wallet file] [method]',
 	description='Does useful little tasks involving your bip32 wallet. The'
-	+ ' method is one of the following: display- shows addresses and balances.'
+	+ ' method is one of the following: display - shows addresses and balances.'
 	+ ' displayall - shows ALL addresses and balances.'
 	+ ' summary - shows a summary of mixing depth balances.' 
 	+ ' generate - generates a new wallet.'
 	+ ' recover - recovers a wallet from the 12 word recovery seed.'
 	+ ' showseed - shows the wallet recovery seed and hex seed.'
-	+ ' importprivkey - adds privkeys to this wallet (privkeys are spaces or commas separated)')
+	+ ' importprivkey - adds privkeys to this wallet (privkeys are spaces or commas separated)'
+	+ ' listwallets - lists all wallets with creator and timestamp')
 parser.add_option('-p', '--privkey', action='store_true', dest='showprivkey',
 	help='print private key along with address, default false')
 parser.add_option('-m', '--maxmixdepth', action='store', type='int', dest='maxmixdepth',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by chris-belcher** [comment]
I'd like to do this. I think each line will have to be looked over before merging though. I could find the time to do this this week I think.

**Comment by ghtdak** [comment]
Its easy to re-create the final state of ghtdak/joinmarket/master from belcher/master or JoinMarket-Org/master using [```yapf```](https://github.com/google/yapf)

```bash
pip install yapf
cd <repo>
yapf -i -r -style google .
```
I recommend [```tig```](https://github.com/jonas/tig) for viewing the git repo though [PyCharm Community Edition](https://www.jetbrains.com/pycharm/) is better for detailed git diff analysis of python.

I've run yapf on some pretty big codes (e.g [```twisted```](https://twistedmatrix.com/trac/)) and it doesn't change the results of trial or tox. YMMV but I'd be surprised if it broke Joinmarket.

Note: I used ```google``` style which is slightly different from the ```yapf``` PEP-8 baseline.  I'm sure I have a good reason :-)

meh...

AdamISZ:
First check through: used `git diff -w`. This throws up only cases where the lines have been broken up. Of course, reviewing that manually was .. interesting :) .. so obviously I didn't attempt to do it perfectly, but used a "heuristic" approach. At least, it threw up the amusing fact that there are 2 lines ending in semicolons that get removed. Another thing thrown up: `old_mnemonic.py` word list is one word per line, this is corrected to several words per line. Is there any obscure reason I can't think of why it's helpful to stick to one word per line? Almost certainly it doesn't matter.

This version of the code passes regtest.py and wallet-test.py.

The last check I did was this: create a fresh copy of the JoinMarket-Org/master, apply the `yapf` command above (note: --style not -style), and do a `diff` on the entire directory tree, to this PR. There were no differences, **except** the following two files:
`lib/bitcoin/bci.py`
`lib/bitcoin/ripemd.py`

Not sure the logic of not applying the changes to these files?
Apart from that, the check passed: `diff` showed only difference in git files, not in any code.

Short version; ACK.

Last comment, I want to say how much I appreciate this work @ghtdak , not just for the obvious reason, but also because looking at the updates helps me to understand more concretely what's needed (well, at least at the *formatting* level).

AdamISZ:
Also passes tumbler-test script. This is a significantly richer test, using 8 participants all with wallets randomly filled across 5 mixdepths, and 16 transactions. Of course, these tests do not cover that much even so.

ghtdak:
> The last check I did was this: create a fresh copy of the JoinMarket-Org/master, apply the yapf command above (note: --style not -style), and do a diff on the entire directory tree, to this PR. There were no differences, except the following two files:
> lib/bitcoin/bci.py
> lib/bitcoin/ripemd.py

Yeah, something happened with those two files during the squash squash merge.  The resulting seemed fine so I didn't sweat it though its on my queue of things to look at (priority just got higher :-)

WRT the line splitting... I decided to use the ```google``` style because it seemed to do a nice job and I figured that would be their emphasis.  However, there are many knobs which can be tweaked, many of them continuous variables.  ```Yapf``` a scoring system to make certain decisions and that can be played with.

PR #342, opened 2015-12-01
Enable manual counterparty selection for sweeping

adlai:
It was initially disabled, because sweeping considered all relevant counterparty combinations. The algorithm which replaced that for quantitative reasons also has the qualitative benefit unlocked herein.


adlai:
Closed in favor of #360

PR #350, opened 2015-12-07
Python3 conversion

ghtdak:
This is the minimal effort slash and burn refactor into python3.  It was much easier to remove support for python2 and refactor than continue supporting python 2.  Joinmarket has no requirement for python2

I've removed "lib" and created a joinmarket module.  It needed to be done and is, therefore, official (so let it be written, so let it be done)

Along the way, I fixed a number of glaring errors though there are some which continue to exist and I've market them with comments.  Many uninitialized variables etc.

I would appreciate it if those familiar with the codebase quickly browse through the changes and look where I've marked issues.  There are places in the code where, if executed, it would fail.  In simple cases, I tried to fix things, but there are spots which would take more time than I wanted to spend given the scope I set for myself.


ghtdak:
Rethinking the right approach.

Belay for now.

PR #349, opened 2015-12-06
include all yieldgens

raedah:
This moves the yieldgens into develop. 

The basic yieldgen from master is renamed here to yield-generator-basic.py.
The yieldgen that was in new-yieldgen-algo branch is now yeild-generator-mixdepth.py.
The yieldgen that I developed and was reviewed at https://github.com/JoinMarket-Org/joinmarket/pull/329 , is included here as yield-generator-deluxe.py.

Next these will be moved into their own folder with the includes modified as needed. As a first step, putting them in the top folder where they all work and the files have not been otherwise modified. 

After this is pulled, I will close my previous pull #329, and we can get rid of the new-yeildgen-algo branch. I verified with belcher on irc that the yield-generator.py was the only file modified in the new-yieldgen-algo branch.

PR #353, opened 2015-12-09
refactoring for yieldgens

raedah:
These ran and completed a test transaction. Changes were for compatibility with latest refactor ( and yapf formatting).

### DIFF
--- a/joinmarket/irc.py
+++ b/joinmarket/irc.py
@@ -11,7 +11,7 @@ from joinmarket.configure import jm_single, get_config_irc_channel
 from joinmarket.message_channel import MessageChannel, CJPeerError
 from joinmarket.enc_wrapper import encrypt_encode, decode_decrypt
 from joinmarket.support import get_log, chunks
-from joinmarket.socks import socksocket, setdefaultproxy
+from joinmarket.socks import socksocket, setdefaultproxy, PROXY_TYPE_SOCKS5
 
 MAX_PRIVMSG_LEN = 400
 COMMAND_PREFIX = '!'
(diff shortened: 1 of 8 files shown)


AdamISZ:
(Don't worry about the travis fail, I'll fix that).
This will need to be tested against tumbler-test for each of the new yield-generator types. See the comment to #349 .

AdamISZ:
Running tumbler tests now.

The -mixdepth version is crashing, it seems because of this: https://github.com/JoinMarket-Org/joinmarket/pull/353/files#diff-3c58b00fab3804e7220db1ac4cdd652bR155

The mixdepth filtering is not checking against the 'amount' variable passed as argument to oid_to_order.

Second comment (can be put off): it seems weird to be re-defining the YieldGenerator *class* in these files separately, we should probably have a base yieldgenerator class and each of these can derive from it. But, no matter, we can look at that later.

Third comment, I don't understand why there's lots of reformatting code in this PR (e.g. irc.py, sendpayment.py). That should already be in @ghtdak's stuff? Or is it reformatting that he missed? (e.g. the print() statements in sendpayment).

raedah:
1) the -mixdepth version crash, that code is exactly as it was in the new-yieldgen-algo branch. I haven't touched it. My next pull request was going to be a merging over the order filling code from -deluxe because of obvious problems with what is currently done in the -mixdepth version code. Mostly in regards to https://github.com/JoinMarket-Org/joinmarket/issues/156 , but ill look at this error too.

2) I have just been keeping with the way it was done. Ill ask you more about this.

3) The reformatting code is stuff that was missed, includes, prints, and I needed to get the tests to run. I also modified the tests to make it easier for my configuration.

raedah:
I replaced most of the order filling code 'oid_to_order' in the yg-mixdepth.py code with the new code from the yg-deluxe.py.  This should fix the first issue @AdamISZ mentioned above, and well as fix the security vulnerability discussed over here https://github.com/JoinMarket-Org/joinmarket/issues/156#issuecomment-123750171

Issue #356, opened 2015-12-11
choose_sweep_orders algorithm


AdamISZ:
In running tumbler tests on the new versions of the yield generator (-mixdepth, -deluxe), I am getting "not enough liquidity" errors, where there is actually sufficient liquidity in the orderbook.

At the moment, it filters the orders based on minsize (not maxsize, understood: because we don't yet know the cj_amount), then passes to chooseOrdersBy and then chooses one at a time. The problem seems to be that, since it filters out all other orders from the same counterparty after choosing one, it might by chance pick too-small orders from each counterparty, and not achieve a sufficiently large list (n) of orders with a sufficient size.

This is partly ameliorated, assuming chooseOrdersBy is random weighted choice, by repeatedly trying after the liquidity error until you by chance get the orders with sufficient size. (Hence the -q option for tumbler can help). An alternative would be to choose the largest order for each counterparty, but that would mean being insensitive to cjfee price.

If we did not remove all orders from a single counterparty after trying once, then this problem goes away, but that reintroduces the problem of a maker gaming the system by flooding with lots of repeated orders.

Two questions: have I understood right, and is there a better way?

raedah:
I havent used the tumbler much. Why doesnt it know the cj_amount? Its counter intuitive to try to pick the fill order when you dont know your cj_amount yet.

AdamISZ:
@raedah It's because of the fact that: suppose you have 3BTC in the mixdepth. You want to create a transaction to spend all of it. But, until you know exactly which counterparties you're transacting with, you don't know the fees you're going to be spending (both cjfees and tx fees). And so you don't know the exact size of the coinjoin output.

This is what makes sweep logic rather tricky.

raedah:
I have this one my list, though it looks like a bigger item so it may not be the first thing I get to.

chris-belcher:
Note that we don't have to assume chooseOrdersBy, we can check with

    if chooseOrdersBy == weighted_order_choose:

PR #357, opened 2015-12-13
yieldgen order filling fixes and dynamic txfee contributions

raedah:


AdamISZ:
W.r.t. https://github.com/JoinMarket-Org/joinmarket/pull/357/files#diff-0bcdba6c62af8e26a6012223336d22cbR114

Doesn't this still need to use the modular arithmetic rule to respect the wrapping from max mixdepth to zero?

raedah:
@AdamISZ You are correct. I thought coins stayed at max mixdepth. Rotating circularly is the behavior yg-basic already does, so it will stay as is. I have rewritten the code in yg-deluxe to follow this circular behavior, though in a different coding style which is more modular and allows for other options.

For yg-mixdepth, it gets reverted, but what was being done the yg-mixdepth code previously https://github.com/JoinMarket-Org/joinmarket/blob/new-yieldgen-algo/yield-generator.py#L112 was buggy because it didn't check the amount, and it was tying offers to mixdepths which was an unmixing vulnerability. A better way to achieve that direction is to empty out small mixdepths by filling orders with the closest sized mixdepths, which is an option I have commented out in yg-deluxe here https://github.com/JoinMarket-Org/joinmarket/blob/develop/yield-generator-deluxe.py#L331 . So that code is becoming the default in yg-mixdepth now.

There is no clear way to combine the two goals of clumping and also using the closest sized mixdepth to an order. We choose one or the other. Right now its commented out in yg-deluxe but I could make it a user config setting. Clumping pushes all coins towards the largest mixdepth without issue, so using closest sized is not ever necessary, but may qualify as an optimization under some circumstances. The complicated way would be to use a weighted score for how close the order size is to the mixdepth size to determine how to prioritize, at which point we would probably want to also be considering the sizes of the utxos too which the yieldgens currently dont access at all. Another strategy would be creating discounted offers which are exactly sized to fill a total amount from within a mixdepth with no change outputs. I am keeping these ideas in a list.

yield-generator-deluxe.py
```diff
@@ -289,59 +289,53 @@ def create_my_orders(self):
...
-            m for m in filtered_mix_balance if m[0] <= largest_mixdepth
-        ]  #use mixdepths before and including the largest
+                                  reverse=True)[0][0]  # find largest amount
         filtered_mix_balance = sorted(
             filtered_mix_balance,
-            key=lambda x: x[0])  #make sure we are in seq of mixdepth num
+            key=lambda x: x[0])  # make sure we are in seq of mixdepth num
 
-        #use mix depth with the most coins, 
-        #creates a more even distribution across mix depths
```
raedah: these diffs are comparing the wrong lines

yield-generator-deluxe.py
```diff
@@ -289,59 +289,53 @@ def create_my_orders(self):
...
                                   key=lambda x: x[1],
-                                  reverse=True)[0][0]  #find largest amount
-        filtered_mix_balance = [
-            m for m in filtered_mix_balance if m[0] <= largest_mixdepth
-        ]  #use mixdepths before and including the largest
+                                  reverse=True)[0][0]  # find largest amount
         filtered_mix_balance = sorted(
             filtered_mix_balance,
-            key=lambda x: x[0])  #make sure we are in seq of mixdepth num
+            key=lambda x: x[0])  # make sure we are in seq of mixdepth num
```
raedah: below diffs are comparing the wrong lines

AdamISZ:
Running off this PR, I found:

tumbler-test still runs correctly off yield-generator-basic

tumbler-test runs correctly off yield-generator-mixdepth **if** I edit it to add more coins to the yield generators. I'm reasonably convinced that this is a result of #356 and therefore is not because of any error in this code.

tumbler-test failed off yield-generator-deluxe at the first transaction. The maker bots which were chosen crashed with this error:

```
    File "yield-generator-deluxe.py", line 462, in main
    irc.run()
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/irc.py", line 559, in run
    self.__handle_line(line)
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/irc.py", line 447, in __handle_line
    self.__handle_privmsg(_chunks[0], _chunks[2], get_irc_text(line))
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/irc.py", line 398, in __handle_privmsg
    self.__on_privmsg(nick, parsed)
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/irc.py", line 276, in __on_privmsg
    self.on_order_fill(nick, oid, amount, taker_pk)
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/maker.py", line 238, in on_order_fill
    taker_pubkey)
  File "/home/adam/DevRepos/JoinMarket/testing/joinmarket/joinmarket/maker.py", line 50, in __init__
    self, oid, amount)
  File "yield-generator-deluxe.py", line 338, in oid_to_order
    (filtered_mix_balance.index(largest_mixdepth) + 1
ValueError: 1 is not in list
```

The figure '1' is variously 0,1,3,4 in other bots.

I didn't try regtest.py, maybe it would flag it also.

I have updated and corrected the Testing and Testing2 wiki pages on the wiki (they're a lot tidier now hopefully) to reflect updates and specifically running the tests using miniircd (which is now merged into the travis builds also). Please use this setup as a sanity check for the code, there should no longer be any block to doing this I hope.

raedah:
regtest on deluxe, Ran 2 tests in 236.510s

AdamISZ:
@raedah sorry we're not around at the same time on IRC; w.r.t. the problem you had with miniircd, did you resolve it? If not, the symptoms you described match what I originally saw until I made sure that the full path to the MOTD file was specified, as in the Testing wiki description: ./miniircd --daemon --motd=/full/path/to/motdfile

So, please do still make sure you can get the tumbler-test to run, but if the regtest is now passing that's a reasonable baseline (considering this is just an edit to existing scripts), so, merging now.

PR #362, opened 2015-12-14
Incentivise Network Diversity

adlai:
This commit makes naive sybil attacks more difficult by keying counterparty
deduplication off IRC hostmasks, disincentivising running multiple yield
generators from the same machine or network.

It also fixes a minor bug: the transaction initiator must deduplicate BEFORE
counting the number of liquid counterparties.

It also tidies a bit of suboptimal yapfage, and factors out a rhyming utility.

It is a revival of #313, neé #311

### DIFF
--- a/joinmarket/support.py
+++ b/joinmarket/support.py
@@ -235,6 +235,14 @@ def pick_order(orders, n, feekey):
             return orders[pickedOrderIndex]
         pickedOrderIndex = -1
 
+def offer_profit(offer):
+    """
+    Calculates the net profit to the maker of the given offer.
+    Expects an offer in the form (counterparty, oid, cjfee, txfee)
+    where cjfee and txfee are integer amounts of satoshi.
+    """
+    return offer[2] - offer[3]
+
 
 def choose_orders(db, cj_amount, n, chooseOrdersBy, ignored_makers=None):
     if ignored_makers is None:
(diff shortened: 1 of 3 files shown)


adlai:
It should be noted that yield generators using the hidden service will be at a disadvantage when transactions are initiated by users of this code, since liquidity providers are incentivised to use unique IPs. If anything, this is a net benefit, because connecting via the hidden service incurs higher latency without tangible privacy benefit: hidden services conceal the location of the server, but provide no better privacy for clients.

joinmarket/support.py
```diff
@@ -290,7 +293,7 @@ def choose_sweep_orders(db,
                         txfee,
                         n,
                         chooseOrdersBy,
-                        ignored_makers=None):
```
AdamISZ: Comment also applies to L305-307.

See: https://github.com/JoinMarket-Org/joinmarket/blob/dc8cfd06aa7b0441ba5c903d814c28e5fde2ed34/joinmarket/__init__.py#L716-L729

@ghtdak explained to me that this change is due to a 'feature' in Python; see: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

I suggest reverting these changes.

adlai:
As discussed on IRC:
```
  adlai | deduping counterparties by hostmask messes up the tests, since all the scripts are on the same host
  adlai | it works fine on an actual server with distinct connections
waxwing | that's a very interesting example of a test failing because the code works correctly
  adlai | my reflex is to just include our own diddled miniircd, for testing
waxwing | adlai: probably, yes, we could keep it in the test folder. i get your point: you might need to diddle the server code to, what, assign random hostmasks?
  adlai | just append host[1] (ie, source port) in https://github.com/jrosdahl/miniircd/blob/ca46821/miniircd#L128
```

So this PR is pending: fixing this test, alerting in the forum, and an :ok: from @chris-belcher.

chris-belcher:
I agree in principle I guess. Although its trivial for an attacker to get around this, at least they have to work for it a bit.
I'm slightly nervous about this being cyberguerilla specific. Other IRC servers might not allow tor connections from exit nodes.

I'd say merge but when the multiirc branch is finalized maybe have some check to make this only fire for cyberguerilla. I haven't even started looking for other IRC networks to use so I don't know how multiirc will end up looking like. The multiirc branch probably makes this kind of attack easier anyway.

adlai:
Rebased onto develop (e0f7c59)

PR #365, opened 2015-12-15
Update common.py

anduck:
If there are lots of orders to pick from, it's better to have descending order so cheaper offers are better visible.


anduck:
Some bugging, needs fixing before pr.

Issue #39, opened 2015-02-05
Add unconfirmed coins to output from wallet-tool


AdamISZ:
(low priority but worth doing at some point)

BitpopCoin:
This seems To be A blockr issue

BitpopCoin:
Including unconfirmed (zero confirmation) transactions.

Adding an optional parameter unconfirmed=1 to unspent API, the API searches and returns unspent transactions also among transactions that are not yet part of the blocks (unconfirmed transactions). 
Whenever unconfirmed=1 is used a new field "with_unconfirmed": true is added to the response to let developers know there are unsafe (unconfirmed) transactions in the output.
http://btc.blockr.io/api/v1/address/unspent/198aMn6ZYAczwrE5NvNTUMyJ5qkfy4g3Hi?unconfirmed=1

chris-belcher:
yes, the blockr.io API supports checking for unconfirmed transactions.
It's also quite easy to use Bitcoin Core json-rpc for unconfirmed transactions.
I'm also keen to use insight API as in #64 for further decentralization.

If you're planning to program this up and submit a pull request, be aware that makers should not send unconfirmed transactions are UTXO inputs.

chris-belcher:
Note that this requires somewhat redesigning the Wallet class

chris-belcher:
If we ever do this, be mindful of transaction malleability. See this thread for the havoc caused by poorly written handling of unconfirmed txes https://www.reddit.com/r/Bitcoin/comments/3nfb2y/eli5_for_double_spends_bitcoin_being_sent_twice/

PR #367, opened 2015-12-16
add confirmation-spendability knob for bitcoind-rpc

adlai:
This addresses #39 but does not solve it for blockr users.

When using a full node, `wallet-tool.py` will display unconfirmed balances by default, but other scripts won't actually spend them unless that behavior is manually configured.

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -135,6 +135,18 @@ merge_algorithm = default
 # as our default. Note that for clients not using a local blockchain
 # instance, we retrieve an estimate from the API at blockcypher.com, currently.
 tx_fees = 3
+# the range of confirmations passed to the `listunspent` bitcoind RPC call
+# 1st value is the inclusive minimum, defaults to one confirmation
+# 2nd value is the exclusive maximum, defaults to most-positive-bignum (Google Me!)
+# leave it unset to defer to bitcoind's default values, ie [1, 9999999]
+#listunspent = []
+# unless you have a specific reason, eg:
+#  spend from unconfirmed transactions:  listunspent_args = [0]
+# display only unconfirmed transactions: listunspent_args = [0, 1]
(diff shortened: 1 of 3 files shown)


AdamISZ:
It occurred to me earlier today that what people want is for the wallet-tool to by *default* show both confirmed and unconfirmed amounts. At least, that's what I'd want. That's not to say that fine-grained control isn't nice, too.

AdamISZ:
Apart from that nit, utACK.

AdamISZ:
W.r.t. https://github.com/JoinMarket-Org/joinmarket/pull/367/files#diff-c1c696807da8327197e6ee5f13c1389eR148 - I am wondering, apart from the malleation warning, whether it might not actually work to use unconfirmed (esp. if we consider counterparties not using this setting). Seems unobvious.

AdamISZ:
Tried on mainnet and confirmed functionality: with nothing specified in config, wallet-tool (working off Core interface) showed the unconfirmed coins.

joinmarket/configure.py
```diff
@@ -135,6 +135,18 @@ def jm_single():
 # as our default. Note that for clients not using a local blockchain
 # instance, we retrieve an estimate from the API at blockcypher.com, currently.
 tx_fees = 3
+# the range of confirmations passed to the `listunspent` bitcoind RPC call
+# 1st value is the inclusive minimum, defaults to one confirmation
+# 2nd value is the exclusive maximum, defaults to most-positive-bignum (Google Me!)
+# leave it unset to defer to bitcoind's default values, ie [1, 9999999]
+#listunspent = []
```
AdamISZ: shouldn't this line show `#listunspent_args = []` ?

adlai: Thank you, I'll fix it.

PR #368, opened 2015-12-17
Restructure for_change HD branch as {in,ex}ternal

### LINKED ISSUE #283: Wallet should auto-pay only to the change address branch
The change and receive address branches function the same when being spent from, they are both parts of the mixing depth.

When a wallet is generating an address to send to itself, it should send to the change address branch. It has happened that a user of yield-generator has sent new coins to an address on the receive branch. When the yield generator requires a new address it will generate the same address(!) Leading the address re-use.

Just a few lines of editing. Should rename the two branches, perhaps to "internal" and "external".

@adlai 's idea originally.

adlai:
Fixes #283

Note that this is MORE than a naming change: tumbler.py, maker.py,
and the yield-generator scrypts have been modified to use only the
internal branch, specifically to avoid reuse of receiving address.

### DIFF
--- a/joinmarket/maker.py
+++ b/joinmarket/maker.py
@@ -343,8 +343,8 @@ class Maker(CoinJoinerPeer):
 		"""
 
         order = [o for o in self.orderlist if o['oid'] == oid][0]
-        cj_addr = self.wallet.get_receive_addr(order['mixdepth'] + 1)
-        change_addr = self.wallet.get_change_addr(order['mixdepth'])
+        cj_addr = self.wallet.get_internal_addr(order['mixdepth'] + 1)
+        change_addr = self.wallet.get_internal_addr(order['mixdepth'])
         return [order['utxo']], cj_addr, change_addr
 
     def get_next_oid(self):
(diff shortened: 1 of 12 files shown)


chris-belcher:
Good PR, this should be merged IMO.

When develop is released, the wiki page on github about the wallet should be slightly edited too.

adlai:
It's worth noting that BitcoinCoreWallet had no method previously for `get_receive_addr()`, and thus no `get_external_addr()` was added. This seems incomplete, but I've not yet investigated this interface, having used only the internal wallet myself.

AdamISZ:
tumbler-test passes.
wallet-tool seems to function OK, and sendpayment run OK.

AdamISZ:
Thought about the logic for a bit. Can't see any flaw in the reasoning. ACK

chris-belcher:
BitcoinCoreWallet is only used when sendpayment is run with --rpcwallet. The user inputs their own coinjoin address and the wallet object only needs a change/internal address.
It's not possible to run a yield generator or tumbler from BitcoinCoreWallet so get_external_address() is never needed.

raedah:
ACK

Issue #118, opened 2015-07-01
add_tx_notify needs a timeout for giving up


chris-belcher:
add_tx_notify() from BlockchainInterface needs some kind of timeout when it gives up waiting for confirmations. Then a yield generator can call sync_unspent() again to get back to normal and discard a tx that will never confirm.

chris-belcher:
Of course there already is a timeout when using blockr.io
There would need to be another callback function invoked when the timeout is reached which the yield generator can used.

tailsjoin:
Had another tumble session hung on me for some hours due to a failed push to blockr.

https://gist.github.com/anonymous/281738824147f807b82f

chris-belcher:
That must be frustrating.
I've decoded the txhex (using [coinjumble](https://github.com/chris-belcher/coinjumble)) and I can't see anything wrong with it.

Did you manage to see any error message printed out? You could try broadcasting the tx (or I could do it) and see if it gets relayed. If so then we could try having the bot rebroadcast transactions again after the timeout. But if the tx fails every time then what do we do? I guess discard it and start a coinjoin from scratch?

tailsjoin:
No error message, I copy/pasted from the terminal. 

In the past I have manually pushed these txs, even using blockrs api. So a second attempt might be all that's needed for most cases (should wait maybe 10 seconds and then reattempt). If that fails, then starting over would probably be the cleanest way to handle it.

I already spent those utxos, so that signed tx is useless now.

That was 6 of 18 on the session.

PR #369, opened 2015-12-17
Handle pushtx failures. Catch jsonrpcerror exception and retry for tu…

AdamISZ:
…mbler.

~~NOT READY TO MERGE.~~

Previously, pushtx caught the JsonRpcConnectionError, but not the JsonRpcError, which is thrown when, for example, the transaction contains a spent utxo (this was the situation that some users mentioned, I also saw it myself). Crashing on this exception is no big deal in the case of a sendpayment run, because we've finished at that point anyway, but for tumbler, it makes more sense to simply retry the transaction again. To address this, caught the exception and printed the error code to the debug log.

However, this is not ready to merge because there is something clunky about the tumbler restarting, that I'm not sure of the right way to deal with. See finish_callback edit: https://github.com/JoinMarket-Org/joinmarket/pull/369/files#diff-7dca4cc17829f93a2517b5a8dd3ef018L112

This doesn't seem right; add_tx_notify spins up a thread waiting for the walletnotify, and then pushtx() is called. We can restart the transaction if pushtx() fails, but this leaves the notify thread dangling (infinitely I think?). On the other hand, we (presumably?) cannot start the walletnotify thread *after* calling pushtx() because the unconfirm callback might get called too quicky? (I wasn't at all sure about this; if that's wrong, then that might be the right solution). As it is now, it could potentially keep creating new unwanted notify threads for transactions that were never pushed. Advice on the right way to do it?

Edit: the above has now been addressed; see the clarifications in the below comments.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -18,7 +18,7 @@ import bitcoin as btc
 # This can be removed once CliJsonRpc is gone.
 import subprocess
 
-from joinmarket.jsonrpc import JsonRpcConnectionError
+from joinmarket.jsonrpc import JsonRpcConnectionError, JsonRpcError
 from joinmarket.configure import get_p2pk_vbyte, jm_single
 from joinmarket.support import get_log, chunks
 
@@ -646,9 +646,12 @@ class BitcoinCoreInterface(BlockchainInterface):
 
     def pushtx(self, txhex):
         try:
-            return self.rpc('sendrawtransaction', [txhex])
-        except JsonRpcConnectionError:
-            return None
+            txid = self.rpc('sendrawtransaction', [txhex])
+        except JsonRpcConnectionError as e:
+	    return (False, repr(e))
+        except JsonRpcError as e:
+	    return (False, str(e.code) + " " + str(e.message))
+        return (True, txid)
 
     def query_utxo_set(self, txout):
         if not isinstance(txout, list):
--- a/joinmarket/maker.py
+++ b/joinmarket/maker.py
@@ -257,9 +257,11 @@ class Maker(CoinJoinerPeer):
 
     def on_push_tx(self, nick, txhex):
         log.debug('received txhex from ' + nick + ' to push\n' + txhex)
-        txid = jm_single().bc_interface.pushtx(txhex)
-        log.debug('pushed tx ' + str(txid))
-        if txid is None:
+        pushed = jm_single().bc_interface.pushtx(txhex)
+        if pushed[0]:
+            log.debug('pushed tx ' + str(pushed[1]))
+        else:
+            log.debug('failed to push tx, reason: '+str(pushed[1]))
             self.msgchan.send_error(nick, 'Unable to push tx')
 
     def on_welcome(self):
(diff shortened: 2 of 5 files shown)


AdamISZ:
Testing:
regtest and tumbler-test passed.

Then to trigger an error I edited the rpc call to sendrawtransaction, [txhex[::-1]] , and got the expected error:

    2015-12-17 20:21:59,793 [MainThread  ] [DEBUG]  unable to pushtx, reason: -22 TX decode failed
    2015-12-17 20:21:59,793 [MainThread  ] [DEBUG]  Failed to push transaction, trying again.

("trying again" here because tumbler).

chris-belcher:
The notify thread (when using blockr.io) has timeouts and eventually gives up.

AdamISZ:
Right; I was looking at the Core notification thread and didn't see any timeout. Taking another look now, it seems that from the taker side (so tumbler) it won't matter if the unconfirm callback doesn't get called before the confirm callback, and so it would be OK to fire off the notification thread only after a successful push.

Let me know if that's wrong, but if it's right that solves the issue.

chris-belcher:
Ah, well the core notification thread is only started once. It doesn't matter that it lives forever.

AdamISZ:
Ah! Didn't read carefully, yes: we are just adding notify functions to a list in the thread. But is it also true what I said above; if we add_tx_notify *after* doing pushtx() in tumbler, it's OK, because even if we 'miss' the unconfirmed event, it doesn't really matter (just prints to log, doesn't update any state).

chris-belcher:
It probably is OK, yes.

But what's the downside to calling add_tx_notify() before pushtx()
If you're worried about the `txnotify_fun` list growing too long, there is an issue about a timeout for giving up which would reduce the size of that list #118

AdamISZ:
Yes, when I erroneously thought it was a separate thread each time, it was a concern, but just adding to a list is much less of an issue. So leaving it as is (and #118 hopefully being addressed at some point later).

I'll run an extra test with random pushtx() failures (say 1 in 10) as a final sanity check then.

AdamISZ:
Carried out the above mentioned test; set a 30% chance of garbled txhex to be pushed, and tumbler-test with 19 transactions passed OK (received=expected), there were 5 instances of "Failed to push tx" in the log file, with the expected json rpc error.

I consider this ready to merge now. Comments appreciated.

chris-belcher:
You forgot to edit the pushtx() of the blockr code too ;)

AdamISZ:
@chris-belcher  not sure what you mean, this PR was to catch the JsonRpcError, blockr doesn't use JsonRpc

AdamISZ:
OK, I see it now. Hmm it seems unlikely that (a) I didn't pay any attention to that point (pushed[0]) or more to the point, that it hasnt caused problems... I'll take another look.

AdamISZ:
Yes, it's a really bad oversight. Damn.

chris-belcher:
No its good. Nobody complained or noticed which means everyone is using Bitcoin Core :D

I only noticed because I was using blockr with testnet (I would never send my addresses to the blockr server)

Issue #358, opened 2015-12-14
[new-yieldgen-algo] Lower cjfee when a mix depth is void


the9ull:
When one of the mix depths doesn't have a positive balance the cjfee are lowered because the highest one is cutted off by the orders. I think that should be used the highest cjfees.

Mix depth 0 | Mix depth 1 | Mix depth 2
------------ | ------------- | -------------
.2% | .1% | .05%
10Ƀ | 2Ƀ | 0Ƀ

Now the yield generator orders are:

Balance | Fee
------------ | -------------
10Ƀ | .1%
2Ƀ | .05%

chris-belcher:
I think you're right. This is within the spirit of the new-yieldgen-algo as I intended.

BTW, new-yieldgen-algo probably needs a new name given that we have many of them now.

the9ull:
fixed-steps-fee? — From my low knowledge of English

raedah:
I can take this one.

raedah:
This fix was merged in https://github.com/JoinMarket-Org/joinmarket/pull/371 . This can be closed.

PR #371, opened 2015-12-20
fixes issues #106 and #358

raedah:
This resolves issues #106 and #358.

Also fixed an issue where the txfee contribution was not being taken account of when filling orders, which could have rarely been causing crashes if the total amount surpassed the utxo selection.

Added dust exact sized absolute offers. 

ygbasic and ygmixdepth both pass regtest
ygdeluxe only passes when its configuration options are greatly simplified so that it only creates a single order. I guess an issue with how the test selects orders. The error being thrown is  https://bpaste.net/show/e82a247f1608

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import, print_function
 import datetime
 import os
 import time
-import sys
 
 from joinmarket import Maker, IRCMessageChannel
 from joinmarket import BlockrInterface
@@ -74,7 +73,7 @@ class YieldGenerator(Maker):
         mix_balance = self.wallet.get_balance_by_mixdepth()
         if len([b for m, b in mix_balance.iteritems() if b > 0]) == 0:
             log.debug('do not have any coins left')
-            sys.exit(0)
+            return []
 
         # print mix_balance
         max_mix = max(mix_balance, key=mix_balance.get)
@@ -84,9 +83,16 @@ class YieldGenerator(Maker):
                  'maxsize': mix_balance[max_mix] - jm_single().DUST_THRESHOLD,
                  'txfee': txfee,
                  'cjfee': cjfee}
+
+        # sanity check
+        assert order['minsize'] >= 0
+        assert order['maxsize'] > 0
+        assert order['minsize'] <= order['maxsize']
+
         return [order]
 
     def oid_to_order(self, cjorder, oid, amount):
+        total_amount = amount + cjorder.txfee
         mix_balance = self.wallet.get_balance_by_mixdepth()
         max_mix = max(mix_balance, key=mix_balance.get)
 
@@ -95,7 +101,7 @@ class YieldGenerator(Maker):
         log.debug('finding suitable mixdepth')
         mixdepth = (max_mix - 1) % self.wallet.max_mix_depth
         while True:
-            if mixdepth in mix_balance and mix_balance[mixdepth] >= amount:
+            if mixdepth in mix_balance and mix_balance[mixdepth] >= total_amount:
                 break
             mixdepth = (mixdepth - 1) % self.wallet.max_mix_depth
         # mixdepth is the chosen depth we'll be spending from
@@ -103,7 +109,7 @@ class YieldGenerator(Maker):
                                                self.wallet.max_mix_depth)
         change_addr = self.wallet.get_change_addr(mixdepth)
 
-        utxos = self.wallet.select_utxos(mixdepth, amount)
+        utxos = self.wallet.select_utxos(mixdepth, total_amount)
         my_total_in = sum([va['value'] for va in utxos.values()])
         real_cjfee = calc_cj_fee(cjorder.ordertype, cjorder.cjfee, amount)
         change_value = my_total_in - amount - cjorder.txfee + real_cjfee
@@ -112,7 +118,7 @@ class YieldGenerator(Maker):
                        'finding new utxos').format(change_value))
             try:
                 utxos = self.wallet.select_utxos(
-                    mixdepth, amount + jm_single().DUST_THRESHOLD)
+                    mixdepth, total_amount + jm_single().DUST_THRESHOLD)
             except Exception:
                 log.debug('dont have the required UTXOs to make a '
                           'output above the dust threshold, quitting')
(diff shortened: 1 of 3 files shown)


AdamISZ:
W.r.t the None exception, I remember seeing it, and I *seem* to remember it was to do with bots quitting (e.g. the yield generators quitting or crashing), but I can't recall. You'll have to figure out exactly what's causing it from the logs.

As of right now, I think the right thing for us to do is to mark both the -mixdepth and -deluxe versions as "experimental" in the README. Please do keep working on them until at the very least both the regtest and tumbler-test scripts pass. The -basic seems to pass all tests, and I know I (and maybe some others) have been using it on mainnet for a while, seems stable.

> Also fixed an issue where the txfee contribution was not being taken account of when filling orders, which could have rarely been causing crashes if the total amount surpassed the utxo selection.

I'll take a look at this. I seem to remember fixing that exact issue myself already :) It could have been in another branch or something.

chris-belcher:
I have been using the algorithm now known as yg-mixdepth for months. It has been quite stable. I've looked a bit through the file and it doesn't look like much has changed, unless it's something subtle it looks the same to me.

I'd say move them both to another folder actually. New people downloading the project might see all the algorithms and fall victim to the paradox of choice.

raedah:
For testing...
Regtest passes all 3 without issue.

tumblertest is giving the 'ERROR not enough liquidity in the orderbook' for ygmixdepth and ygdeluxe which is the usual state of affairs.

tumbler-test.py passes with ygbasic, but its not using wnb.sh at all. I checked, and that file is not getting executed at all. Additionally, tumbler-test.py passes without issue when wnb.sh is taken out of the configuration.

The 'NoneType' object error I mentioned above comes from recover_from_nonrespondants(), which means its just how the error occurs when something breaks down and stops responding. So just about any breakdown that doesnt result in an error, but rather in a nonrespondant, will occur this way. Including the 'not enough liquidity' error.

The only other resolution left here is to rewrite the order selection handling in tumbler.py.

raedah:
I am against marking anything as experimental. This is a development code base, and its becoming stable. It should be released as such. I am also against giving any yieldgen priority over others. The only reason we have 3 yieldgens is because there was a call to have multiple versions and choice. From my perspective, the older two can be dropped and if we give the user a single one we give them the latest. The other two are being depreciated.

AdamISZ:
> tumbler-test.py passes with ygbasic, but its not using wnb.sh at all. I checked, and that file is not getting executed at all. 

If your bitcoin.conf references it:

    `walletnotify=/wherever/wnb.sh %s`

Then it will get called by a wallet event. If not, something is wrong with your setup. Did you chmod it? Did you try what I suggested in IRC, remove the for loop and just copy the line (I'm not sure if I screwed up the for loop):

    #!/bin/bash
    curl -sI --max-time 0.6 http://localhost:62612/walletnotify?$1
    curl -sI --max-time 0.6 http://localhost:62613/walletnotify?$1
    ....
    curl -sI --max-time 0.6 http://localhost:62620/walletnotify?$1 

>  Additionally, tumbler-test.py passes without issue when wnb.sh is taken out of the configuration. So it would seem this step is not necessary. For ygmixdepth and ygdeluxe, tumbler-test.py logs show 'ERROR not enough liquidity in the orderbook' which is the usual state of affairs. What I found out is that the test does in fact work and complete. Just leave it running. Ran 1 test in 14982.975s. That's 4 hours. 

Not necessary: in theory it is not, but in practice, the interaction of the scripts messes up when you have 7 or so different bots on one machine. If you're using the up-to-date version of tumbler test with -w 0.3 then it definitely won't take more than 5-15 minutes to run, not 4 hours, so clearly something is not right. I did it yesterday to test one of adlai's PRs and it took 400 seconds iirc. I've been using it constantly throughout the last month. Are you looking at the log files to see what's happening?

AdamISZ:
@chris-belcher re: -mixdepth. I was under the impression that significant changes were made between new-yieldgen-algo and yield-generator-mixdepth. It's not as easy as usual to see the diff, but OK.

@raedah re: experimental. Maybe that's not the best word, but I don't want to have to offer support to people trying to use the bots, unless I know myself that they should work.

I know that these comments are orthogonal to the PR (which is a set of fixes), but as you say, this is close to a release. I don't want to release and say "here are three yieldgen bots you can use" when actually we have no idea whether one(?) of them will crash all the time. We don't have to call them "experimental" but it would just be wrong to not mark them somehow as not-heavily-tested.

Testing on mainnet is great, from today I will start testing -mixdepth off this PR, but it's hard to get many transactions. That's why I'm pushing you, specifically, to get tumbler-test running (I don't think anyone else has bothered yet, so it's not hugely surprising you're having trouble), because you need to see what the bots do against large numbers of transactions.

AdamISZ:
tumbler-test: yield-generator-basic passed, yield-generator-mixdepth passed (700 seconds for 15 transactions).
Two edits that help (I will update the test to include this): up the average utxo size for the yield generators to 2.5 from 1.5, to make a low liquidity error in the final transaction not likely/impossible. also use -q 5 as a flag to tumbler in case a low liquidity error does happen.
Trying yield-generator-deluxe now.

AdamISZ:
yg-deluxe passed OK also. 940 seconds for 15 transactions (interesting that it's slower).

I'd note that all three passed *first time*, except for needing to bump up the utxo sizes as mentioned above at the start. So I think that's a good sign.

@chris-belcher do you have any code review comments?

Assuming all is well we could go for language like this in the readme/wiki instructions:

" `yield-generator-basic.py` is the default version, which was previously called `yield-generator.py`, offering one order at a time. `yield-generator-mixdepth.py` is an updated version of the yield generator that was previously held in the newyieldgenalgo branch, and offers a separate order for each mixdepth, allowing you to configure different coinjoin fees for different amounts. `yield-generator-deluxe.py` is a new version of the yield generator with several additional configuration variables as explained in the notes in the script. "

@raedah what do you think about that language? feel free to rewrite it a bit.

AdamISZ:
Looked at this:

> Also fixed an issue where the txfee contribution was not being taken account of when filling orders, which could have rarely been causing crashes if the total amount surpassed the utxo selection.

Ignore my earlier comment "thought I'd already done this". That was confusion with the taker side. Now I've looked into it, I believe the code here is a correct fix (of course, as you mention, it's an unlikely event).

ACK - with the proviso I haven't reviewed the detailed code for the -deluxe version yet (but that proviso applied before this PR anyway).

raedah:
tumbler, basic, Ran 1 test in 844.307s
tumbler, mixdepth,  Ran 1 test in 824.609s
tumbler, deluxe, Ran 1 test in 2699.737s
probably slower because its creating an initial distribution of smaller orders.

PR #370, opened 2015-12-18
Add gaplimit option (-g) to sendpayment.py.

domob1812:
Add the gaplimit option `-g` (that is present for `wallet-tool.py`) also to `sendpayment.py`.

PR #372, opened 2015-12-20
Add gaplimit option for sendpayment.py.

domob1812:
Add the `--gaplimit` option that is present for `wallet-tool.py` also for `sendpayment.py`.

This is https://github.com/JoinMarket-Org/joinmarket/pull/370 rebased onto `develop`.

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -245,6 +245,13 @@ def main():
                       dest='mixdepth',
                       help='mixing depth to spend from, default=0',
                       default=0)
+    parser.add_option('-g',
+                      '--gap-limit',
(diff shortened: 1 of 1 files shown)


chris-belcher:
I think the default should be 5, in line with the default for wallet-tool and yield-generator.

chris-belcher:
Thanks for this.

AdamISZ:
@chris-belcher do I take it that counts as a 'ACK' or 'utACK'? :)

This seems trivial enough to merge immediately, agreed?

chris-belcher:
Agreed. Let's merge.

PR #383, opened 2016-01-04
Adding helpful message when an unhandled path is sent to the http notify request

jgostylo:
I was stuck for a time with a typo in my notify request that I set up in bitcoin.conf.  This message would have been really helpful to let me know what I should check for when running the project to get past my issue.


chris-belcher:
This is a good idea IMO.

raedah:
I have been hit by this one as well. Looks good.

PR #387, opened 2016-01-11
clump coins to max mixdepth, dev

raedah:
With the current 'clumping' oid_to_order code, since it tries to clump into what is already the largest pile, coins wont move through the mixdepths if no large orders come in. Once a small amount of coins leave the largest pile, they will circulate all the way around and end up in the big pile again. I propose the way this should be dealt with is to prioritize by mixdepth number. So that coins in mixdepth0 are spent first, then mixdepth1, and lastly max_mix_depth. This will still tend towards clumping, but always at the last 
mixdepth, causing coins to move through to the max depth when deposited in mixdepth 0.

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -96,17 +96,17 @@ class YieldGenerator(Maker):
         mix_balance = self.wallet.get_balance_by_mixdepth()
         max_mix = max(mix_balance, key=mix_balance.get)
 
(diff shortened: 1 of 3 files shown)


raedah:
regtest deluxe
Ran 2 tests in 144.040s

regtest basic
Ran 2 tests in 145.133s

PR #388, opened 2016-01-11
orderbook change notification for makers

raedah:
so that a yieldgen can modify its offers after other offers come in

### DIFF
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -465,19 +465,25 @@ class OrderbookWatch(CoinJoinerPeer):
             log.debug("Got invalid cjfee: " + cjfee + " from " + counterparty)
         except:
             log.debug("Error parsing order " + oid + " from " + counterparty)
+        self.orderbook_change()
 
     def on_order_cancel(self, counterparty, oid):
         self.db.execute(("DELETE FROM orderbook WHERE "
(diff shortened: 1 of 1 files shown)


chris-belcher:
Widespread use of this feature could create never-ending feedback loops if two bots modify their orders based off each other, which would flood up whatever messaging channel we use.

We already break the IRC protocol rule of not replying to a PRIVMSG with another PRIVMSG: 
https://tools.ietf.org/html/rfc1459#section-4.4.2

> The difference between NOTICE and PRIVMSG is that automatic replies must never be sent in response to a NOTICE message.  This rule applies to servers too - they must not send any error reply back to the client on receipt of a notice.  The object of this rule is to avoid loops between a client automatically sending something in response to something it received.  This is typically used by automatons (clients with either an AI or other interactive program controlling their actions) which are always seen to be replying lest they end up in a loop with another automaton.

Right now the JM protocol is set up so that these loops never happen. That would change with this PR. For this reason I'm against going down this road (I know anyone could include it in their own bot without uploading it here, but I hope they don't)

raedah:
A valid concern. I do have safeguards such as update interval and price floor settings. I am interested in doing this safety in order to improve market efficiency.

Issue #390, opened 2016-01-14
invalid literal


solsimpais:
Getting this error message:
```
 1: {},
 2: {},
 3: {}}
2016-01-14 10:21:55,203 [Thread-3    ] [DEBUG]  coinjoining 16410404 satoshi
Exception in thread Thread-3:
Traceback (most recent call last):
  File "C:\TOOLS\Python27\lib\threading.py", line 810, in __bootstrap_inner
    self.run()
  File "tumbler.py", line 320, in run
    self.init_tx(tx, self.balance_by_mixdepth[tx['srcmixdepth']], sweep)
  File "tumbler.py", line 280, in init_tx
    self.create_tx()
  File "tumbler.py", line 229, in create_tx
    cj_amount, self.tx['makercount'])
  File "tumbler.py", line 142, in tumbler_choose_orders
    self.ignored_makers + active_nicks)
  File "C:\TOOLS\JOINMARKET\joinmarket\support.py", line 247, in choose_orders
    'counterparty'] not in ignored_makers]
  File "C:\TOOLS\JOINMARKET\joinmarket\support.py", line 168, in calc_cj_fee
    real_cjfee = int(cjfee)
ValueError: invalid literal for int() with base 10: '5E-9'
```
Thanks

AdamISZ:
This is the fault of one of the makers in the transaction, who's publishing an absorder with a non-integer amount (should be in satoshis).

 We should be catching this exception, adding the maker to the ignored_makers list and retrying the transaction.

Thanks for the report.

solsimpais:
Thanks, I have changed joinmarket/taker.py as you have suggested and tumbling is on its way! I just get:a few such messages along the way:
```
2016-01-14 17:07:01,244 [MainThread  ] [DEBUG]  Got non integer coinjoin fee: 8783 for an absorder from Fixolegib
```

raedah:
This seems to be blocking the orderbook table from updating at all.
Sending !cancel 0 to the pit from any user generates:
[DEBUG]  !cancel ValueError("invalid literal for int() with base 10: '5E-9'",)
This makes me believe existing users are no longer able to track valid orderbook updates.
A fix will need to make it to master.

AdamISZ:
Fix applied to master, see tag v0.1.1

PR #391, opened 2016-01-14
disallow non-integer absorders in on_order_seen

AdamISZ:
See #390 .

It makes me realise that any protocol 'expansions' have to pay particular attention to on_order_seen.

This is basically assuming that we don't need to catch a valueerror anywhere else based on the value of 'cjfee', as long as we don't let anything invalid come in in the first place.

Minimal testing done, but it seems uncontroversial. ACKs or NACKs appreciated.

### DIFF
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -456,6 +456,13 @@ class OrderbookWatch(CoinJoinerPeer):
                        "from {}").format
                 log.debug(fmt(minsize, maxsize, counterparty))
                 return
+            if ordertype == 'absorder' and not isinstance(cjfee, int):
+                try:
+                    cjfee = int(cjfee)
+                except ValueError:
+                    log.debug("Got non integer coinjoin fee: " + str(cjfee) +
+                            " for an absorder from " + counterparty)
+                    return
             self.db.execute(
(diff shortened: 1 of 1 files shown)


Pigeons:
tested briefly with the existing broken order in the books and it worked

raedah:
It will invalidate a lot of valid offers on the offerbooks that look like they are coming in as a str? Such as
Got non integer coinjoin fee: 2723 for an absorder from Boyohahev
While there is really only one truly invalid offer on the books currently.
Got non integer coinjoin fee: 5e-09 for an absorder from Hurufohoz_

AdamISZ:
> It will invalidate a lot of valid offers on the offerbooks that are coming in as a string. Such as
Got non integer coinjoin fee: 2723 for an absorder from Boyohahev

Damn! That was stupid rushing on my part. Thanks for noticing/catching.

Your fix seems correct, I will try it.

> The maker code should also have a validation check for this too so that these offers are killed before they ever get pushed to the channel, and sparing the rest of us from extra error displays. Ill add this to the yieldgen sanity check.

No, I don't see that as the same thing (of course, please do apply any checks you like to yieldgen code). The point is that people can (and do) make their bots print out all kinds of rubbish.

AdamISZ:
I went ahead and merged it with that fix. There is currently a stream of people complaining about this :)

Issue #395, opened 2016-01-20
yield-generator-deluxe.py txfees causing negative net satoshis


grubles:
It seems that using -deluxe.py with the default configuration results in loss of satoshis. 

`2016-01-19 19:44:54,717 [Thread-3    ] [DEBUG]  earned = -158`

min_revenue setting: 
`min_revenue = random.randrange(txfee_low, 12000)  # varied`

raedah:
An quick fix would be changing txfee_low to txfee_high.
Thats going to make all the offers on the lower txfee side more expensive.
But no more satoshi loss.

PR #397, opened 2016-01-21
fix issue 395

raedah:
#395


raedah:
regtest
Ran 2 tests in 145.022s

chris-belcher:
In future some of these PRs and commits should have more descriptive messages so somebody reading later will find it easier to understand what they changed and why

PR #402, opened 2016-01-24
reformat with yapify, no other changes

raedah:
Many files needed reformatting with yapf. I used the command 

```yapf -i *.py; yapf -i --style google *.py```

No other changes were made.


AdamISZ:
There is something not quite right here; I suspect it's that the yapfify settings you're using are not the same as the ones originally used by @ghtdak .

Here's why I say that; I noticed a *lot* of changes, even in for example, old_mnemonic.py, which definitely hasn't changed since the last re-format.

A concrete example: this line split in your PR: https://github.com/JoinMarket-Org/joinmarket/pull/402/files#diff-f4873928b5e1da8793421866c1a92d72L284

is an edit to this commit from @ghtdak : https://github.com/JoinMarket-Org/joinmarket/commit/d8b14dad446bae74e3c28a3b1e0ee949c9edf72c

raedah:
Do you have a yapf cmd you can run on support.py that doesnt produce changes? I dont.

AdamISZ:
Looking at: https://github.com/JoinMarket-Org/joinmarket/pull/340#issuecomment-161161378 , it seems that @ghtdak used `yapf -i -r --style google .` . I don't know if that is going to produce any difference (I wouldn't have thought so; but then again, you are doing `yapf` without the google flag first, and then with?), or whether there might be some version difference in yapf. Sorry for not investigating in more detail, perhaps someone else wants to chime in?

raedah:
the -r flag just means  'run recursively over directories' so that shouldnt matter. I run it once without the --style google option and then with the --style google options because thats how Ive found to get consistent output. The first command will make modifications that the second one will not, so I run it in that order to make sure all modifications are applied. As for why @ghtdak original commits dont appear to have the changes, has anyone seen him lately?

AdamISZ:
Of course, I understood -r :) As to the 2 phase process, I honestly, on reflection, can't see any logical reason why that *wouldn't* result in a different output. "The first command will make modifications that the second one will not, so I run it in that order to make sure all modifications are applied." <-- why do you think this is what's required? I mean, I presume it's a matter of arbitrary choice whether we use the google style or whatever the default style is, but if the google style is what we used on the first run through it doesn't make sense to apply another style also, thus resulting in a huge block of changes again; does it?

adlai:
I think we should minimize this kind of change, but am not opposed to this one.

@raedah please rebase it onto 816242d (or whichever will be the latest on develop)

AdamISZ:
@adlai Did you read the details above? For example, this changes the file `old_mnemonic.py`, which was of course only changed once by @ghtdak 's full repo reformat. Do you really think we should change the entire repo to a new format? What's the point?

adlai:
I'm not in favor of doing it needlessly. We should at least know which command @ghtdak used, for consistency. I'm not opposed to closing this PR and just fixing formatting locally as we go forwards.

AdamISZ:
> We should at least know which command @ghtdak used

I dug it up from #340 and put it in a comment above.

raedah:
current status is we are unable to reproduce the output from @ghtdak. We need to confirm the command he used as well as the yapf version and find out what other variables could be causing the difference.

raedah:
<ghtdak> my yapf master is commit: 8fc7cf9030883267f06149229f822187490639dd branch master (11/13/15)

I pushed a new version, based on the latest develop, using 'yapf -i -r --style google .'  with the latest version of yapf straight from the yapf master branch.

PR #406, opened 2016-01-29
Loop on irc.connect() in order to survive to connection errors

the9ull:
My yield-generator dies on connection errors. I solved the problem in this way

> 2016-01-29 01:55:26,584 [MainThread  ] [DEBUG]  None
2016-01-29 01:55:26,584 [MainThread  ] [DEBUG]  Traceback (most recent call last):
  File "./yield-generator-mixdepth.py", line 315, in main
    irc.run()
  File "/home/joinmarket/multijm/2/joinmarket/irc.py", line 617, in run
    self.sock.connect(self.serverport)
  File "/home/joinmarket/multijm/2/joinmarket/socks.py", line 391, in connect
    self.__negotiatesocks5(destpair[0], destpair[1])
  File "/home/joinmarket/multijm/2/joinmarket/socks.py", line 255, in __negotiatesocks5
    raise Socks5Error(_socks5errors[min(9, ord(resp[1]))])
Socks5Error: 'TTL expired'

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -185,17 +185,18 @@ def main():
                                 "BLOCKCHAIN", "blockchain_source"),
                             password=nickserv_password)
     maker = YieldGenerator(irc, wallet)
-    try:
-        log.debug('connecting to irc')
-        irc.run()
-    except:
-        log.debug('CRASHING, DUMPING EVERYTHING')
-        debug_dump_object(wallet, ['addr_cache', 'keys', 'seed'])
-        debug_dump_object(maker)
-        debug_dump_object(irc)
-        import traceback
-        log.debug(traceback.format_exc())
-
+    while True:
+        try:
+            log.debug('connecting to irc')
+            irc.run()
+        except:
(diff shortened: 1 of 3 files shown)


chris-belcher:
This would mean that if the bot crashes for any reason, even unrelated to Socks5 stuff, you'll probably never see the traceback. Some of the times the bot crashes is because things are seriously wrong and there's no point even trying to recover.

I'm against, I think a better way is to use this part which recovers and loops from connection errors https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/irc.py#L640

Probably the best way is to make ProxyError inherit from IOError instead of Exception. https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/socks.py#L44

adlai:
Also, this PR should probably point into develop rather than master.

adlai:
@chris-belcher implemented the proper approach in 74980

chris-belcher:
Oh yeah I should've linked here too.

PR #408, opened 2016-01-30
Created broadcast-tx.py and hacked push_tx() into working

chris-belcher:
Not many people know that all makers running today are able to broadcast transactions sent to them. This script can be used to ask a maker to broadcast one of your transactions. It's not much more than a proof of concept.

Partly inspired to do this now by threads like this http://www.donotlink.com/i2e5

To add this to takers, blockchaininterface first needs to be edited to have a configurable timeout after which it will invoke a callback as in #118, this is because makers might not broadcast the taker's transaction or there may have been an error. The taker needs to know if the transaction has been seen on the network.

### DIFF
--- a/broadcast-tx.py
+++ b/broadcast-tx.py
@@ -1,0 +1,85 @@
+#! /usr/bin/env python
+from __future__ import absolute_import
+
+import sys
+import threading
+from optparse import OptionParser
+import time
+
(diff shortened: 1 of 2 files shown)


AdamISZ:
Is there any reason not to merge this? I guess it's not urgent?

PR #412, opened 2016-02-02
require liquidity providers to provide liquidity

adlai:
certain malicious and/or incompetent liquidity providers send inputs totalling less than the coinjoin amount! this leads to a change output of zero satoshis, so the invalid transaction fails harmlessly; now we fail earlier, with a clear message.

reopens #407 which currently stands at ACK from me, and utACK from @chris-belcher and @AdamISZ


adlai:
Reopening onto develop instead of master.

PR #417, opened 2016-02-04
recover from selection of spent utxos

### LINKED ISSUE #74: Gracefully recover when utxo already spent
I came across a situation just now which caused a yield generator to crash.

The yieldgen Y had a deal done with a taker A, after Y had sent the signature but before A broadcast the tx, a second taker B started a coinjoin deal with Y.

Meanwhile A had broadcast the tx and the bitcoin core updated it's chain state

Then when Y went to do a sanity check, it saw the utxo was already spent, and paniced and quit.

The sanity checking code is here https://github.com/chris-belcher/joinmarket/blob/master/lib/maker.py#L44

What the bot should actually do is gracefully remove the spent utxo from its internal wallet and call oid_to_order() again. If not possible to fill the order it should send back an error and take the reputation hit as in https://github.com/chris-belcher/joinmarket/issues/57 . This means sending back an error in the early stages of the deal should be punished less harshly than timing out with no messages in the later stages.

Note: the other sanity checking code that checks the utxos are labelled with the correct value, should remain.

adlai:
fixes #74

### DIFF
--- a/joinmarket/maker.py
+++ b/joinmarket/maker.py
@@ -46,23 +46,31 @@ class CoinJoinOrder(object):
         self.txfee = order['txfee']
         self.cjfee = order['cjfee']
         log.debug('new cjorder nick=%s oid=%d amount=%d' % (nick, oid, amount))
-        self.utxos, self.cj_addr, self.change_addr = maker.oid_to_order(
+
+        def populate_utxo_data():
+            self.utxos, self.cj_addr, self.change_addr = maker.oid_to_order(
                 self, oid, amount)
-        self.maker.wallet.update_cache_index()
-        if not self.utxos:
-            self.maker.msgchan.send_error(
+            self.maker.wallet.update_cache_index()
+            if not self.utxos:
+                self.maker.msgchan.send_error(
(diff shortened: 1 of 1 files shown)


AdamISZ:
Interesting!

Were you able to replicate #74 in some reliable way (not the kind of artificial way I did, but a scenario that can happen in real life)? I wonder could you give some explanation of the thinking?

adlai:
I don't have formal test cases, although I've replicated the crash manually by joining with myself.

FWIW I've been running this code without trouble for longer than the PR has existed.

chris-belcher:
I agree with the concept, looked at the code and think it's good.

AdamISZ:
ACK; merging.

OK, having read the code now, I can see the logic and it looks clear. This puts to one side the complaint that we had in #74 that it is suboptimal to produce a patch without knowing the exact cause; first, there may be more than one cause (spending is plausible; did you mean from outside this joinmarket bot?), but in any case, this code provides a logical method of recovery, whatever the cause. I can't see a downside to it.

Ran a tumbler test OK.

Can I get an explanation for how RLock (reentrant) changes things here, and what the rationale is?

PR #419, opened 2016-02-07
stop yield-generator-mixdepth announcing absorders if they havent bee…

chris-belcher:
…n modified

To stop stuff like this where the same absorder is announced again and again

	<Wipiziyik> !absorder 0 2731 68179088 5000 8182 ~
	<Lebiyikon> !absorder 0 2731 68179088 5000 8864 ~
	<Wipiziyik> !absorder 0 2731 68179088 5000 8182 ~
	<Lebiyikon> !absorder 0 2731 68179088 5000 8864 ~
	<Wipiziyik> !absorder 0 2731 68179088 5000 8182 ~
	<Povovepix> !cancel 2
	<Wipiziyik> !cancel 1
	<Povovepix> !absorder 0 2731 68179088 5000 8182!relorder 1 68179089 897087792 5000 0.00015 ~
	<Wipiziyik> !absorder 0 2731 31090346 5000 7500 ~
	<Wipiziyik> !absorder 0 2731 68179088 5000 8182!relorder 1 68179089 802405132 5000 0.00015 ~
	<Povovepix> !absorder 0 2731 68179088 5000 8182 ~
	<Povovepix> !absorder 0 2731 68179088 5000 8182 ~
	<Wipiziyik> !absorder 0 2731 68179088 5000 8182 ~
	<Wipiziyik> !absorder 0 2731 68179088 5000 8182 ~
	<Povovepix> !absorder 0 2731 68179088 5000 8182 ~

The cause of the bug is this

     oldorders = [o for o in oldorders if o['ordertype'] == 'relorder']

which means `oldorders` contains only relorders, but later this line will be called

    old_abs = [o for o in oldorders if o['ordertype'] == 'absorder']

since oldorders only contains relorders, `old_abs` will always be empty

### DIFF
--- a/yield-generator-mixdepth.py
+++ b/yield-generator-mixdepth.py
@@ -218,15 +218,15 @@ class YieldGenerator(Maker):
 		'''
 
         myorders = self.create_my_orders()
-        oldorders = self.orderlist
+        oldorderlist = self.orderlist
         if len(myorders) == 0:
-            return ([o['oid'] for o in oldorders], [])
+            return ([o['oid'] for o in oldorderlist], [])
 
         cancel_orders = []
         ann_orders = []
 
         neworders = [o for o in myorders if o['ordertype'] == 'relorder']
-        oldorders = [o for o in oldorders if o['ordertype'] == 'relorder']
+        oldorders = [o for o in oldorderlist if o['ordertype'] == 'relorder']
         #new_setdiff_old = The relative complement of `new` in `old` = members in `new` which are not in `old`
         new_setdiff_old = [o for o in neworders if o not in oldorders]
         old_setdiff_new = [o for o in oldorders if o not in neworders]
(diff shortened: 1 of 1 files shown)


raedah:
Fix looks okay. You could also use this code here https://github.com/JoinMarket-Org/joinmarket/blob/master/yield-generator-deluxe.py#L510

chris-belcher:
Code like that used to be used many months ago, it would have the same problem of announcing orders which already existed. If an order-id is changed that code will see it as a changed order even if all the other properties are the same.

AdamISZ:
ACK. Seems entirely uncontroversial (thanks for the explanation btw, helps).

@chris-belcher @adlai You don't always have to wait for me to merge, right? :)

If there're no more comments tomorrow I'll merge.

raedah:
> If an order-id is changed that code will see it as a changed order even if all the other properties are the same.

The absorder is always oid 0. It doesnt change. Its hard coded on line 139.

chris-belcher:
@raedah Yes but not for relorders.

raedah:
You are not checking for oid changes on relorders here.

chris-belcher:
When I wrote this https://github.com/JoinMarket-Org/joinmarket/pull/419#issuecomment-181028133 I was only referring to your suggestion of this code https://github.com/JoinMarket-Org/joinmarket/pull/419#issuecomment-180943608 , which doesn't have anything to do with the PR edits.

raedah:
ACK

PR #422, opened 2016-02-08
Obfuscate maker bot identities

nukaa:
Currently the codebase makes weak attempts to disclose the long term identity of a  maker, which somewhat irks me.

These patches make the bot assume a new identity on every IRC reconnect. While it can still be somewhat tracked via its published summary amounts it makes it at least somewhat non-trivial.

Future improvements: To make the identity obfuscation more robust, we'd need a split personality - basically small number of sybils, but in a good way.

The feature is optional and disabled by default.

I am aware this is a possible tool of sybil attackers. That would happen one way or another, massive sybils must be prevented through effective means - market forces, coin age and privately, F2F established identities - not nicknames / u@h.

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -116,11 +116,16 @@ socks5 = false
 socks5_host = localhost
 socks5_port = 9050
 #for tor
-#host = 6dvj6v5imhny3anf.onion
+#newnym = true
+#newnym_delay = 60
+#tor_host = localhost
+#tor_port = 9051
+#tor_pass =
 #port = 6697
 #usessl = true
 #socks5 = true
 maker_timeout_sec = 30
+reconnect_delay = 30
 
 [POLICY]
 # for dust sweeping, try merge_algorithm = gradual
--- a/joinmarket/irc.py
+++ b/joinmarket/irc.py
@@ -8,6 +8,7 @@ import threading
 import time
 import Queue
 
+from ConfigParser import NoOptionError
 from joinmarket.configure import jm_single, get_config_irc_channel
 from joinmarket.message_channel import MessageChannel, CJPeerError
 from joinmarket.enc_wrapper import encrypt_encode, decode_decrypt
@@ -497,8 +498,9 @@ class IRCMessageChannel(MessageChannel):
                 if self.on_nick_leave:
                     self.on_nick_leave(nick)
         elif _chunks[1] == '433':  # nick in use
-            # self.nick = random_nick()
             self.nick += '_'  # helps keep identity constant if just _ added
+            if self.newnyms:
+                self.nick = random_nick()
             self.send_raw('NICK ' + self.nick)
         if self.password:
             if _chunks[1] == 'CAP':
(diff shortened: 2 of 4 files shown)


joinmarket/irc.py
```diff
@@ -583,6 +585,20 @@ def __init__(self,
         self.socks5_port = int(config.get("MESSAGING", "socks5_port"))
         self.channel = get_config_irc_channel()
         self.userrealname = (username, realname)
+        self.reconnect_delay = 30
+        self.newnyms = False
+        try:
+            self.reconnect_delaty = int(config.get("MESSAGING", "reconnect_delay"))
```
chris-belcher: mispelling here, should be reconnect_delay

joinmarket/maker.py
```diff
@@ -213,7 +213,8 @@ def __init__(self, msgchan, wallet):
         self.active_orders = {}
         self.wallet = wallet
         self.nextoid = -1
-        self.orderlist = self.create_my_orders()
+        if not self.msgchan.newnyms:
+            self.orderlist = self.create_my_orders()
```
chris-belcher: if create_my_orders() is now called in on_welcome() then there's no need to have it here too

joinmarket/configure.py
```diff
@@ -116,11 +116,16 @@ def jm_single():
 socks5_host = localhost
 socks5_port = 9050
 #for tor
-#host = 6dvj6v5imhny3anf.onion
```
chris-belcher: Why is the onion host gone? I think this is a mistake.

joinmarket/configure.py
```diff
@@ -116,11 +116,16 @@ def jm_single():
...
+#newnym = true
+#newnym_delay = 60
+#tor_host = localhost
+#tor_port = 9051
```
chris-belcher: A better name for these may be `tor_control_host`, at first I thought it was another name for socks5 port.

joinmarket/maker.py
nukaa: This is on purpose - it's called in on_welcome only when newnyms is enabled. The reasoning is that custom maker scripts might not expect this, so when newnyms iss disabled, it falls back to the original order (create orders on instantiation).

joinmarket/irc.py
nukaa: Fixed, force-updated.

chris-belcher:
> Currently the codebase makes weak attempts to disclose the long term identity of a maker,

What do you mean by this? There is no identity attached, JoinMarket does not ask for any proof of ID before running. I assume you mean something else like IP address or bitcoin amounts(?)

Why does your code only create a new tor circuit when the bot disconnects? It seems like an arbitrary time to do it. Would it maybe be better to disconnect and reconnect at random time intervals? I'm not sure cycling to a new tor circuit is helpful when the bitcoin amounts can be correlated very easily. Happy to be proven wrong though.

nukaa:
@chris-belcher 
Summary of IRC convo:

This patch is about the nyms of the comms channel, not with regards to JM protocol itself. Current weak attempts is retaining a nickname across reconnects, not changing Tor circuit on reconnect and keeping same orders across reconnect. Meaning the long term identity of a single pseudo nym is eagerly disclosed.

The patch does the opposite of "eager". First, it's much harder for passive adversary to even determine you're using tor (which is now trivial). Second, if he starts probing Tor entry guards he can't (easily) tell the returning bot is the same he's trying to traffic confirm.

Note that advertised orders need some fixing up to make it a bit harder, I'll keep elaborating ideas on irc and mention here if it's something worthwhile.

Regarding traffic confirmation, see https://blog.torproject.org/blog/traffic-correlation-using-netflows and related google searches.

In short, with interactive protocols like IRC it works like this: First, you confirm an entry guard. If you can confirm with high degree of certainty, you issue a court order on said guard to acquire netflow packet dumps. After that, it's a gameover for whomever is behind said guard.

joinmarket/configure.py
nukaa: Using hidserv publicly reveals the fact you're using tor, plus makes your bot unstable. Use HS only on irc networks which don't allow direct tor connections. Opionated I know, I'll revert this if you strongly disagree.

nukaa: Made check for .onion when newnym is used (and bail on it).

joinmarket/configure.py
nukaa: added tor_ctrl_* everywhere (_control_ would cause too lines i'm too lazy to break)

Issue #418, opened 2016-02-07
Support coinjoin transactions with no change output


raedah:
As mentioned in pull requests #407 and #413, the current code is always assigning a change output, even in cases where there is no change, so the change output has a value of zero which creates an invalid transaction. #413 patches it so that it fails early and with an error msg. I am opening this issue to track fixing this case for when there is no change and the transaction can go through without a change output needed.

https://github.com/JoinMarket-Org/joinmarket/pull/407#issuecomment-177021387
Quoting belcher here: "coinjoin sending with no change address (i.e. sweeping) has never really been supported. It's not clear where the maker fees would go. Although it might be useful in the rare occasion when the maker gets an order for exactly the right coinjoin amount."

raedah:
This will be useful for allowing yield generators to make offers which clear out mix depth dust, as well as other offers which are intended to match utxo size with no change.

Issue #171, opened 2015-08-03
A new protocol


chris-belcher:
Issue which combines all the features we want when the protocol is updated.

Talking to some people, it seems doing a breaking change, a "hardfork", is quite an unpopular idea. It would split the liquidity, updating would end up having to be rushed by some people, giving not enough time to check the code. So it seems the compromise solution is to code it that new bots will accept both protocols and the old protocol is gradually phased out.

Stuff to fix:
#88 Fix bug where maker must known the private key of the coinjoin address.
#83 Using p2sh / multisig as inputs.
#45 Some cleanup, changing names, adding a field
#29 Minimum other maker count field
https://github.com/chris-belcher/joinmarket/pull/90#issuecomment-108648823 Allowing taker to not authenticate
#120 Fixing miner fees. Changing maker's tx fee contribution to a fee/kB and having a new field of "lowest fee/kB tx this maker is willing to sign off for"

It should also be noted that the patientsendpayment.py script *does not work at all* in the current protocol, because makers must know the private key of the coinjoin address they're sending to.

#90 Some code already written
The branch https://github.com/chris-belcher/joinmarket/tree/newprotocol

chris-belcher:
More code / bugfix to add #170

chris-belcher:
Allow option of maker not having a change address https://github.com/chris-belcher/joinmarket/issues/27#issuecomment-74863580

dcousens:
@chris-belcher is the algorithm/protocol written up anywhere?

chris-belcher:
Not really

kristovatlas:
I might try to help with documentation in the future...

chris-belcher:
Worth noting that this won't be the last time we update the protocol. We'll probably end up adding segwit and op_schnorr to JoinMarket eventually too. So the code for new protocol should be written in a way that makes adding more order types easier.

chris-belcher:
And anything we decide on regarding this issue will be included in a new protocol too: #156

AdamISZ:
#568  is now a protocol change (so as to find a logical way to use multiple messaging servers without impersonation).

I am not sure what people feel comfortable with as a reasonable set of changes; anti- #156 is priority 1, it seems; in line with #343 goals for 0.2 I think the above multi-mc should also be included. Re: the above list, I'm not sure, but maybe we should hash out, here, what set is both highly desired, and also a realistic goal for a new version.

chris-belcher:
[...]

Issue #27, opened 2015-02-01
implement sweep for patientsendpayment.py


chris-belcher:

chris-belcher:
Problem with this. The protocol currently assumes the maker always has a coinjoin address.

Could be a simple hack with if statement checking if change_value == 0, then dont even add a change address.

PR #423, opened 2016-02-10
skip empty change addresses for makers

adlai:
works towards #418 but does not enable this behavior maker-side; that would require recognizing when a subset of utxos match the cjout size precisely, and short-circuiting past the addition of a dust amount.


chris-belcher:
I have read the diff and I think it's not wrong.

chris-belcher:
Could someone else have a look at this, it contains modifications to `verify_unsigned_tx()` which is quite important for making sure all the coins go to where they should.

AdamISZ:
Taking a look now, will feed back if I have a comment.

chris-belcher:
We can surely merge this now?

adlai:
This should be merged, as it addresses #27 which is required by #171

PR #428, opened 2016-02-16
fix small profit default config

raedah:
This fixes the default config for deluxe so that users dont initially get joins for only a single satoshi profit.


chris-belcher:
Seems ok.

PR #430, opened 2016-02-18
Fix crash in yield generator mixdepth when no unconfirm notify arrives

chris-belcher:
One edit that moves a line into the if statement.

Aside: How does it happen that maker's don't see the unconfirmed transaction sometimes? It gets relayed to the miners but not to us for some reason.

### DIFF
--- a/yield-generator-mixdepth.py
+++ b/yield-generator-mixdepth.py
@@ -268,9 +268,9 @@ class YieldGenerator(Maker):
         if cjorder.cj_addr in self.tx_unconfirm_timestamp:
             confirm_time = int(time.time()) - self.tx_unconfirm_timestamp[
                 cjorder.cj_addr]
+            del self.tx_unconfirm_timestamp[cjorder.cj_addr]
         else:
             confirm_time = 0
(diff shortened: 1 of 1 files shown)


raedah:
Please describe why this is needed.

chris-belcher:
If an unconfirmed transaction does not make its way to the yield generator, the on_tx_unconfirmed() function wont be run. That would mean that the `self.tx_unconfirm_timestamp` dictionary won't contain `cjorder.cj_addr` and therefore when `del self.tx_unxconfirm_timestamp[cjorder.cj_addr]` is called it will raise an IndexError. Moving the del statement to inside the if statement means it will only execute if the dictionary contains `cjorder.cj_addr`.

raedah:
Sounds good, thanks. Ill put this into my other ygs.

chris-belcher:
Add it in sure. The del is only used to clear up memory, and to not clutter up the debug variable dump when the bot crashes. It makes no difference at all to the operation of yield generators.

raedah:
ACK

PR #433, opened 2016-02-19
fix comment placement in support.py and remove encryption_protocol.tx…

AdamISZ:
…t as no longer needed here.

No code change.


chris-belcher:
agreed

PR #434, opened 2016-02-26
add a yield-generator with static fee

veqtrus:


raedah:
This seems mostly code from -basic. I just see the OptionParser additions, and a change to an absolute offer.  Would it be possible to make the offer type an option as well, and then merge this with basic? Also, would you be willing to check the other yield-generators and see if you can add the OptionParser to them for their options? At least for the standard options that make sense.

veqtrus:
Current -basic needs to choose its own minimum size to stay profitable; what should it do after merging with -absolute?

I will do the OptionParser stuff for other generators in a separate PR.

raedah:
Here is what I would do
`minsize = int(1.2 * txfee / float(cjfee))`
The only thing they are really going to be changing here is the 1.2, so move the rest of the code down below the settings, and make the minsize just be the 1.2.
`minsize = 1.2`

And I assume you have some kind of fee selector? like fee_type = 'absolute', so 
`if fee_type = 'absolute':
    minsizex = minsize
elseif fee_type = 'relative':
    minsizex = int(minsize * txfee / float(cjfee))`

something like that perhaps

chris-belcher:
if you choose `fee_type = "relative"` then isn't this the same as yield-generator-basic ?

raedah:
actually, fee_type isnt even needed because you can just check the type of cjfee
`
if isinstance(offer['cjfee'], int)
elif isinstance(offer['cjfee'], float)
`

chris-belcher:
@veqtrus do you run this day-to-day?
Just trying to find out if anyone's tried this yet.

veqtrus:
Yes, but I'm going to merge it with -basic as suggested. If you want to
close the PR for now it's OK.

PR #435, opened 2016-02-27
add yg oscillator

raedah:
YG that dynamically raises or lowers fee based on the users transaction volume, and uses the transaction history from the csv statement. Fee rates are fully configurable by the user. I guessed at some sane defaults. This will benefit the market by rotating less used offers to lower prices, and more used offers to higher prices, diversifying which maker utxos are being used. This solves the problem for the user of having to manually adjust their offer pricing, and allows them to better gather data about their pricing strategy. Included with this, csv files now show entries for 'unconfirmed' transactions which gets updated once they are confirmed. Separate statement tracking for different wallets, csv file name is now yigen-statement-yourwallet.csv. Enhanced mix depth selection, Slinky Clumping, which first pushes all coins towards the largest mixdepth, and then spends into the next mixdepth until it is full, ad infinitum. This insures the largest amount of coins will be available to offer, and the largest selection of utxos will be available to select from.

### DIFF
--- a/yield-generator-oscillator.py
+++ b/yield-generator-oscillator.py
@@ -1,0 +1,727 @@
+#! /usr/bin/env python
+from __future__ import absolute_import, print_function
+
+import os
+import sys
+import datetime
+import time
+import random
+import decimal
+import ConfigParser
+import csv
+import threading
+import copy
+
+from joinmarket import Maker, IRCMessageChannel, OrderbookWatch
+from joinmarket import blockchaininterface, BlockrInterface
+from joinmarket import jm_single, get_network, load_program_config
+from joinmarket import random_nick
+from joinmarket import get_log, calc_cj_fee, debug_dump_object
+from joinmarket import Wallet
+
+config = ConfigParser.RawConfigParser()
+config.read('joinmarket.cfg')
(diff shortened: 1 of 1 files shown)


riclas:
i've been using it since it was in development and it's very stable. the fee algorithm works really well to get to know the market.

chris-belcher:
Alright sounds good.

wozz:
I've tested a slightly older version and it seems good from a stability perspective.  I'm still unsure if I like the algorithm.

raedah:
@wozz One of the last tweaks (after getting all the underlying mechanics right) was changing from just a fib scale increases starting at 0, to user defined increases within any static or percentage increment which are within a min and a max range and defined on a per offer basis. So you can make it whatever you want now. Let me know if you can think of a desired behavior that isn't supported yet.

wozz:
ok, that makes more sense... I'll try it out soon

raedah:
rebased.

AdamISZ:
Passed tumbler-test with 16 transactions, seems reasonable to merge now, since others have tested.

Ongoing discussions about how to refactor, maybe also create separate repo for more yg variants, but not urgent.

Issue #436, opened 2016-02-27
Always call unconfirmfun() in BitcoinCoreInterface even if we first see a confirmed tx


chris-belcher:
The code here https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/blockchaininterface.py#L426-L437 should keep track of whether unconfirmfun() was previous invoked when a confirmed transaction is seen, and if not, call it before confirmfun().

This will mean yield generators wont have to deal with the logic of unconfirmfun() sometimes not appearing. They will know unconfirmfun() will always be called before confirmfun().

Also this can make JoinMarket work better with -blocksonly because the logic wont need to be changed. Blocksonly could be very useful for bandwidth-constrained tumbler.py users (for example Tails OS users which run over tor) Depending on how bandwidth-constrained tor is, it may be worth having tailsjoin scripts choose blocksonly by default.

dreboli:
-blocksonly also seems to take off a lot of load from the CPU. I have about 5% CPU load on a raspi b+ between blocks, and there's 100% load for about a minute after a new block.

In the current state, does joinmarket work with -blocksonly option? (sorry it's still not clear to me after reading the comment above)

dreboli:
Without -blocksonly the average CPU load is about 90% on the pi.

chris-belcher:
In the current state it does not work, it just needs a small edit to make unconfirmfun() always be called.

When that edit is made, the tumbler.py bot would probably benefit from it. The script can just send out coinjoin transactions and wait until they're confirmed, no need to see the unconfirmed txes.

blocksonly isn't appropriate for yield-generator because it needs to see unconfirmed txes.

sendpayment.py doesnt even register itself to be notified for unconfirmed txes blocksonly doesnt do anything.

patientsendpayment.py needs to know about unconfirmed txes so blockonly probably cant be used. (Although nobody uses this script anyway because it doesnt work for other reasons)

PR #437, opened 2016-02-29
BitcoinCoreInterface now guarentees that unconfirmfun() will always b…

chris-belcher:
…e invoked, see issue #436

How this works: The list txnotify_fun now has elements which are 4-tuples instead of 3-tuples. The added tuple element is a boolean that remembers whether unconfirmfun() has been called yet. If it hasn't when the confirmed transaction is seen then call it.

Please test. One thing that can be done is editing the True/False boolean part to make it as if unconfirmfun() hadn't been called yet.


AdamISZ:
Pls rebase onto develop?

chris-belcher:
Whoops

PR #438, opened 2016-02-29
BitcoinCoreInterface now guarentees that unconfirmfun() will always b…

chris-belcher:
…e invoked, issue #436 

How this works: The list txnotify_fun now has elements which are 4-tuples instead of 3-tuples. The added tuple element is a boolean that remembers whether unconfirmfun() has been called yet. If it hasn't when the confirmed transaction is seen then call it.

Merging into develop this time. Please test. One thing that can be done is editing the True/False boolean part to make it as if unconfirmfun() hadn't been called yet.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -144,16 +144,9 @@ class BlockrInterface(BlockchainInterface):
                     blockr_url = 'https://' + self.blockr_domain
                     blockr_url += '.blockr.io/api/v1/address/txs/'
 
-                    # print 'downloading, lastusedaddr = ' + last_used_addr +
-                    #  ' unusedaddrcount= ' + str(unused_addr_count)
-
                     res = btc.make_request(blockr_url + ','.join(addrs))
                     data = json.loads(res)['data']
                     for dat in data:
-                        # todo: remove these commented out code sections
-                        # if forchange == 0: print ' nbtxs ' + str(dat[
-                        # 'nb_txs']) + ' addr=' + dat['address'] + ' unused='
-                        # + str(unused_addr_count)
                         if dat['nb_txs'] != 0:
                             last_used_addr = dat['address']
                             unused_addr_count = 0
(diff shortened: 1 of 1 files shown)


raedah:
Is this related to how 'on_tx_confirmed' calls 'on_tx_unconfirmed' in the ygs? Will it fix it so it doesnt do that anymore?

chris-belcher:
Not really @raedah thats just to reuse lines of code. It will fix forever those bugs where unconfirm_fun() is sometimes never called which leads to a crash in confirm_fun()

dreboli:
bitcoind -prune=550 -daemon -proxy=127.0.0.1:9050 -pid=/home/pi/bitcoind.pid -datadir=/home/pi/.bitcoin/ -dbcache=4 -checkblocks=1 -blocksonly=1

    git clone -b develop https://github.com/JoinMarket-Org/joinmarket.git

```
$ python wallet-tool.py wallet.json 
2016-03-01 00:37:15,786 [MainThread  ] [DEBUG]  hello joinmarket
  File "wallet-tool.py", line 89, in <module>
    load_program_config()
  File "/home/.../joinmarket/configure.py", line 220, in load_program_config
    global_singleton.config)
  File "/home/.../joinmarket/configure.py", line 241, in get_blockchain_interface_instance
    bitcoin_cli_cmd = _config.get("BLOCKCHAIN", "bitcoin_cli_cmd").split(' ')
  File "/usr/lib/python2.7/ConfigParser.py", line 618, in get
    raise NoOptionError(option, section)
ConfigParser.NoOptionError: No option 'bitcoin_cli_cmd' in section: 'BLOCKCHAIN'
```

Yesterday's git cloned master works fine.

chris-belcher:
@dreboli just so we understand, that error is unrelated to this PR ?

To fix your crash, I'd say set `blockchain_source = bitcoin-rpc` in joinmarket.cfg which is probably what you actually intended.

`blockchain_source = json-rpc` is an old obsolete option (which we should probably remove now that it's causing confusion)

dreboli:
I thought it's related and didn't really know where to report it (and I don't really know how github works either).
But blockchain_source = bitcoin-rpc solved the problem, thanks.

chris-belcher:
I tested this by commenting out the relevant lines in the unconfirmfun part. It's hard to test with the regtest setup we have and making a unconfirmed notification happen.
I suggest we merge.

Issue #444, opened 2016-03-04
Tumbler crash: URLError: <urlopen error [Errno 111] Connection refused>


dreboli:
Tumbler crash: URLError: <urlopen error [Errno 111] Connection refused>

Is this a crash, or it would have been ok to just let the script run after the error message?

Latest commit was 440357abf04b8bbecfefec74ac5dfc10d0ab3ed4

```
python tumbler.py --addrcount=2 --wait-time=120 --donateamount=1 --liquiditywait=120 --mixdepthsource=2 tumblerwallet addr1 addr2
```

The log file is missing the the error I saw on console so here's both:
Console output:
```
47,501 [Thread-5    ] [DEBUG]  that was 1 tx out of 13
47,508 [Thread-5    ] [DEBUG]  ran unconfirmfun
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 57165)
Traceback (most recent call last):
  File "/usr/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/pi/joinmarket_0.1.3_develop/joinmarket/blockchaininterface.py", line 392, in __init__
    self, request, client_address, base_server)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/home/pi/joinmarket_0.1.3_develop/joinmarket/blockchaininterface.py", line 449, in do_HEAD
    urllib2.urlopen(request)
  File "/usr/lib/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/lib/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/lib/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/urllib2.py", line 1227, in http_open
[...]

PR #445, opened 2016-03-04
Since os.system isn't catching errors anymore, we need to do that

OverlordQ:
Address issue #444


chris-belcher:
Haven't tested but seems alright.

adlai:
Thank you for catching my omission.

PR #449, opened 2016-03-05
Adjust throttle parameters and some others

chris-belcher:
Empirically found parameters that work for cyberguerilla.

Used create-unsigned-tx.py with -N 10 

Use this `print_throttle_msg` flag so the console doesn't get spammed too much.


AdamISZ:
Thanks for doing the testing.

I think it'd be better to not go as high as 450 bytes/sec, but it's debatable, plus the other stuff seems good and there will be a fairly big overall speedup, so ACK.

Hopefully we will be able to do tests with *all* parties using these settings before it goes into a release.

PR #450, opened 2016-03-05
maker_timeout_sec is readjusted for the second stage (!tx sending) ba…

AdamISZ:
…sed on transaction size, and IRC throttling variables are module vars

Tested against regtest.py and tumbler-test.py OK (latter: 17 transactions with 3-7 counterparties). Made some manual checks of correct calculation (base maker timeout sec + 1.8\*Ncounterparties\*tx_size)

Also ran a test with 11 counterparties on regtest just to sanity check the size of the extra timeout (in that case it was bumped by 120 seconds). throttling and the timeout all seemed to be OK.


chris-belcher:
I've read the code, it looks alright.

AdamISZ:
I doubt anyone else will test this separately, so merging. What's most interesting is the combination of this with the throttle optimisation anyway.

PR #455, opened 2016-03-09
taker checks for duplicate utxos now in recv_txio()

chris-belcher:
Please review

### DIFF
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -112,14 +112,19 @@ class CoinJoinTX(object):
             log.debug(('recv_txio => nick={} not in '
                        'nonrespondants {}').format(nick, self.nonrespondants))
             return
-        self.utxos[nick] = utxo_list
-        utxo_data = jm_single().bc_interface.query_utxo_set(self.utxos[nick])
+        duplicate_utxos = set(sum(self.utxos, [])).intersection(set(utxo_list))
+        if len(duplicate_utxos) > 0:
+            log.debug('ERROR duplicate UTXOs: \n' + pprint.pformat(
+                duplicate_utxos))
+            return
+        utxo_data = jm_single().bc_interface.query_utxo_set(utxo_list)
         if None in utxo_data:
             log.debug(('ERROR outputs unconfirmed or already spent. '
                        'utxo_data={}').format(pprint.pformat(utxo_data)))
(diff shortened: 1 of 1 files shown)


joinmarket/taker.py
```diff
@@ -112,14 +112,19 @@ def recv_txio(self, nick, utxo_list, cj_pub, change_addr):
             log.debug(('recv_txio => nick={} not in '
                        'nonrespondants {}').format(nick, self.nonrespondants))
             return
-        self.utxos[nick] = utxo_list
-        utxo_data = jm_single().bc_interface.query_utxo_set(self.utxos[nick])
+        duplicate_utxos = set(sum(self.utxos.values(), [])).intersection(set(utxo_list))
```
adlai: - `self.utxos.values()` is the sets of coins sent by each nick as inputs
- `sum(list_of_lists, [])` is an idiom for [flatten](https://stackoverflow.com/questions/406121/flattening-a-shallow-list-in-python#comment1198171_408281)

joinmarket/taker.py
```diff
@@ -112,14 +112,19 @@ def recv_txio(self, nick, utxo_list, cj_pub, change_addr):
...
+        if len(duplicate_utxos) > 0:
+            log.debug('ERROR duplicate UTXOs: \n' + pprint.pformat(
+                duplicate_utxos))
+            return
```
adlai: - maker providing duplicate coin effectively treated as non responding

joinmarket/taker.py
chris-belcher: effectively [] + list_of_lists[0] + list_of_lists[1] + ...

AdamISZ:
Although the taker could choose to infer that *both* copies are from a "naughty" maker and treat both as non-responding, I presume there (might?) be a dos vector that way (although a bit theoretical? auth is under encryption).

Do we consider that this needs a test that triggers the error?

chris-belcher:
Yes you're right, it could blame bother makers rather than the one that happened to respond second. That should've be too hard, I'll add it to that branch soonish.

Yes a test would be great. I'll add that too (maybe ask you guys on IRC some details about it)

chris-belcher:
The reason I've left this so long is it's not trivial, there would need to be some records kept in the code for which maker contributed what in fees and utxos, so that both makers can be evicted if they clash.

PR #456, opened 2016-03-10
"rebase" libsecp256k1

adlai:
Strictly speaking, this is a revert of a revert, although it's qualitatively similar to rebasing the original branch onto the current develop.

This is not intended to be merged as-is, but serves to keep the integration code compatible with the current develop branch.


AdamISZ:
@adlai I assume we can remove this one?

PR #461, opened 2016-03-12
bump default makercount to randint(2,4)

adlai:
- randomized default behavior!
- increases complexity of input-output correlation!
- estimated taker-side "fee event": 1.5x - 2x

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python
 from __future__ import absolute_import
 
+import random
 import sys
 import threading
 from optparse import OptionParser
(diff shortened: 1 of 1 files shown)


chris-belcher:
How is cjhunt going ? Are you able to always correlate inputs and change addresses. I had a feeling there might be too many unknowns to solve for bit you probably still could by bounding many of them to below a few thousand satoshi. Didn't actually think about it much yet.

For the PR, yeah why not.

adlai:
[cjhunt](http://github.com/adlai/cjhunt) is going slowly, as it doesn't have a well-defined goal so it's more exploratory at the moment. Correlating inputs to change works most of the time, but only within a single transaction. This change should increase the difficulty of correlating identities across multiple transactions.

AdamISZ:
@chris-belcher i believe that input-change linkage is easy in the large majority of real world cases. afaik cjhunt is just looking for jm txs in blocks, using the n equal outs, n or n-1 non-equal outs, inputs >= n heuristic. When I spent a little time looking at outputs, it looked like there was a fairly big false positive set at n=3, but I'd bet it's picking up all the real JMtxs.

PR #475, opened 2016-03-20
Add electrum blockchain interface

wozz:
Implements an electrum blockchain interface.  The initial implementation duplicates the logic for watching blockr.io, however electrum servers support watching addresses with a subscription, which should make this more efficient.


chris-belcher:
It might be a good idea to put all this into a new file called something like electruminterface.py so blockchaininterface.py doesnt get too large. Another time we could split out BlockrInterface and BitcoinCoreInterface into new files.

chris-belcher:
One thing you could also do is make electrum servers the default, here https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/configure.py#L100

AdamISZ:
Is this usable yet? I've never thought about this, but can one test electrum against testnet?

wozz:
I've been using it for the last couple weeks with a maker script.  Unfortunately there are no _maintained_ testnet servers available that I know about.  It's relatively simple to make the code changes to make a server work, but then it takes a while to get it synced with the network.

Here's the latest branch I know about for the electrum-server to support testnet: https://github.com/erasmospunk/electrum-server/tree/testnet

raedah:
This is great. I look forward to this feature.

AdamISZ:
So I ran wallet-tool against a big-ish wallet (indexes ~ 200 in each mixdepth). It took 488 seconds. Is that normal @wozz ? Or was I just unlucky.

In case it gives any info, this is what the output looked like:

```
2016-04-06 22:43:48,085 [MainThread  ] [DEBUG]  hello joinmarket
2016-04-06 22:43:48,402 [MainThread  ] [DEBUG]  sending server ping
2016-04-06 22:43:48,603 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:43:53,642 [MainThread  ] [DEBUG]  downloading wallet history from electrum server
2016-04-06 22:44:48,406 [Thread-2    ] [DEBUG]  sending server ping
2016-04-06 22:44:48,586 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:45:48,407 [Thread-3    ] [DEBUG]  sending server ping
2016-04-06 22:45:48,605 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:46:48,407 [Thread-4    ] [DEBUG]  sending server ping
2016-04-06 22:46:48,572 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:47:48,408 [Thread-5    ] [DEBUG]  sending server ping
2016-04-06 22:47:48,588 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:48:48,408 [Thread-6    ] [DEBUG]  sending server ping
2016-04-06 22:48:48,602 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:49:48,409 [Thread-7    ] [DEBUG]  sending server ping
2016-04-06 22:49:48,577 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:50:48,409 [Thread-8    ] [DEBUG]  sending server ping
2016-04-06 22:50:48,572 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:51:39,047 [MainThread  ] [DEBUG]  {"params": [406060], "id": null, "method": "blockchain.numblocks.subscribe"}
2016-04-06 22:51:48,410 [Thread-9    ] [DEBUG]  sending server ping
2016-04-06 22:51:48,597 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:52:48,410 [Thread-10   ] [DEBUG]  sending server ping
2016-04-06 22:52:48,543 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:53:48,411 [Thread-11   ] [DEBUG]  sending server ping
2016-04-06 22:53:49,109 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:54:48,411 [Thread-12   ] [DEBUG]  sending server ping
2016-04-06 22:54:48,582 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:55:48,412 [Thread-13   ] [DEBUG]  sending server ping
2016-04-06 22:55:48,593 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:56:48,412 [Thread-14   ] [DEBUG]  sending server ping
2016-04-06 22:56:48,660 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:57:48,413 [Thread-15   ] [DEBUG]  sending server ping
2016-04-06 22:57:48,659 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:58:48,414 [Thread-16   ] [DEBUG]  sending server ping
2016-04-06 22:58:48,626 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 22:59:48,414 [Thread-17   ] [DEBUG]  sending server ping
2016-04-06 22:59:48,577 [MainThread  ] [DEBUG]  {"id": 0, "result": "1.0"}
2016-04-06 23:00:40,510 [MainThread  ] [DEBUG]  electrum sync_unspent took 488.141548157sec
mixing depth 0 m/0/0/
etc
```

wozz:
That's normal from my experience (unfortunately). However I'm working on parallelizing the requests to speed up the connection in my next version of the interface. 

I also have a proxy tool I've been working on that caches stratum requests, but it needs extra logic to handle which requests need to be updated from the server and which ones can be cached longer term.

AdamISZ:
OK.

I just did a sendpayment OK (although that's not so hard). Given that you've run a maker on it for a while, I am ready to merge this, so ACK from me.

But I will wait for @chris-belcher and/or @adlai on this.

chris-belcher:
I agree this can be merged, we could even make it the default blockchain interface because it's better than blockr.io (Although I haven't tried it myself yet)

It should be rebased onto one commit.

It would be interesting to see how the real Electrum sync's so quickly.

A couple of things I'd like but we can leave for another time:
1. Rewrite add_tx_notify() to use blockchain.address.subscribe instead of polling
1. Use blockchain.address.listunspent with a list of addresses instead of single addresses one-by-one
1. On startup connect to many servers and download the longest chain of block headers, then when getting the confirmed transaction also get the merkle tree as a proof of inclusion. Then at least the Electrum server has to do a proof of work if it wants to print new bitcoins.

It's worth mentioning that running a full node is still the best way to get blockchain information for a yield generator. I know that yield generators as a group are mostly interested in profit and not in privacy, so we should be careful in case large numbers of yield generators switch to something other than a full node because it would discredit JoinMarket in the eyes of it's users. It could all end in a tragedy of the commons situation, where yield generators as a whole benefit if they all run full nodes but individual ygens find it cheaper and easier to use electrum servers or something else.
I think we'll be okay because connecting to third-party servers always has the risk that they stall or lag.

AdamISZ:
> we could even make it the default blockchain interface because it's better than blockr.io

I definitely wouldn't do that yet; it seems too prone to failing to connect, or just being very slow. Much better for it to be non-default for a while.

chris-belcher:
An extra thing I think it's worth adding is printing out the Electrum server motd (maybe with a 5 second sleep after to encourage people reading it) because many operators put donation addresses in there, since the server resources are being used it seems nice to. Also there might be other important information.

wozz:
I definitely agree with the motd idea. I'll add that when I get a chance. 

Along the lines of privacy, perhaps it would make sense to connect to different servers for each mix level. I'll add that to my wish list that I'll be working on eventually. 

I think it would make more sense to not enable this as default right away.

chris-belcher:
If you randomly connect to servers for different mixing depths, consider that if the user sync's their wallet several times, an individual server still eventually sees all their addresses linked together.

In my view complete privacy from Electrum servers will always be impossible. But it's still much better than blockr.io

chris-belcher:
I've realised Electrum servers have SSL certificates which the client uses on a trust-on-first-seen basis. That could be useful to add to this one day too.

raedah:
@wozz I would really like to see a btcd interface as well. Just sayin'

chris-belcher:
I've thought of a way to test this.

Make Wallet able to generate the wallet from a xpub key, and then have a mainnet xpub key in the test code containing a small amount of satoshi. Then this can be synchronized with an electrum server with no danger of anyone moving around the coins.

mustyoshi:
I've made a change to this patch that utilizes 2 servers per mix_depth, but at least in my case, for nearly 40k addresses, it still only goes at about 22 addresses per second, the largest mix depth is still the bottleneck, every other mixdepth finishes and there's just the one left that takes a long time. I suppose that is still a boost in performance, I'm trying to find a way to allow for multiple servers to be used on the same mix depth, but due to the sequential nature of searching the gaplimit.... Not sure what to do, any ideas?

wozz:
I was working on a proxy that could help with what you want to do: https://github.com/wozz/stratum-proxy

The idea of that was to help speed up the connections by caching data and potentially adding support for multiple backend servers like you're describing.

Keep in mind though that you'll definitely be sacrificing some privacy to do this unless you own the servers you're connecting to.

mustyoshi:
Perhaps that would work. I think I need to rewrite my code to use the multiprocessing library as opposed to the multithreading library though.

AdamISZ:
In light of recent difficulties with blockr, I've done the following:

1. rebased off of current develop
2. added includeconf variable to query_utxo_set (needed for utxo age feature)
3. added same code to the *clientserver repo and tested on both, including 1 full transaction (sorry but it's hard to do mainnet tests and takes a long time!)

I've then pushed to develop directly myself, retains your attribution @wozz so I guess that's fine. I decided not to bother squashing, see nothing wrong with these 5 or so commits, not worth the hassle of messing around.

There is stuff left incomplete, e.g. : should we update the default server list? @chris-belcher 's suggestion about batch retrieval of address data seems relevant, not done yet. Probably a bunch of optimizations are possible (subscription?).

The drawbacks are in practice (1) it's not uncommon to have the server connection fail, but the error message is sufficiently informative (just restart for a new random server). (2) It is certainly really slow compared to a local node, and probably a bit slower than blockr, but faster (in my limited tests) than bc.i which is atrociously slow.

The point is, it is not default, so I see no harm and it gives people another option especially on Tor/TAILS(?).

PR #476, opened 2016-03-21
add length check to address validation; should be 20 bytes for both p…

AdamISZ:
…2pkh and p2sh after stripping checksum and version


chris-belcher:
ACK

PR #378, opened 2015-12-24
Add py.test

WyseNynja:
http://pytest.org/latest/ is really nice for writing tests

I moved over the tests in enc_wrapper just as a starting point.

What do you think? I'm happy to move the current tests over to pytest if you like it.

```
$ PYTHONPATH=.:$PYTHONPATH py.test
================================================= test session starts ==================================================
platform darwin -- Python 2.7.10, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /Users/bwstitt/code/joinmarket, inifile: setup.cfg
plugins: cov-2.2.0
collected 4 items 

test/test_enc_wrapper.py ....
----------------------------------- coverage: platform darwin, python 2.7.10-final-0 -----------------------------------
Name                                Stmts   Miss  Cover
-------------------------------------------------------
joinmarket/__init__.py                 21      4    81%
joinmarket/blockchaininterface.py     488    422    14%
joinmarket/configure.py               113     71    37%
joinmarket/enc_wrapper.py              26      6    77%
joinmarket/irc.py                     481    427    11%
joinmarket/jsonrpc.py                  48     36    25%
joinmarket/maker.py                   254    214    16%
joinmarket/message_channel.py          68     47    31%
joinmarket/old_mnemonic.py             30     25    17%
joinmarket/slowaes.py                 357    318    11%
joinmarket/socks.py                   194    153    21%
joinmarket/support.py                 214    183    14%
joinmarket/taker.py                   359    308    14%
joinmarket/wallet.py                  238    193    19%
test/commontest.py                     50     50     0%
test/fee-estimation.py                 90     90     0%
test/randomfunc-test.py                59     59     0%
test/regtest.py                        81     81     0%
test/test_enc_wrapper.py               26      0   100%
test/tumbler-test.py                   72     72     0%
test/wallet-test.py                    84     84     0%
-------------------------------------------------------
TOTAL                                3353   2843    15%

=============================================== 4 passed in 3.20 seconds ============================================
```

PR #477, opened 2016-03-21
Test improvements

AdamISZ:
This is a combination of a few things with the goal of improving our tests:

#378 from @WyseNynja (the commit is included and a merge) to use pytest, see his explanation there.

@chris-belcher 's set of transaction serialize/deserialize tests, with a bunch more added from Bitcoin Core [here](https://github.com/bitcoin/bitcoin/blob/master/src/test/data/tx_valid.json) (on the basis that they represent various interesting edge cases, but more for validation, not for serialization; in any case it doesn't hurt). I might add more.

I have already folded into the pytest setup also the bip32 test vectors, and some valid/invalid bitcoin address tests (already threw up an error, see PR #476).

Next up would be some more tests for the other pybitcointools functions we currently use and don't currently have an upgrade path for (i.e. tests for EC and ECDSA will *not* be done until we fold in secp256k1, but testing e.g. WIF private keys will).

Probably the most important practical additional step would be to break down the regtest and tumbler-test and wallet-test scripts to run without use of subprocess, so that they can be included in the pytest setup. miniircd and regtest setup can be automated, but then call the functions in scripts like sendpayment and yieldgenerator directly, so that coverage can be measured. I hope this won't be too hard, but haven't tried yet. Once it's done, the travis PR-tester will be simplified and a lot more useful.

So this isn't really for merging until at least that last point is addressed. Apart from that, don't have to do everything at once of course.


AdamISZ:
Forgot to mention, also removed the unused pybitcointools modules as part of the cleanup.

AdamISZ:
Another note, this PR will continue to fail at the moment in the absence of fix #476 ; but anyway it's more complicated than that because the new version of the travis.yml will be using py.test and it needs the regtest to be updated for that to be included. (TLDR it's not ready)

AdamISZ:
(I will continue to add commits and squash at the end when ready)

A big step forward now:

Fully automated end to end test is working. The bitcoin(regtest) and irc(miniircd, snapshot included) will automatically start up and shut down within the test framework. Pre-configuration of bitcoin.conf and rpc creds aside, there is nothing else needed to do a test run. The command to run tests is like this:

    PYTHONPATH=.:$PYTHONPATH py.test -s --btcroot=/path/to/bitcoin/bin --btcconf=/path/to/bitcoin.conf --btcpwd=password

Takes around 1 minute at the moment, but will get longer in future.

Tests currently included are for: bitcoin addresses, libnacl encryption, transaction serialization, and sendpayment.

How I've addressed the above discussed issue "how to do tests without subprocess": sendpayment is run as a script here, and the yield generators are subprocesses (so not within the code coverage). The idea is to make several tests, in each of which only one bot is being covered by the tests, while all the others run in the background. In a way it's clunkier, but it's probably more correct anyway: isolating a test to see if we can trigger errors in a specific bot instance. I'd envisage having 3 or 4 "main tests" of standard running conditions that can be included in travis, with more, deeper, more complex tests being possible to run outside travis in a bigger test suite that takes longer. Pytest makes it easy to do that by passing command line arguments that filter which of the various test functions are to be included.

The coverage is now ~ 50% and will increase further with some trivial additions/changes, but note the last 10-20% or so of coverage for bitcoin code will of course not be addressed (EC stuff) deliberately until secp256k1.

AdamISZ:
Now integrated with coveralls.io for coverage stats, as you can see in the build check details above ( 52%, so we have a lot of work to do on that ..), squashed the commits down (but preserved the commit messages largely). It will show up in a badge in the README.

If anyone's interested, you can either browse the coverage reports at https://coveralls.io/github/JoinMarket-Org/joinmarket# or run locally to get html output with the same info (use the extra argument `--cov-report html` in the above command to run pytest).

I will make one last edit tonight to yapfify the new code, then it's ready to merge IMO. (but more tests to be done; lots more)

PR #481, opened 2016-03-25
Blockchain.info API support

lacksfish:
Fully functional blockchain.info API

**How to use**

Change the source in joinmarket.cfg to

> blockchain_source = blockchain.info

Then, tumble as usual

**Notes**

The pushtx() function in the BlockchaininfoInterface class is functional but could need a quick review. I left my findings and questions in the comments.

The interface got an extra make_bci_request() function for directing requests to the API. Includes API error handling for common errors.


joinmarket/blockchaininterface.py
```diff
@@ -115,6 +115,341 @@ def estimate_fee_per_kb(self, N):
...
+                unused_addr_count = 0
+                last_used_addr = ''
+                while (unused_addr_count < wallet.gaplimit or
+                           not is_index_ahead_of_cache(
+                                   wallet, mix_depth, forchange)):
+                    addrs = [wallet.get_new_addr(mix_depth, forchange)
+                             for _ in range(self.BCINFO_MAX_ADDR_REQ_COUNT)]
+
+                    # TODO send a pull request to pybitcointools
+                    # because this surely should be possible with a function from it
```
chris-belcher: I'd say remove these lines

joinmarket/blockchaininterface.py
```diff
@@ -115,6 +115,341 @@ def estimate_fee_per_kb(self, N):
...
+                        return
+                    bcinfo_url = 'https://blockchain.info/unspent?active='
+                    random.shuffle(self.output_addresses) # seriously weird bug with blockr.io
+                    # might not be needed for blockchain.info
+                    
+                    # create list for requests of all addresses
+                    data_list = []
+                    #for each of the output addresses, request data and add it to list
+                    for output_address in self.output_addresses:
+                        res = BlockchaininfoInterface().make_bci_request(bcinfo_url + output_address)
```
chris-belcher: This creates and discards an entire object every time this look is run. Maybe it's worth putting `make_bci_request` outside the BlockchaininfoInterface class

joinmarket/blockchaininterface.py
```diff
@@ -115,6 +115,341 @@ def estimate_fee_per_kb(self, N):
...
+        NotifyThread(txd, unconfirmfun, confirmfun).start()
+
+    def pushtx(self, txhex):
+        try:
+            json_str = btc.bci_pushtx(txhex)
+        except Exception:
+            log.debug('failed blockchain.info pushtx')
+            return None
+
+        # This part is adapted from BlockrInterface pushtx()
```
chris-belcher: I'd actually recommend to remove all this code and return (True, 'dummy txid'), the returned txid is never used for anything until that transaction confirms because of transaction malleability. The code only ever checks if self.txid has a string inside.

pushtx() cant ever guarantee that the transaction has actually been seen by every node. I've been thinking of implementing timeouts to add_tx_notify() for this but that's not related to this PR.

joinmarket/configure.py
```diff
@@ -252,6 +252,8 @@ def get_blockchain_interface_instance(_config):
         bc_interface = RegtestBitcoinCoreInterface(rpc)
     elif source == 'blockr':
         bc_interface = BlockrInterface(testnet)
+    elif source == 'blockchaininfo':
```
chris-belcher: I'm slightly concerned about blockchain.info being confused with _the_ blockchain, which the company itself seems to encourage. So maybe it's better to have this configuration be `bci` or `blockchaindotinfo`.

chris-belcher:
Great works, thanks. It's good for people who need less strong privacy that bc.i can give if blockr.io is ever down.

I commented some stuff on your code but overall it's pretty good.

raedah:
Good to see support for more backends coming in. This should help with onboarding new users.

chris-belcher:
This could be tested (after the edits are made) by import lots of used brainwallet addresses.

Or we could just ask someone to take one for the team and tell bc.i what all of their addresses are.

AdamISZ:
Ah brainwallet addresses, of course! I will make sure to include "andreasantonopoulos" :) (see UCL paper for further similar amusement)

joinmarket/blockchaininterface.py
lacksfish: I'm making it a static method as it logically belongs to the class

joinmarket/configure.py
lacksfish: Let's go with blockchain.info then. Seems to be the clearest.

joinmarket/blockchaininterface.py
lacksfish: ok

joinmarket/blockchaininterface.py
lacksfish: I'd feel more comfortable looking up the existence of the tx in the network before proceeding to make sure the push succeeded, but I am admitting the approach in pushtx() is far from perfect.

chris-belcher:
It looks alright to me. Has anyone tried it for a tumbler run ?

Also it should be rebased into one commit before merging.

AdamISZ:
I'm going to pull and try to test now. Is anyone else testing at all? @lacksfish what about you, could you give info on any tests you've done?

Note we now have a test_blockr.py; but I only put in a test for testnet there, still, in principle, it would be desirable for PR commiters to deliver tests with the code. I do appreciate that this is a particularly difficult one, but still.

AdamISZ:
First test, OK - can run wallet-test against real wallets and get correct info, although I noted a difference of 19s for bc.i vs 4s for blockr.io, don't know how repeatable that is. I will try a sendpayment.

AdamISZ:
Took a bit of burrowing, but I think I found the problem. The sync_addresses code implicitly assumes that the list of data returned is in the same order as the address list passed to the API, but it isn't for bci it seems. See here: https://github.com/JoinMarket-Org/joinmarket/pull/481/files#diff-3b408ed85f23411e81db7c0ae451edfdR188 (it's the same in the blockr code of course).

Continuing to do tests; if I got that right, it'll be fiddly but it can be fixed.

AdamISZ:
Yes, that solved it. @lacksfish you can add insert [here](https://github.com/JoinMarket-Org/joinmarket/pull/481/files#diff-3b408ed85f23411e81db7c0ae451edfdR175) this code:

    new_data = []
    for a in addrs:
        new_data.append([d for d in data if d['address']==a][0])

and then the next line obviously becomes:

    for addr in new_data:

With this edit the balance is reported correctly. Without it you will get something unpredictable because the ending value of `last_used_addr` in each loop iteration depends on the ordering of the list that bci delivers (I don't know if it's random or sorted somehow).

lacksfish:
Sorry for the late replay, am quite busy at the moment. I will look into it in the coming two weeks.

lacksfish:
AdamISZ, added your solution and also updated my repository.

AdamISZ:
Seems to have picked up a bunch of old stuff, something's gone wrong. By the way I notice a lot of it is joinmarket-qt stuff which shouldn't be here (it's supposed to be only in the gui branch).

lacksfish:
I'd be very thankful for someone testing this updated version of the API Interface, I implemented AdamISZ's fix for properly sorting the data array in sync_addresses()

AdamISZ:
@lacksfish did you see my earlier comment? This needs to be rebased somehow.

AdamISZ:
I've added this (although had to fix up the bugfix above, didn't seem to have been done properly) in fa61dbd645f008ba68342f43c4fa2b9c755ea22d ; of course it would be much better to give proper attribution but in the absence of cleaning up this PR I just went ahead.

Apart from that attribution issue, the other issue is that it's dreadfully slow (even not on Tor, but of course especially on Tor as you regularly get endpoint limit reached and waits). I see it as something to use in extremis when nothing else is available.

Closing; @lacksfish let me know if you want to fix this up somehow so you can get attribution.

Issue #200, opened 2015-08-23
Display transaction history of internal wallet


chris-belcher:
Need a script which obtains all the transactions on every address in the wallet, understands them and prints out a list akin to the History tab in Electrum.

It should understand mixing depths and joinmarkets fee so be able to say things like "coinjoin tx to internal mixing depth of 3" and "earned 50000 satoshi"

The Wallet class from common.py does not store the history, which is not needed for the operation of JoinMarket. So this script will have to call the listtransactions RPC call on its own and parse the result.

chris-belcher:
If querying the RPC, it will require txindex=1

If querying an online blockchain explorer it will reveal which addresses are yours to the server

the9ull:
I'm working on this. I'll make only the Blockr version.

> it should understand mixing depths and joinmarkets fee

It does not understand mixing depths nor cj fees now. It can list deposits/withdrawals or/and cj txs. It prints for every tx the local addresses and the moved bitcoins.

I'll make a PR.

chris-belcher:
Would you consider making it work for Bitcoin Core with txindex=1 before releasing?
Having this work only with blockr.io would probably result in a situation that coinbase.com could unmix almost everyone's coinjoins, because all the yield generators will be querying their service whenever they want to see how much money they made.

the9ull:
I understood, I try if I'll have enough time.
I could use testnet network, do you have a testnet JM wallet with an history of txs?

chris-belcher:
I've got testnet wallets you can use, PM me on reddit /u/belcher_ or bitcointalk user belcher and I'll tell you the brainwallet passphrase.

PR #484, opened 2016-03-26
Wallet history

chris-belcher:
Implementation of #200 

You need to be configured to use Bitcoin Core (-txindex=1 not required), run it with:

`python wallet-tool.py wallet.json history`

Prints a summary for every transaction. If you have numpy/scipy installed it also calculates the effective interest rate you achieved as if yield-generators were a savings account. For my own yield generator wallet the figure is 0.67% per annum

It checks if the amount of bitcoins in the wallet is the same as the resulting amount when added up as iterating through all transactions. It also checks if the number of UTXOs in the wallet match the number found by iterating through transactions. Both these checks fail for me, and I can't figure out why.

You can create a csv file for opening with spreadsheet software too:

`python wallet-tool.py --csv wallet.json history >> history.csv`

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -5,29 +5,14 @@ import getpass
 import json
 import os
 import sys
+import sqlite3
 from optparse import OptionParser
 
 from joinmarket import load_program_config, get_network, Wallet, encryptData, \
-    get_p2pk_vbyte, jm_single, mn_decode, mn_encode
+    get_p2pk_vbyte, jm_single, mn_decode, mn_encode, BitcoinCoreInterface, \
+    JsonRpcError
 
 import bitcoin as btc
-
-# data_dir = os.path.dirname(os.path.realpath(__file__))
-# sys.path.insert(0, os.path.join(data_dir, 'joinmarket'))
-
-# structure for cj market wallet
-# m/0/ root key
-# m/0/n/ nth mixing depth, where n=0 is unmixed, n=1 is coinjoined once, etc
-# pay in coins to mix at n=0 addresses
-
-# coins move up a level when they are cj'd and stay at same level if they're
-# the change from a coinjoin
-
-# using coins from different levels as inputs to the same tx is probably
-# detrimental to privacy
-
-# m/0/n/0/k kth external address, for mixing depth n
-# m/0/n/1/k kth internal address, for mixing depth n
 
 description = (
     'Does useful little tasks involving your bip32 wallet. The '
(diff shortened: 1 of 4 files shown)


AdamISZ:
Appears to be well-isolated, can't see any impact here, so utACK.

AdamISZ:
Quick record of test result as discussed on IRC:

Csv data was created OK.

>It checks if the amount of bitcoins in the wallet is the same as the resulting amount when added up as iterating through all transactions. It also checks if the number of UTXOs in the wallet match the number found by iterating through transactions. Both these checks fail for me, and I can't figure out why.

Same here, these checks failed.
Additional vague anecdotal: the number of "cj withdraw" looks too high (I see 21, expect < 15), the number of "deposit" also (see 14 but probably around 10). Sorry for vagueness, memory not good enough.

`getblockheader` was not found running Bitcoin Core 0.11.2, had to use `getblock` instead. So presumably either account for this (including the intersection with pruning) or mention somewhere that 0.12 is required (unfortunately not yet recorded [here](https://bitcoin.org/en/developer-reference#rpcs)).

chris-belcher:
Edited to fall back to `getblock` if `getblockheader` doesn't exist.

I checked my cj withdraw transactions very carefully to find some error but everything seemed to work. Although it sounds like waxwing's problem is too many withdraws rather than incorrectly the calculated balance.

chris-belcher:
grubles from IRC tried this (and got 3.2% as an effective interest rate) but didn't get any of the errors. He only deposited coins into the wallet and didnt withdraw, which suggests again it's a problem with `cj withdrawal` somehow.

chris-belcher:
An idea for finding the bug, for me the UTXO count from history is higher than the wallet UTXO count. So history believes there are extra UTXOs which dont actually exist. So if I compare the two UTXO sets and find their difference it should give me some clues to what went wrong.

chris-belcher:
Used the above idea to find and fix the bug.
Turns out some transactions are in walletname="", in my case I had exported a privatekey and spent the UTXO entirely to some other address. This didnt appear with the label "joinmarket-wallet-whatever".

Seems to work now. Please, everybody try this.

AdamISZ:
Works correctly, feel free to merge.

wallet-tool.py
```diff
@@ -269,3 +282,200 @@ def cus_print(s):
...
+        def f(r, deposits, deposit_times, now, final_balance):
+            return np.sum(np.exp((now - deposit_times) / 60.0 / 60 / 24 /
+                365)**r * deposits) - final_balance
+        r = brentq(f, a=0.1, b=-0.1, args=(deposits, deposit_times, now,
+            balance))
+        print('continuously compounded equivalent annual interest rate = ' +
+            str(r * 100) + ' %')
+        print('(as if yield generator was a bank account)')
+    except ImportError:
+        print('numpy/scipy not installed to unable to calculate effective ' +
```
chris-belcher: spelling mistake here

chris-belcher:
note to self: must add history to the --help documentation

chris-belcher:
Done everything, ready to merge except for two extra commits that have appeared in this PR. Somehow their hash must've changed because of me doing git rebase

I have half a mind to merge anyway.

PR #486, opened 2016-03-27
New re-integration of secp256k1 via ludbb binding; no longer using

AdamISZ:
a local version but assuming the binding is installed already with
"pip install secp256k1". Does not yet include an updated form of the
donation code for custom nonces.
add mnemonic tests
Modification so that either secp256k1 or pybtc can be used;
secp256k1 python binding is detected automatically.

NOT YET READY TO MERGE.

Passes test suite currently both with and without secp256k1 module installed on the system.

Tests ongoing. Todo: re-add alternative code for donation tx. Test tumbler and modify if needed (I expect not). Apologies for re-formatting of create-unsigned-tx; will also need some testing there.

Will also need to add warning/info message if secp256k1 is not detected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,19 +31,26 @@ Widespread use of JoinMarket could improve bitcoin's fungibility as a commodity.
     make check
     sudo make install
     ```
++ (optional, recommended): Install the libsecp256k1 Python library:
 
+    ```
+    pip install secp256k1
+    ```
+
+ Note that this requires `pip`. This library will make use of libsecp256k1 if you already have it installed on your system. Most likely you won't, and it will try to build libsecp256k1 automatically for you. This requires some development packages; please read the installation details [here](https://github.com/ludbb/secp256k1-py#installation), provided for Debian/Ubuntu and OS X. Please also note that if you can't complete this step, Joinmarket will still run correctly, with two disadvantages: wallet loading is much slower without libsecp256k1, and the ECC crypto code is far less robust and well tested.
 + Matplotlib for displaying the graphs in orderbook-watcher (optional)
 
 ###DEBIAN / UBUNTU QUICK INSTALL FOR USERS:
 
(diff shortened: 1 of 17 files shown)


chris-belcher:
I think this should be added after new release that pushes out all the fixes we've done in the last few months.

AdamISZ:
@chris-belcher sounds reasonable; I've been thinking we should make a new release, too.

AdamISZ:
Tumbler testing is passing with and without; with the new testing framework it'll be possible to add a variety of different setups, but in that case tests will take hours, so obviously won't be included in build tests. I'll probably do something along those lines later. Next up is the custom nonce issue. After that I'll do some tests of old and new bots together and then some live testing I guess.

AdamISZ:
I've run some tests of this version of the donate code on both versions, it seems to operate fine - I checked that the receiving party can reconstruct the receiving key OK - not that there was much chance of it going wrong as it's essentially entirely unchanged.

The trick/hack is to simply import the necessary ECC functions *directly* from bitcoin.main in those 2 functions (see the diff), so that whether we are running with secp256k1 or not, it uses the old version of the code.

In related news, the owner of the secp256k1 repo is happy to pull the custom_nonce PR as long as I write a test for it. That update can happen whenever it's convenient.

The remaining work is to do more tests, in particular compatibility test of secp256k1 bots and non-secp256k1 bots together. I don't expect any trip ups at all here, as I dealt with that mess last November.

Final point, builds are passing, but coverage is not, for a specific reason: the addition of 3 new files (secp256k1_[main, deterministic, transaction].py) reduces the coverage, which counts as a failure to coveralls. Since builds already take 3-5 minutes in total including installation, I am not enamoured with the idea of adding a complete re-run of tests for both versions of the code. Suggestions on that point are welcome.

AdamISZ:
Had forgotten to update the code for importing privkeys into wallet; this is now corrected, and so is the "meta-problem" that importprivkey was not tested.

PR #489, opened 2016-04-01
Add absolute fees to yield-generator-basic and command line configuration

veqtrus:
Continuation of #434

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, print_function
 import datetime
 import os
 import time
+from optparse import OptionParser
 
 from joinmarket import Maker, IRCMessageChannel
 from joinmarket import BlockrInterface
@@ -17,18 +18,17 @@ from joinmarket import Wallet
 
 # import blockchaininterface
 
-txfee = 1000
-cjfee = '0.002'  # 0.2% fee
-jm_single().nickname = random_nick()
+txfee = 5000
+cjfee_a = 200
+cjfee_r = '0.002'
+ordertype = 'relorder'
+jm_single().nickname = ''
 nickserv_password = ''
-
-# minimum size is such that you always net profit at least 20% of the miner fee
-minsize = int(1.2 * txfee / float(cjfee))
-
+minsize = 100000
 mix_levels = 5
 
-log = get_log()
 
+log = get_log()
 
 # is a maker for the purposes of generating a yield from held
 # bitcoins without ruining privacy for the taker, the taker could easily check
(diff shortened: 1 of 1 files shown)


AdamISZ:
Haven't had a chance to review, but certainly agree with the basic idea of optionize-ing configuration of yieldgen. However, you'll need to add any non-default options to the launch of the yield generator subprocess to make the build test pass (see test/test_regtest.py). At least I presume that's the issue, will probably have time to look more carefully tomorrow I guess.

yield-generator-basic.py
```diff
@@ -156,9 +161,51 @@ def on_tx_confirmed(self, cjorder, confirmations, txid):
 
 
 def main():
-    load_program_config()
+    global txfee, cjfee_a, cjfee_r, ordertype, nickserv_password, minsize, mix_levels
```
raedah: Why is this global line needed?

veqtrus: Either that or `self` as a parameter. I just wanted to explicitly specify which global variables are modified.

veqtrus:
There are no non-default options.

AdamISZ:
Unfortunately the test build times out so we don't get the logs, but a quick look suggests to me that the problem is the second sendpayment test with parameters [here](https://github.com/JoinMarket-Org/joinmarket/blob/develop/test/test_regtest.py#L33). The amount is < 3000000 satoshis and if I calculate correctly, that's smaller than the minsize you get with default fees. Can you confirm @veqtrus by running a test as described [here](https://github.com/JoinMarket-Org/joinmarket/wiki/Testing) ?

If not, I can do it later. If that is the cause, probably we can just bump up that very small coin size in the test.

AdamISZ:
@veqtrus ok i just pulled this to test it, and noticed it's based off an old version, the last commit you took was from Feb 4th. You'll need to bring it up to date with the develop branch.

The comment I made above is probably also true (edit [this line](https://github.com/JoinMarket-Org/joinmarket/blob/develop/test/test_regtest.py#L33) and make the amount > 3000000)). I don't think the test will pass otherwise.

I think it'd be great to merge this as an improvement to the top level scripts (first step, I guess).

AdamISZ:
OK, it seems the coverage decrease is the same reason as for the newsecp256k1 (explanation in that PR). That's a technical issue. Edits to the yigen script itself aren't included in the coverage.

But it passes travis, which is more important.

I looked through the code and also I did some tests on regtest by editing the yield generator command to look like this:

    ygp = local_command([python_cmd, yg_cmd, '-t', '1000', '-c', '0.0011',
                             '-m', '6', '-o', 'relorder',
                             str(wallets[i]['seed'])], bg=True)

Obviously these are just some random option values, but it passed OK.
Ideally the tests should be updated like that, either you or I can make such a commit at some point.

So merging.

PR #490, opened 2016-04-04
More tests

AdamISZ:
Not pressing time-wise, but this contains a lot of extra tests.

There *are* also some code modifications, in particular to the bitcoin/ subdirectory - they are almost all of the form of removing unused code/modules. Exception to that rule: the handling of non-standard sighash flags was wrong and has been fixed (and included in tests).

This brings the coverage up to around 77%. The tumbler test is now included in the tests, but excluded from the build because it takes so long.

I can continue to create more tests, but some areas are quite difficult; code paths in blockchaininterface, maker and taker, and a few in irc in particular. It is of course a big time sink, so I'm not sure how quickly/how far I'll go further with it. Others are welcome to pitch in.


AdamISZ:
Sorry, i've got screwed up again with the commits. there was only supposed to be 1. I'll look at it later.

PR #491, opened 2016-04-05
Add several new tests, remove unused code in bitcoin module

AdamISZ:
add several new tests, remove unused code in bitcoin module
add tests of ecc, donations, irc, and several improvements
covering more error cases
add test of spend to p2sh
fix up sighash codes in transaction code, more tests
test spend-from-p2sh
Add tumbler test to repo (but not build test)
Fix make_wallets to only reset index for syncing after coins allocated,
since otherwise all coins in one mixdepth go to the same address.
Minor cleanup.
add dependencies for secp256k1 to build
fix/remove ripemd dependencies
minor fix for nonsecp256k1 tests

====
Fixed version of #490 ; all commits squashed into 1 (I always get problems when merging during a PR!)

So I will close #490 and add some TLDR here shortly. Again, it isn't urgent.


AdamISZ:
Just some general notes while I remember:

* bci.py contained a load of old crap like "helloblock" (doesn't exist anymore), eligius push tx (which still exists but does nothing) etc. removed all except what we use.
* ripemd.py is not being used. i have no intention of vetting it as an alternative ripemd implementation, so dumped it.
* Non-standard SIGHASH flags were not implemented correctly (specifically, SINGLE and NONE were missing the sequence number edit, plus there was a minor syntax error, so they didn't work at all). Fixed them.
* Some other code that we don't use was removed, for example: helper functions for command line use of pybtc.
* Amongst the tests, I added some minimal ones for spending to and from p2sh. This is very important with segwit coming up (and for other reasons of course)
* A trick point is testing legacy vs testing new secp256k1 code. I've made the new travis build pick up the secp256k1 version, and removed the legacy versions of the *main.py, *deterministic.py and *transaction.py from the coverage. Really the tests should cover both, but running a build against both means something like 8 minutes, too long.
* test_tumbler is added to the commit, but not included in travis build. It currently only has one set of parameters (includes donations), but you can easily add other sets of parameters to do multiple runs. One run currently takes around 5-10 minutes.
* The testing Wiki page has been updated. I hope that the much easier/better method using pytest encourages some people to do tests. Or to write some more.
* Coverage (with above qualifications) now around 76%, which is a big step forward, but we need extra paths tested irc, maker, taker and blockchaininterface. A lot of it will be quite hard to test.

PR #494, opened 2016-04-09
Transaction un/confirm timeouts

chris-belcher:
Implementing #118 

Also modified tumbler.py to take advantage of this. I've tested this a bit but it more testing is useful so I can recommend users to check out this branch and try it.

In joinmarket.cfg also now has a field [TIMEOUT] which has

> maker_timeout_sec = 30
> unconfirm_timeout_sec = 90
> confirm_timeout_hours = 6

#475 and #481 should be merged after this, they both need slight modifications to add in the timeout callback.


AdamISZ:
I can't figure out why this didn't trigger a build test. Maybe because it's a PR from a branch in the same repo. That's the only thing that's different here I think.

chris-belcher:
Edited test_tumbler.py to test the add_tx_notify() timeouts 

Now I'll work on creating a new test that tests these timeouts without running the entire tumbler script

AdamISZ:
Is it not possible to just subclass RegtestBlockchainInterface within the test? It seems much better not to make alterations to the core codebase, don't you think? I didn't think it through in detail, maybe it's not possible for some reason.

chris-belcher:
Yes it's possible, I just thought it would be cleaner this way. Because the default behaviour of RegtestBlockchainInterface is unchanged and that class is almost entirely used only for tests anyway. I didn't think altering the core codebase would be a problem because the new feature and any bug fixes do that too.

AdamISZ:
Ah sorry, didn't notice that was added to regtest* pushtx, not the main pushtx, yes, sure it's only for tests by definition, so that's fine.

Issue #499, opened 2016-04-16
Add signalling so tests know when yield-generator has started


chris-belcher:
One issue on some machines is the test suite has a hardcoded wait time after creating a yield-generator.
It would be better to have the yield generator signal to the test suite that it's ready.

Here is one way to do this:
When yield-generator is ready it does the equivalent of `touch logs/yigen-statement.csv`.
The test suite simply loops and waits for the last modified time to be updated.

chris-belcher:
A problem with this is it only works for one yield generator. Maybe use another file instead of yigen-statement.csv

PR #500, opened 2016-04-17
Added option to pass rpc username in tests

chris-belcher:
I needed this to make tests work for me

Also added a commented-out `time.sleep(90)` in test_irc_messaging.py which seems to be required for me but not for anyone else. In time we could move to something like #499 

I've tested it and ran it so there shouldn't be any obvious crashes.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -8,6 +8,7 @@ from joinmarket import load_program_config
 bitcoin_path = None
 bitcoin_conf = None
 bitcoin_rpcpassword = None
+bitcoin_rpcusername = None
 miniircd_proc = None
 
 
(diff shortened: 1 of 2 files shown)


AdamISZ:
I'm still really confused as to why I don't need this? I recently upgraded from 0.11.2 to 0.12.1 and it still works fine. What's the difference do you think? Given that it's a default, no problem, I'm just confused.

Re: the timeout, that's a really substantial point I think, it seems to have just been luck that both my machine and Travis were both happy with the previously set timeouts. A 90 second timeout would be a big extra delay in the build tests that I was trying to avoid (I think it's currently around 4-5 minutes on Travis). There should be some kind of "I'm ready" signalling shoved in there, no doubt.

AdamISZ:
New theory: it's due to the fact that the rpcuser is not set in the joinmarket.cfg the same as in the bitcoin.conf (I notice the default in configure.py is "bitcoin" not "bitcoinrpc")

PR #505, opened 2016-04-27
Refactor message channel

AdamISZ:
This is not intended to be merged right now, although technically the only thing blocking it is testing and review, making a PR now seemed a good way to facilitate that.

Idea: before attempting to plug in any alternative message channel implementation other that IRC, it's logical to do something we talked about since the early days: properly separate all joinmarket messaging protocol code from the messaging implementation, which is currently only IRC.

This PR sets MessageChannel as an abstract base class, with the following functions that must be implemented by subclasses:

```
@abc.abstractmethod
    def run(self):
        """Main running loop of the message channel"""

    @abc.abstractmethod
    def shutdown(self):
        """Stop the main loop of the message channel,
        shutting down subsidiary resources gracefully.
        Note that unexpected disconnections MUST be
        handled by the implementation itself (restarting
        as appropriate)."""

    @abc.abstractmethod
    def _pubmsg(self, msg):
        """Send a message onto the shared, public
        channel (the joinmarket pit)."""

    @abc.abstractmethod
    def _privmsg(self, nick, cmd, message):
        """Send a message to a specific counterparty"""

    @abc.abstractmethod
    def _announce_orders(self, orderlist, nick):
        """Send orders defined in list orderlist either
        to the shared public channel (pit), if nick=None,
        or to an individual counterparty nick. Note that
        calling code will access this via self.announce_orders."""
```

The functions are prepended with '\_' not '\_\_' to allow polymorphism to work. The "top layer functions" are thus privmsg, pubmsg etc. and are implemented in MessageChannel. These functions pass the responsibility for message sending to the corresponding underscore prepended function, but themselves handle joinmarket's own messaging protocol (a reminder there is a somewhat incomplete documentation [here](https://github.com/JoinMarket-Org/JoinMarket-Docs/blob/master/Joinmarket-messaging-protocol.md)): encrypting for certain commands, controlling acceptable command names, calling callbacks, keeping track of encryption boxes etc. What remains in irc.py includes for example chunking, since that's implementation specific.

I'm going to try to implement joinmarket over matrix now by writing a `matrixmc.py` implementing this, which hopefully should be a good test of how sane it is.

One last point, I've deliberately made several isolated commits (helped with testing, but might help code review so you can see the stages of the change).


chris-belcher:
This refactor means MessageChannel is assumed to be in a plaintext line-separated format. It looks like it cant be used for a binary protocol format.

What do you think of creating a new subclass called something like LineSeparatedMessageChannel which contains all this, irc and matrix can both inherit from that to reuse code, and any future binary protocol (some p2p fancyness) could still inherit from the old MessageChannel.

AdamISZ:
Yes, I think that's quite right.

The current "joinmarket messaging protocol doc" is assuming a specific plaintext format, e.g. space separated fields.

We can just define the messages in object form like `msg1 = (cmd='auth',fields=[<pubkey>, <sig>])`, and then instead of calling `self._privmsg(nick, cmd='auth', msg= pubkey + '  ' + sig)` we can call `self._privmsg(nick, msg1)`. Then the serialization of the privmsg can be deferred to the implementation class.

Although not disagreeing with the idea of creating another abstraction layer, either. We could probably do even more along those lines, like creating a "JoinMarketMessage" class if we feel like it.

AdamISZ:
I think this is an important step, even without further stratification. That can be done in a second step and should be significantly easier (no longer needing to untangle e.g. encryption code). @chris-belcher if you agree, I will squash these commits and merge. A conflict has already been picked up which will take work and extra testing.

chris-belcher:
I'd prefer to avoid the situation where one day we want to use a binary messaging channel and then we have to refactor it again to separate out the parts which involve plain text.
But if it's easier to have it like this, then lets say go for it.

AdamISZ:
@chris-belcher yes, i understand. what this step does is remove the joinmarket code out of IRC. So the next step of refactoring/stratifying will only happen at the messagechannel level (up to perhaps some trivial changes). so i guess i'm just saying that doing it in two steps will not involve duplicating that work.

I'm keen to get this in now-ish so that it gets some live testing, and so that it isn't lost to divergence so to speak. I'll do that soon and merge it once it's been re-tested.

AdamISZ:
The conflict is with: https://github.com/JoinMarket-Org/joinmarket/commit/84f8f3873a4f8f322ffd267bad6ce89e85b03b6d

This change will now apply to message_channel.py, so it seems I can't merge. Is there anything I can do apart from ask him to re-create his PR? (I could just add the change myself to message_channel but then he loses authorship). Ping @AlexCato

AdamISZ:
In the absence of a response from @AlexCato and also after merge, squashing is complicated, I am going to merge as is. This is *potentially* a bit more urgent since we are now having IRC connection issues again. Messaging centralization is again proving the weakpoint.

In other news, Travis builds seem definitely not to be triggering any more. In any case, I ran the test suite locally, it is passing.

PR #514, opened 2016-05-05
Update README.md

AlexCato:
Update installation instructions from 0.1.3 to 0.1.4.
Since 0.1.4 has been released already, I do believe this should be in master right away, so visitors who heard about the new release do not see 0.1.3 still on the first page of joinmarket


chris-belcher:
It needs to be in both the develop and master branch. I will merge to master soon because of the other bug.

Issue #518, opened 2016-05-07
irc.py: index error parsing chunks


AlexCato:
Running the latest 0.1.4 release of joinmarket, I just noticed this in the DEBUG mesages (while running a yield generator).
The yield generator does not crash, I do not know whether that error has an effect.

```
2016-05-07 11:39:41,693 [MainThread  ] [DEBUG]  <<pubmsg nick=no-one1 message=!relorder 1 476190 62263382 0 4.2e-06!relorder 2 62263382 124400526 0 9.5e-06!relorder 3 124400526 186537670 0 1.48e-05!relorder 4 186537670 310811959 0 2.01e-05!relorder 5 310811959 497223391 0 3.08e-05!relorder 6
2016-05-07 11:39:41,693 [MainThread  ] [ERROR]  list index out of range
Traceback (most recent call last):
  File "/home/user/joinmarket-0.1.4/joinmarket/irc.py", line 312, in check_for_orders
    minsize = _chunks[2]
IndexError: list index out of range
2016-05-07 11:39:41,693 [MainThread  ] [DEBUG]  index error parsing chunks
2016-05-07 11:39:52,140 [MainThread  ] [DEBUG]  <<pubmsg nick=no-one1 message=!absorder 0 100000 11297270 1250 1750 ~
2016-05-07 11:40:01,426 [MainThread  ] [DEBUG]  <<pubmsg nick=no-one1 message=!absorder 3 40000000 66717385 0 4631!absorder 2 10000000 39999999 0 3518!absorder 1 2730 9999999 0 2600 ~
2016-05-07 11:40:58,115 [MainThread  ] [DEBUG]  <<pubmsg nick=Sibovihon_ message=!absorder 0 100000 11297270 1250 1750 ~
2016-05-07 11:41:01,901 [MainThread  ] [DEBUG]  <<pubmsg nick=no-one1 message=!relorder 1 476190 62263382 0 4.2e-06!relorder 2 62263382 124400526 0 9.5e-06!relorder 3 124400526 186537670 0 1.48e-05!relorder 4 186537670 310811959 0 2.01e-05!relorder 5 310811959 497223391 0 3.08e-05!relorder 6
2016-05-07 11:41:01,901 [MainThread  ] [ERROR]  list index out of range
Traceback (most recent call last):
  File "/home/user/joinmarket-0.1.4/joinmarket/irc.py", line 312, in check_for_orders
    minsize = _chunks[2]
IndexError: list index out of range
[...]

PR #519, opened 2016-05-07
Update irc.py: improve error message fun check_for_orders

AlexCato:
Changed exception to warning and logging more verbose message to debug log. No changes in program logic.
Inspired by [this question](https://github.com/JoinMarket-Org/joinmarket/issues/518)


chris-belcher:
looks good to me

eventually we should move to using all the features of the logging class, not just sending everything to debug()
then people could switch off debug and only read the important messages

AdamISZ:
ACK; merge?

PR #521, opened 2016-05-08
Catch exceptions in encryption initialization

AdamISZ:
Inexplicably I did not do any sanity checking on whether the pubkey/key objects passed into the wrapper functions in enc_wrapper when I set it up.

In this PR there is a placeholder `NaclError` exception type and the pubkeys and keypair objects are checked to be of the right format.

The specific relevance is the `fill` and `pubkey` messages pass a pubkey in hex, now it is checked to be of the right format and if not, the message is ignored. See the edits to the `CoinJoinTx` and `CoinJoinOrder` objects in taker and maker.

There are some tests added also in test_enc_wrapper and in test_irc_messaging (bad `fill`), although not for the exception catch in `CoinJoinTx.start_encryption`. That test is a little tricky to write. Will see if coverage complains or not. And since the crucial point here is to catch on the maker side, I thought it better to PR as is.

### DIFF
--- a/joinmarket/maker.py
+++ b/joinmarket/maker.py
@@ -9,7 +9,8 @@ import threading
 import bitcoin as btc
 from joinmarket import IRCMessageChannel
 from joinmarket.configure import get_p2pk_vbyte, load_program_config, jm_single
-from joinmarket.enc_wrapper import init_keypair, as_init_encryption, init_pubkey
+from joinmarket.enc_wrapper import init_keypair, as_init_encryption, init_pubkey, \
+     NaclError
 
 from joinmarket.support import get_log, calc_cj_fee, debug_dump_object
 from joinmarket.taker import CoinJoinerPeer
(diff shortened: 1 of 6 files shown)


joinmarket/maker.py
```diff
@@ -32,9 +33,14 @@ def __init__(self, maker, nick, oid, amount, taker_pk):
...
         self.kp = init_keypair()
-        # the encryption channel crypto box for this Order object
-        self.crypto_box = as_init_encryption(self.kp,
+        # the encryption channel crypto box for this Order object.
+        # Invalid pubkeys must be handled by giving up gracefully (otherwise DOS)
+        try:
+            self.crypto_box = as_init_encryption(self.kp,
                                              init_pubkey(taker_pk))
+        except NaclError as e:
+            log.debug("Unable to setup crypto box with counterparty: " + repr(e))
```
chris-belcher: this should send an error back to the taker, msgchan.send_error()

joinmarket/taker.py
```diff
@@ -85,8 +86,12 @@ def start_encryption(self, nick, maker_pk):
         if nick not in self.active_orders.keys():
             log.debug("Counterparty not part of this transaction. Ignoring")
             return
-        self.crypto_boxes[nick] = [maker_pk, as_init_encryption(
+        try:
+            self.crypto_boxes[nick] = [maker_pk, as_init_encryption(
                 self.kp, init_pubkey(maker_pk))]
+        except NaclError as e:
+            log.debug("Unable to setup crypto box with " + nick + ": " + repr(e))
```
chris-belcher: also should send an error back to the maker, msgchan.send_error()

AdamISZ:
send_error lines added, thanks.

PR #531, opened 2016-05-22
added config absurd_fee_per_kb default 150k sats, and test case

AdamISZ:
Exceeding the absurd value raises a ValueError so is a crash condition. On balance this seems the right choice.

The value is set in the config file, default given, user prompted if it's not present.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -749,6 +749,13 @@ class RegtestBitcoinCoreInterface(BitcoinCoreInterface):
         super(RegtestBitcoinCoreInterface, self).__init__(jsonRpc, 'regtest')
         self.pushtx_failure_prob = 0
         self.tick_forward_chain_interval = 2
+	self.absurd_fees = False
+
+    def estimate_fee_per_kb(self, N):
+	if not self.absurd_fees:
+	    return super(RegtestBitcoinCoreInterface, self).estimate_fee_per_kb(N)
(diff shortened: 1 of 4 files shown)


chris-belcher:
Looks great!

I wonder if we'll one day regret crashing the scripts one day when fees genuinely go up. What if we just made it a ceiling that fees never go above?
Then again it would still mean they pay 150 satoshi per byte. (Although today's rate is ~50 sat/b so not too far)

AdamISZ:
> I wonder if we'll one day regret crashing the scripts one day when fees genuinely go up. What if we just made it a ceiling that fees never go above?

It seems like two points there: first, crashing the scripts just means the user needs to figure out if something is wrong, and if necessary bump the config val. Second, I did consider whether it was valuable to create an additional, non-relative absolute value limit. I can certainly see that; Core has it anyway (for those using Core). But I wanted to do the simplest thing that addresses the problem, this seems to be OK for that.

chris-belcher:
Yes, I thought an alternative could be to clamp the fee at a fee ceiling. But that would mean it silently spends 150 satoshi per byte on a transaction, which is a lot.
So crashing is better, but it's something to remember for the future if fees go up.

AdamISZ:
A bit of a mystery - second time seen this - the travis build was not triggered by the PR. I have no idea what causes it. I am going to merge and run my yg on it. I have run the test suite locally of course.

PR #535, opened 2016-05-24
Fix NotifyThread

wozz:
Extra line, looks like a merge error

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -343,8 +343,7 @@ class BlockrInterface(BlockchainInterface):
                 self.confirmfun(
                     btc.deserialize(confirmed_txhex), confirmed_txid, 1)
 
-        NotifyThread(self.blockr_domain(), txd, unconfirmfun, confirmfun).start()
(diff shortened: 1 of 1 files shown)


AdamISZ:
I suspect this is connected to whatever caused the merge of joinmarket-qt code into master (you might remember @chris-belcher we discussed it on IRC).

I think problems already started before the last release because master had diverged from develop. I suspect the right solution is to force-push from develop and make a new release.

AdamISZ:
Seems the stale version of the call would have the thread die in the background due to the wrong syntax (blockr_domain) so probably not causing users problems right now.

wozz:
What happens when it's like this is it completely crashes on the first call because the number of arguments is wrong. It doesn't make it to the second call.

Removing the first line then causes the thread to die like you described.  So the fix is to add the last argument to the first call.

AdamISZ:
Oh I see, my mistake; I thought this was the release, it's only the most recent commit which is after that.

chris-belcher:
Very weird that this happened. I think it might be because I didnt always git pull before I git merge'd.

PR #541, opened 2016-05-27
Matrix message channel basic implementation

AdamISZ:
This PR is currently for TESTING, not yet merge-able.

If you want to test this PR with me, do something like:

    git fetch origin refs/pull/541/head:pull_541

where `origin` should be replaced with whatever your name is for this repo (origin is the default). A branch is automatically created with name pull_541 in your local repo. Make sure to also `git checkout pull_541` so as to run this code.

You also need to `pip install matrix-client`.

Logic: If the user sets the variables:

```
matrix_host = joinmarket.me
matrix_port = 8448
matrix_chan = joinmarket-pit
```

in the [MESSAGING] section of the config, then the code will connect over the matrix homeserver at joinmarket.me:8448, room joinmarket-pit. If `matrix_host` is not found, you default back to IRC settings.

If anyone with a testnet wallet could do this (ping me), we could carry out some tests live. I have already tested with regtest of course (the test suite). It is somewhat slower due to setup time, but actual transactions process faster due to no throttling.

The main TODO is to duplex this, allowing to read and write orders and updates to both IRC and matrix.

The secondary TODO is more fuzzy but maybe important: in order to do privmsgs a temporary private room has to be setup, with the default code from the sdk this involves some delay (currently hardcoded at 2 seconds, yuck - timing tests seem to indicate that room creation has a delay of ~1s); would like to make this quicker/cleaner. Some investigation around the end of April ended up with me being somewhat dispirited about how complex the API is; in particular, it seems to be built with a lot of statefulness in mind; by default the client is downloading everything and anything. Anyway, this needs a lot of work to improve.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "miniircd"]
 	path = miniircd
 	url = https://github.com/JoinMarket-Org/miniircd
+[submodule "matrix-python-sdk"]
+	path = matrix-python-sdk
+	url = https://github.com/matrix-org/matrix-python-sdk
--- a/joinmarket/__init__.py
+++ b/joinmarket/__init__.py
@@ -9,6 +9,7 @@ from .support import get_log, calc_cj_fee, debug_dump_object, \
 from .enc_wrapper import as_init_encryption, decode_decrypt, \
     encrypt_encode, init_keypair, init_pubkey, get_pubkey, NaclError
 from .irc import IRCMessageChannel, random_nick, B_PER_SEC
+from .matrixmc import MatrixMessageChannel
 from .jsonrpc import JsonRpcError, JsonRpcConnectionError, JsonRpc
 from .maker import Maker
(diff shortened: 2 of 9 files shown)


chris-belcher:
Looks good. I read over the code a little just now.

AdamISZ:
A couple of test transactions on testnet e.g. d57f394e276a8def4cd14f44cf38c71108cc04c990a4dac5aabeae5fd00a726a

Issue discovered: on starting up (entering), it seems that there are big delays, this is very much related to the final paragraph in the above (although it seems unclear: seems like much *longer* room-creation delays on just joining the server), and I don't know how to solve it at the moment.

AdamISZ:
I tried to log in via a different homeserver (matrix.org) but most homeservers are configured to require a configured login; I think that'd mean at the very least a password and email for every new bot. I have mine configured to register with a new password only, so it can be non-interactive. Yuck.

Issue #540, opened 2016-05-26
sanity_check in yield-generator-oscillator.py rejects long integers, breaks with 22 bitcoin


jamesphillipturpin:
So if offer['maxsize'] is 2147483647 or less it is type int but it is instead 2147483648 or more then it is type long. The sanity_check rejects anything other than type int. These are satoshis so it breaks with 22 bitcoin. So I'm not sure if I should modify the sanity check by changing this line:
`assert isinstance(offer['maxsize'], int)`
to:
`assert (isinstance(offer['maxsize'], int) or isinstance(offer['maxsize'], long))`

Or modify offers just before doing sanity check:
`for offer in offers:  if isinstance(offer['maxsize'],long): offer['maxsize'] =  2147483647`

I suppose I could even just comment out the sanity check. I'm just wondering if any of these options are expected to create problems? I've already run larger transactions through one of the other yield generators in a previous version, so although I could imagine type long numbers creating problems, that doesn't seem to be the case. I thought I should check in with the community before modifying code on the fly though.

raedah:
It shouldn't be a problem for you to modify sanity_check for your needs. That code is straight forward and wont cause any other issues. Question though as to why this bug is still being caused for you. Other use it with amounts over 22 bitcoins and have had no issue, so this may be something specific to your platform architecture.

jamesphillipturpin:
For my bot I've changed line 184 of joinmarket/yield-generator-oscillator.py to:
`        assert (isinstance(offer['maxsize'], int) or isinstance(offer['maxsize'], long))`

I don't know why my local installation of Python uses 32 bit for int when I have a 64 bit processor, other than that the documentation indicates that as the minimum spec.

More generally, this test code indicates that Python refuses to cast integers larger than a 32 bit signed int as int rather than long:
`>>> type(int(pow(2,31)-1))
<type 'int'>
>>> type(int(pow(2,31)))
<type 'long'>`
I've reinstalled Python 2.7.11 from a couple different sources, with no change to this behavior.

jamesphillipturpin:
After making more changes to my personal offers I also had to change line 183 similarly.
So in total I changed lines 183 to 184 of joinmarket/yield-generator-oscillator.py to:

```
assert (isinstance(offer['minsize'], int) or isinstance(offer['minsize'], long))
assert (isinstance(offer['maxsize'], int) or isinstance(offer['maxsize'], long))
```

raedah:
[...]

PR #545, opened 2016-05-29
Change sanity_check in yield-generator-oscillator.py to accept long integers for maxoffer and minoffer.

jamesphillipturpin:
Python 2.7 does not necessarily support 64 bit integers for type int. It's not uncommon that Python 2.7 for Windows will only allow 32 bit signed integers for type int, and that larger integers will be supported as type long.  As a result values as low as pow(2,31) = 214783648 santoshis = 21.47483648 bitcoin will be stored as a long in that case, even if one attempts to cast them as int. Changing the sanity test to allow type long seems to be the minimum change to fix this problem without creating conflicts for other users.

### DIFF
--- a/yield-generator-oscillator.py
+++ b/yield-generator-oscillator.py
@@ -180,8 +180,8 @@ def sanity_check(offers):
         assert offer['txfee'] >= 0
         if offer_high:
             assert offer['maxsize'] <= offer_high
-        assert isinstance(offer['minsize'], int)
-        assert isinstance(offer['maxsize'], int)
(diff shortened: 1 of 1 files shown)


raedah:
This is straight forward and I see no issue with it. Okay by me. Closes issue #540.

raedah:
I suggest changing the pull request title to something more descriptive of the issue.

chris-belcher:
You need to merge it into joinmarket:develop, not joinmarket:master

jamesphillipturpin:
Pursuant to chris-belcher's comment I have submitted a pull request to joinmarket:develop and will be closing this pull request to joinmarket:master presently.

PR #546, opened 2016-05-30
Allow 'maxsize' and 'minsize' to be type long as well as type int.

jamesphillipturpin:
This change is necessary for compatibility with 32 bit int.  The minimum specification for Python 2.7 int type is 32 bit, and apparently this is not uncommon in Python 2.7 for Windows even when using 64 bit processors. This is only a change to the sanity_check function.


chris-belcher:
This looks good now, no need to close it.

Issue #544, opened 2016-05-28
Sendpayment crashes with "ValueError: Calculated transaction fee is too large for our inputs;"


AlexCato:
Needed several attempts to do a sendpayment.py, because the coinjoin always crashed when it tried to create the transactions. On the console it reports:
"ValueError: Calculated transaction fee of: (...) is too large for our inputs;Please try again."

This seems to be the relevant part of the log:

''''
TIMESTAMP_0000000000104 [MainThread  ] [DEBUG]  got estimated tx bytes: 6388
TIMESTAMP_0000000000105 [MainThread  ] [DEBUG]  Based on initial guess: 10000, we estimated a fee of: 106430
TIMESTAMP_0000000000106 [MainThread  ] [DEBUG]  CRASHING, DUMPING EVERYTHING
''''
Here's the full log: [sendpayment crash](http://pastebin.com/gfyMGfVS)

I only could fix this by manually selecting the parties to coinjoin with. I selected those with high fees, because my guess why this happens is: the cheaper joiners have dozens of dozens of super-small outputs to spend and need many of them to create the desired coinjoin-output-size (in my case, about 2 BTC). This creates a huge transaction with way larger fees than estimated (even setting the tx_fee in joinmarket.cfg to 25 did not help here).
More expensive counterparties probably have less "business" and less scattered mixdepths with bigger concentrations. But there's no way for a taker to know this in advance, he just has to eat those tx-fees unless he bumps into a hardcoded maximum like I did, right?

I dont mind this problem myself, but believe that the default settings should work for the average user. We should make it as convenient and easy as possible for them... maybe let makers also send the single biggest output they could spend to give takers a chance to find makers with fewer required outputs, thus creating a smaller transaction, reducing fees and helping the bitcoin network in the process?

chris-belcher:
Yeah the miner fees part of joinmarket is mess :(
The orders should advertise their fee/kb and pay more when they have lots of UTXOs.

AlexCato:
Two thoughts about this:

An easy fix for the taker scripts: when receiving the transaction data for each maker, calculate the size and appropriate miner fee for that size. Reject all maker-offers which do not add enough miner fees.
Reasoning: this makes sure that per default the taker part works and it's in everyones interest to reduce on-chain data usage (currently makers dont care and takers just pay, they dont even get the miner fee info displayed before agreeing).
[...]

PR #548, opened 2016-05-31
issue #544: reject too many inputs from makers

AlexCato:
If there is more than 2 inputs from a maker, check if the maker paid a sufficient fee for the additional inputs. If not, reject him and choose another one.

Implicit joinmarket assumption: taker pays for ~1 input and 2 outputs of each maker. Some makers send very many inputs
and create a huge transaction, which crashes the tumbler/sendpayment scripts because the fee estimates are too
far off then. 

This issue sparked the idea: https://github.com/JoinMarket-Org/joinmarket/issues/544

### DIFF
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -127,6 +127,22 @@ class CoinJoinTX(object):
             # immediately quit; currently, the timeout thread suffices.
             return
 
+        # check for too many maker inputs, which might create absurd tx fees
+        # and could crash this script later otherwise
+        if len(utxo_data) > 2:
+            # Every maker creates 1 inputs and 2 ouputs, which
+            # are agreed to be paid by the taker implicity.
+            # Allow one additional input by default.
+            # If there are more, check if they are paid for by the maker
+            additional_cost_maker = estimate_tx_fee(len(utxo_data)-2, 0)
+            if (additional_cost_maker > self.active_orders[nick]['txfee']):
+                log.debug('Too many inputs (' + str(len(utxo_data)) + ') from ' + nick
+                    + '. The additinal transaction costs are not covered by the maker either'
+                    + ' (estimated add. fee: ' + additional_cost_maker + ', paid for by maker: '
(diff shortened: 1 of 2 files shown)


chris-belcher:
why 2 inputs ?

AlexCato:
Magic number for which I have no clear justification.

For small joins, 1 would be sufficient.
For a big join, there's the problem that every maker's mixdepths will scatter a lot (thinkining of people offering 50 BTC and then a 48 BTC join comes along...)... so they either have to bump their tx fee contribution or are out of luck.
Thought of 2 as long-term viable middle ground. Probably can run merge_algorithm "gradual" with that and avoid getting too many UTXOs in the long term that way.

How would you handle it?

AlexCato:
After our IRC-chat, we think a good way to improve this PR might be:

- try this approach 3-4 times. If no suitable makers are found, relax the conditions for maker selection
- do this by increasing the allowed amount of inputs per maker. Increase the allowed amount of inputs each iteration.
- add another magic number (max iterations) as stop criterion
- also needs to fix the #544 crash then and allow larger miner fees. Increase the allowed fees each iteration.
- let user confirm the increased miner fee

Aside from that, the sendpayment asks the user to confirm the cj fees before starting to initiate the coinjoin (but after selecting the makers). The miner fees should be displayed as well in the question to the user. Since those are not known at this point in time, the whole confirmation dialogue must be delayed until after the maker inputs are known. This will be another PR. Unsure when I find the time, hope this week.

AlexCato:
Oh, didnt see this the first time i read it:

> I suppose there must be some cutoff, but I think with this the reality is that every single maker out there would be rejected.

I've tested the changes of course, both with smaller amounts (1 btc, 1.5btc, 2.5btc) and with a larger one (>20 btc). Worked like a charm with no problems. It's a small sample size though. It sometimes asked me to confirm 3 or 4 times when it restarted after a timeout, but always found a suitable match.

AlexCato:
Will submit a new PR for this once it's done.

PR #550, opened 2016-06-01
Sendpayment.py: Fix tx fee estimation #544

AlexCato:
If no fee estimation is specified by the user, conservatively use the blockchain fee estimation. The bitcoin blockchain currently estimates 3-4 times higher than the default 5000 specified before this PR.

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -305,6 +305,13 @@ def main():
         wallet = BitcoinCoreWallet(fromaccount=wallet_name)
     jm_single().bc_interface.sync_wallet(wallet)
 
+    # Estimate the minimum realistic fee if it currently is the default value.
+    # At this point we do not know even the number of our own inputs, so
+    # we guess conservatively with 2 inputs and 2 outputs each
+    if options.txfee == 5000:
+        options.txfee = max(options.txfee, estimate_tx_fee(2, 2))
+        log.debug("Estimated miner/tx fee for each cj participant: "+str(options.txfee))
+
     irc = IRCMessageChannel(jm_single().nickname)
     taker = SendPayment(irc, wallet, destaddr, amount, options.makercount,
                         options.txfee, options.waittime, options.mixdepth,


AdamISZ:
The default was not 5000 per kilobyte. See https://github.com/JoinMarket-Org/joinmarket/blob/master/sendpayment.py#L215-L218 , in particular the *per participant* part.

Also note that this value is almost never actually used in practice.

The current code is a bit of a mess; it deliberately crashes in two places, but it is *not* simple to make tx fees work sensibly, and to be fair it has worked surprisingly well (those crashes seem to be very rare). This complexity is especially true in case of sweeps, where you don't know the coinjoin amount at the start.

AlexCato:
Yes, it was the estimated fee per participant.

Currently, the blockchain still estimates way higher than the 5k default (per participant, with 2 inputs and 2 outputs):
[MainThread  ] [DEBUG]  rpc: estimatefee [2]
[DEBUG]  got estimated tx bytes: 372
[DEBUG]  Estimated miner/tx fee for each cj participant: 19137

This indeed does not usually cause problems, but as we've seen in #544, sometimes it does. This should fix it. I'm aware this isnt the perfect tx_fee solution, but it should be a no-downside upgrade to the current code imho.

AdamISZ:
In the case of non-sweep, the value in options.txfee is simply ignored; it's [overwritten](https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/taker.py#L158-L165). In case of sweep, we call estimate_fee [here](https://github.com/JoinMarket-Org/joinmarket/blob/master/sendpayment.py#L64-L79) at the point where we already have the number of txouts and make a crude estimate of txins (estimate is 3 per maker). Only if that estimate is lower than what's coming from options.txfee is options.txfee used (so nowadays, basically never, unless the user has bumped it).

Don't be misled by the "based on an initial estimate" debug line. That is something I should have deleted, since as you can see from the above, in many cases it's completely irrelevant (certainly misleading).

It's similar in the tumbler (separate issue; top level scripts should be refactored so as not to duplicate code like this).

Also, what you've got here is a resetting of options.txfee from a per-participant value to an overall value.

AlexCato:
> Also, what you've got here is a resetting of options.txfee from a per-participant value to an overall value

Unless I am completely blind, my code does estimate per participant; not overall. Transactions are really that expensive these days, 19k satoshi for ~350 bytes (= 1 participant) is not unheard of. This would be even higher if the estimate would use 3 per maker as in other parts of the code.

From what I understand by reading the code, the crash I'm trying to fix by this occurs because the taker did choose not enough inputs to cover the miner fees, because the initial estimation was too low. When the "real" fee is known later, it's too late. Thats why this fix will make sure that enough inputs to cover the higher-than-estimated-by-default are initially chosen.

AdamISZ:
> Unless I am completely blind, my code does estimate per participant; not overall. 

options.txfee is multiplied by makercount [here](https://github.com/JoinMarket-Org/joinmarket/blob/master/sendpayment.py#L119). You are resetting this variable before that multiplication occurs.

> From what I understand by reading the code, the crash I'm trying to fix by this occurs because the taker did choose not enough inputs to cover the miner fees, because the initial estimation was too low. When the "real" fee is known later, it's too late. Thats why this fix will make sure that enough inputs to cover the higher-than-estimated-by-default are initially chosen.

I realise that it superficially looks like a fairly simple oversight, but it isn't. We have to estimate the required amount of bitcoins to be used as input before knowing how big the transaction is (this is a problem specific to joinmarket, ordinary wallets don't have any such problem since they are in a position of full information), and therefore before knowing the correct transaction fee estimate. So we fundamentally *cannot* know the right amount of inputs to take via the utxo selection algorithm.

I tried very hard to make the best balance between practicality (it works most of the time without incomprehensible configuration for users) and correctness (it always works).

I think the biggest open issue is to allow users to reject cases where makers are providing way too many inputs and thus bumping up the required fee; this is *partially* addressed by the fact that the code will simply crash if the selected utxos weren't enough after an initial reasonably generous estimate. But obviously it's clunky. Replacing "crash" with "try again" is obviously good, and I seem to remember I tried to do that a bit in the tumbler case (where it's far more important not to crash).

AlexCato:
> options.txfee is multiplied by makercount here. You are resetting this variable before that multiplication occurs.

Yes... that's why i estimated the fee _per participant_ with that line. It's going to be closer to the true per participant value than the 5000 default value:

> options.txfee = max(options.txfee, estimate_tx_fee(2, 2))

> I think the biggest open issue is to allow users to reject cases where makers are providing way too many inputs and thus bumping up the required fee;

I plan to work on this next, so that per default too many maker-inputs are rejected. If that would lead to not being able to coinjoin, though, I'll relax the number of allowed inputs and request that the user confirms the higher miner fees at the point when they are known.

AdamISZ:
Oh, so you're saying that estimate_tx_fee(2,2) is closer than 5000. Well, that's true in practice, but it's a weird thing to do since estimate_tx_fee is a function to estimate a fee based on total size. Still weirdness aside, I get your point: it's a dynamic value.

I suppose I never considered doing that because there are two cases: if a sweep, we do that estimate_fee trick with a bit more information in the code above, and if not a sweep, it gets ignored anyway. I'm not sure it makes sense (I think it probably doesn't) to do another estimate_tx_fee upfront.

I was at one point thinking we should really remove this variable as it's more likely to be confusing, but the original reason for keeping it probably still stands: there may be some exceptional scenarios where a manual override is useful.

AdamISZ:
After the conversation, I'd probably say: it wouldn't hurt to do this, but I'd tend to think it isn't needed; but maybe it's really more to the point to go over the whole estimation process and improve/streamline it.

AdamISZ:
After a little *more* conversation (on IRC), I agree that this does have extra value, because for the non-sweep case, utxo selection happens based on amount +3*options.txfee. So yes, making that dynamic helps. ACK.

AlexCato:
Since I fail at git, i had to create a new PR for this with only one commit. As bonus i also added this to the tumlber in https://github.com/JoinMarket-Org/joinmarket/pull/552

PR #552, opened 2016-06-01
sendpayment / tumbler: dynamic tx fee estimation

AlexCato:
If the user does not give an initial fee estimation, dynamically estimate the initial fee from the blockchain. Required for non-sweeping sendpayments.

This helps to fix issue #544 . Discussion also in PR #550

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -103,13 +103,13 @@ class PaymentThread(threading.Thread):
             total_amount = self.taker.amount + total_cj_fee + \
 	        self.taker.txfee*self.taker.makercount
             print 'total estimated amount spent = ' + str(total_amount)
-            #adjust the required amount upwards to anticipate a tripling of 
-            #transaction fee after re-estimation; this is sufficiently conservative
+            #adjust the required amount upwards to anticipate an increase in 
+            #transaction fees after re-estimation; this is sufficiently conservative
             #to make failures unlikely while keeping the occurence of failure to
-            #find sufficient utxos extremely rare. Indeed, a tripling of 'normal'
+            #find sufficient utxos extremely rare. Indeed, a doubling of 'normal'
             #txfee indicates undesirable behaviour on maker side anyway.
             utxos = self.taker.wallet.select_utxos(self.taker.mixdepth, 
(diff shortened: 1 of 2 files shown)


AdamISZ:
After the long discussion in #550 , I think this is fine. (annoying we are still not getting Travis build triggers on PRs). Apologies for misunderstanding it at first but I guess it's good we're talking about this rather hairy topic again!

If no more discussion I'll merge it shortly.

sendpayment.py
```diff
@@ -295,6 +295,13 @@ def main():
     else:  # choose randomly (weighted)
         chooseOrdersFunc = weighted_order_choose
 
+    # Dynamically estimate a realistic fee if it currently is the default value.
+    # At this point we do not know even the number of our own inputs, so
+    # we guess conservatively with 2 inputs and 2 outputs each
+    if options.txfee == 5000:
```
chris-belcher: I'd say make the default txfee 0 instead of 5000, to mean estimate it.

AlexCato:
Agreed, belcher. Will use -1 as default.

Also this needs one more modification, otherwise the now approximately accurate estimates are still doubled later in the code which might lead to more often "insufficient funds" errors.

AlexCato:
Would like to remove the multiplication by 2 from this line, so only an 100% increase over the first estimation is allowed instead of 200% increase (and would change the same thing in tumbler.py):

https://github.com/AlexCato/joinmarket/blob/ad4efdbc4254746bcc220e84e6f47f45bb646823/sendpayment.py#L112

Reason: less insufficient funds errors, also the estimates should be good enough to not need that much margin for error

Any objections?

AdamISZ:
@AlexCato yes probably that makes sense. It's more an art than a science at this point; but extra 1x (so 2x in total) feels about right to me.

AlexCato:
Updated the PR with belcher's default and the slightly decreased fee estimation discussed.

chris-belcher:
I think it looks good.
So before we merge this you should squish your commits together into one commit.

Here is the tutorials I used in case it helps
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Issue #496, opened 2016-04-10
New features and tests, coverage of PRs


AdamISZ:
This is motivated by increasing size and complexity.

Two comments:

1/ I would much prefer to stop the approach of adding lots more options and configurations all the time; all of them add extra maintenance. Even ignoring numerical flags, the number of different ways JM can be run is combinatorially exploding: 2-4 blockchain interfaces, testnet/mainnet, secp256k1/not, 4 different bots (and 3 different sub-categories for yg) which can be run in multiple modes, different wallets and wallet-tool in different modes of operation, GUI option/branch/repo etc etc. If a PR adds a new feature that just makes things "slightly better" then it should not be a priority. The toughest challenges we have today are: make the tumbler work reliably, have a scalable messaging layer, and prevent sybil. Focus should be on that imo.

2/ Run tests please! when you introduce new features you should provide tests that cover that code. In some cases, this is hard, but should be done, and also: run tests of the tumbler (that can't be in the build test, it takes too long). I'm going to repeat what I said elsewhere: the testing process has been made *much* easier (using pytest). A tumbler test can be run with (short version): `pytest tests/test_tumbler.py`, the right setup is described in the completely updated "Testing" page on the Wiki.

At some point shortly I will start campaigning to block any non-trivial PR that reduces test coverage. Maybe.

PR #554, opened 2016-06-02
yield-generator-cicada.py

jamesphillipturpin:
Yield Generator Cicada: An attempt to optimize Yield Generator Oscillator for profit, money velocity, and information leakage avoidance.

### DIFF
--- a/yield-generator-cicada.py
+++ b/yield-generator-cicada.py
@@ -1,0 +1,875 @@
+#! /usr/bin/env python
+from __future__ import absolute_import, print_function
+
+import os
+import sys
+import datetime
+import time
+import random
+import decimal
+import ConfigParser
+import csv
+import threading
+import copy
+
+from joinmarket import Maker, IRCMessageChannel, OrderbookWatch
+from joinmarket import blockchaininterface, BlockrInterface
+from joinmarket import jm_single, get_network, load_program_config
+from joinmarket import random_nick
+from joinmarket import get_log, calc_cj_fee, debug_dump_object
+from joinmarket import Wallet
+
+config = ConfigParser.RawConfigParser()
+config.read('joinmarket.cfg')
+mix_levels = 5
+nickname = random_nick()
+nickserv_password = ''
+
+from math import sqrt, pow, ceil, floor, exp
+from math import log as logarithm
+import random
+PHI = (sqrt(5.0)+1.0)/2.0
+phi = (sqrt(5.0)-1.0)/2.0
+
+def is_prime(n):
+  if n == 2 or n == 3: return True
+  if n < 2 or n%2 == 0: return False
+  if n < 9: return True
+  if n%3 == 0: return False
+  r = int(n**0.5)
+  f = 5
+  while f <= r:
+    #print '\t',f
+    if n%f == 0: return False
+    if n%(f+2) == 0: return False
+    f +=6
+  return True
+
+"""
+# 24 hours per day, 29.530587981 days per synodic month.
+# The 2*PHI/logarithm(2) term is intended to make the
+# prime numbers average out to about two months, so time 
+# frames aren't biased towards being out-of-phas lunar effects.
+# You might change 29.530587981 to 365.24 if you use the same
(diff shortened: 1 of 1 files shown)


raedah:
I do not know what the changes are you introduced. No code diff either. Can you say more on what your strategy here is? If this is an improvement, why not merge it into the existing oscillator?

jamesphillipturpin:
It does a regression analysis on past transactions, using recent transactions based on randomly chosen amount ranges and time periods, in order to determine the power law of the form earnings = C+A*pow(amount,B) with the highest correlation, and reduces the coefficient A further based on an average amount of transactions per offer range. Then it picks a new set of amount ranges and time periods and generates earnings with some statistical noise based on uncertainties in A and B. It repeats that process every hour.

raedah:
So math is a strong suite of yours? Not for me. Sounds amazing! Extensive comments explaining the code and usage would probably help developers and users alike to understand exactly what it is doing and what purpose it serves. Maybe it will be clearer for me if I give it a run later. I am very interested in where this could go.

jamesphillipturpin:
Yes, it needs more comments for sure. I have related education and intuition about the math but I also have Google if I forget a detail.

Some cicadas only emerge into adulthood every 13 or 17 years. This allows them to satiate their predators without maintaining the population of their predators. This tidbit of knowledge inspired me to use randomly chosen prime numbers for sampling and oscillation time intervals.  Since only I know my time intervals, takers or snoops can't tell much about my history from my announcements even if they know my methods for choosing optimal prices. Also since the intervals are always various different prime numbers, resonance effects are minimized and prices can stabilize rather than falling into some kind of adverse business cycle where supply and demand are out of phase. So an oscillator that doesn't oscillate above the noise. In summary, the cicada insect inspired ideas that allow me to avoid security through obscurity, and I can announce a specific pricing model based on transaction history without fear that my transaction history could be reverse engineered. So "Cicada" because without that idea I would have to keep my modifications secret out of both altruism and self interest, rather than disclose them out of both altruism and self interest.

AdamISZ:
I know we've discussed this before, but it really seems that adding more yield gen algos should probably be in separate repos? I would point you at #496 but arguably that's not *quite* the point here, this is more about how we manage lots of different user-level scripts, not about testing the core code.

There is also the point about refactoring the user level scripts to avoid code duplication; it would be great if someone could do work on that.

Another related issue is why travis stopped triggering test builds a couple of weeks back (nothing to do with this code of course, just a relevant annoyance).

jamesphillipturpin:
I could put this in a separate repo/fork and maintain it myself.

raedah:
@AdamISZ Your direction makes sense. I also want to add that we do want to support a healthy ecosystem of features being developed and shared. Its mostly just about having it organized in a sane way (such as another repo). In an optimal way that really helps the project thrive.

@jamesphillipturpin >I could put this in a separate repo/fork and maintain it myself.

Probably a good strategy for now. Please do link me, as I want to follow this. There is lots of code duplication in the yield generators currently. Actually most of the code outside of create_my_orders is dups. We really need to modularize the ygs so that there is a single yield generator base class which is part of the /joinmarket/ lib, and a shared functions lib for the ygs. At that point, it would should be very easy for a user to grab and add yg addons. It would be easier to get involved as the yg code would be much more compact to read. We could maintain a separate repository for yg variations, while still providing a sensible default yg for the main joinmarket project that is as turnkey as possible. Any help here is much appreciated.

jamesphillipturpin:
OK, I put it here:
[https://github.com/jamesphillipturpin/joinmarket/blob/master/yield-generator-cicada.py](https://github.com/jamesphillipturpin/joinmarket/blob/master/yield-generator-cicada.py)
So I'll close this pull request.

raedah:
@jamesphillipturpin Please enable issues on your repository so I can discuss with you, or come find me on IRC.

chris-belcher:
It might be good to have a list of other ygen implementations. Maybe a new page on the wiki that anyone can edit to add their own.

Issue #103, opened 2015-06-15
Integrating CoinJoin with Bitcoin Core wallet


chris-belcher:
Bitcoin Core 0.11 has a new feature useful to us.

> This release adds an option -walletbroadcast=0 to prevent automatic transaction broadcast and rebroadcast (#5951). This option allows separating transaction submission from the node functionality.

This could be used to implement CoinJoin for Bitcoin Core too. Set -walletbroadcast=0 and get the tx hex, extract from it the inputs and outputs, create a coinjoin from that information and broadcast it. The GUI might show an odd double spend attempt but the coinjoin tx will win because the non-coinjoin was never broadcast.

Of course we already have --rpcwallet which is somewhat cruder albeit more configurable.
https://www.reddit.com/r/joinmarket/comments/37o29o/new_feature_sending_coinjoins_from_the_bitcoin/

PR #555, opened 2016-06-03
First version of bitcoin core integration

chris-belcher:
An implementation of #103 

To use, run Bitcoin-Qt with walletbroadcast=0 along with all the other configurations for using JoinMarket with Bitcoin Core.

In case the coinjoin triggered by walletnotify fails, you ca run this new script with a txid as an argument and it will attempt to create a coinjoin from it again.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -406,7 +406,7 @@ def bitcoincore_timeout_callback(uc_called, txout_set, txnotify_fun_list,
     log.debug('timeoutfun txout_set=\n' + pprint.pformat(txout_set))
     timeoutfun(uc_called)
 
-class NotifyRequestHeader(BaseHTTPServer.BaseHTTPRequestHandler):
+class NotifyRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
     def __init__(self, request, client_address, base_server):
         self.btcinterface = base_server.btcinterface
         self.base_server = base_server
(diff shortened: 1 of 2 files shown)


joinmarket/blockchaininterface.py
```diff
@@ -749,13 +758,13 @@ def __init__(self, jsonRpc):
...
         self.tick_forward_chain_interval = 2
-	self.absurd_fees = False
+        self.absurd_fees = False
 
-    def estimate_fee_per_kb(self, N):
-	if not self.absurd_fees:
-	    return super(RegtestBitcoinCoreInterface, self).estimate_fee_per_kb(N)
-	else:
-	    return jm_single().config.getint("POLICY", "absurd_fee_per_kb") + 100
+     def estimate_fee_per_kb(self, N):
```
AdamISZ: Seems to be 4 spaces-> 5 ? I think this might be why my editor complains about this particular file's formatting.

chris-belcher: This one line looks like it has 5 spaces instead of 4, so I'll fix that. I believed this is going from tab to 4 spaces.

AdamISZ:
I haven't had a chance to read it (I'm not as knowledgeable about Core internals as some other people around here anyway), but thanks for doing it. I get the basic idea.

Have you been able to do any tests with Bitcoin-Qt? Does it show up correctly after broadcast?

chris-belcher:
After broadcast the original non-coinjoin transaction gets market as the 'conflicted' state, this was known since walletbroadcast=0 was created https://github.com/bitcoin/bitcoin/pull/5951

AdamISZ:
OK to merge with me to develop, if that's what you want to do. We'll need to re-merge it into 0.2.0 (and same for anything else we merge from now into dev)

Congrats on getting it working :)

Issue #120, opened 2015-07-02
Make use of maker contribution to tx fee


domob1812:
As far as I can tell, `sendpayment.py` does not make use (nor even display) the maker's contribution to tx fees.  I can imagine the following change (maybe optional):  Say that I want to specify a (total) tx fee of some amount myself as taker, e. g., 10000 satoshi.  I also want manual order selection.  Then, the screen that displays the offers could explicitly tell me about the fee contribution of the makers towards my desired (total) tx fee.  E. g., if someone offers a join for 5000 satoshi but pays 1000 satoshi as tx fee contribution, the screen could show me an "effective" cost of only 4000 satoshi for that order.  (Because the total cost for me including the desired tx fee is 9000 + 5000 = 10000 + 4000 = 14000 satoshi.)

If I understand the current code correctly, the fee I can set is always simply added to those of the makers.  I think that it would be a nice option to specify the desired total fee instead and show the makers' contributions in the way described above.

Does that make sense?  I'm willing to work on an implementation if yes.

chris-belcher:
I agree with the displaying of miner fee contribution in the manual choosing of orders. Also agree with displaying miner fee contribution in the "chosen orders to fill" message.

Neutral/disagree with the concept of effective cost being coinjoin fee minus maker's miner fee contribution. It is best seen that makers pay for their own inputs and outputs.

Miner's fee are a difficult problem, as the taker doesn't know how many UTXOs the maker will send until he's already chosen which order to fill.

sendpayment.py was always meant to be a very light script on top of the Taker and Wallet classes, the true user interface was always planned to be an actual bitcoin wallet. However I can't think of a better easier way right now to improve the miner fee situation so I am neutral/agree with the idea of a price floor option on the total miner fee. Code it so the *tx fee contribution* and *tx fee floor* options mutually exclusive. Widespread use of the tx fee floor option might make the miner fee contribution by makers go to zero.

domob1812:
I understand that `sendpayment.py` is meant as a light wrapper.  Let me start with the two first points you agree on, namely displaying the offered fee.  This should help, at least, advanced users somewhat.

As it is right now, the offered fee is basically insignificant, right?  If I set my yield generator to offer no fee while my competitor offers one, there won't really be any difference.  Or do some of the other scripts (like the tumbler) use the fee offers as well?

adlai:
[...]

PR #556, opened 2016-06-03
taker.py: initially reject too many maker inputs

AlexCato:
Do not merge, this still needs to be improved... just creating this PR for discussions.

As discussed in #548, I'm working on rejecting makers with too many inputs, which should reduce on-chain data usage and tx fees for takers.

Part of the solution is in this PR.

This works, but currently could end up asking the user to confirm quite often whenever a new maker has to be chosen (unless, of course, the --yes option is used in sendpayment). In my tests it usually was confirm --> wait for timeout --> confirm another cj --> wait for timeout --> confirm again --> finally goes through. Even though I dont mind it being that way, this is no good user experience.

The behavior I would like to see is: automatically choose orders first, and only then ask the user to confirm when all fitting makers have been selected and the inputs, the cj fee and the tx fee are known.
But this would have to be done in taker.py, where I cannot check for the "--yes" option which is only available in sendpayment, and I dont want to bother asking the user if he started with that option enabled.

I'm currently thinking about adding a "yes" parameter variable to the CoinJoinTX class and moving the user prompt to taker.py. Any thoughts on this?

### DIFF
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -73,6 +73,8 @@ class CoinJoinTX(object):
         self.nonrespondants = list(self.active_orders.keys())
         self.all_responded = False
         self.latest_tx = None
+        self.current_max_inputs = 2
+        self.tries_fewer_maker_inputs = 0
         # None means they belong to me
         self.utxos = {None: self.input_utxos.keys()}
         self.outputs = []
@@ -126,6 +128,22 @@ class CoinJoinTX(object):
             # when internal reviewing of makers is created, add it here to
             # immediately quit; currently, the timeout thread suffices.
             return
+
+        # check for too many maker inputs, which might create absurd tx fees
+        # and could crash this script later otherwise
+        if len(utxo_data) > self.current_max_inputs:
+            # Every maker creates 1 inputs and 2 ouputs, which
+            # are agreed to be paid by the taker implicity.
+            # If there are more, check if they are paid for by the maker
+            additional_cost_maker = estimate_tx_fee(len(utxo_data)-self.current_max_inputs, 0)
+            if (additional_cost_maker > self.active_orders[nick]['txfee']):
+                log.debug('Too many inputs (' + str(len(utxo_data)) + ') from ' + nick
+                    + '. This would increase transaction costs for the taker. These additinal '
+                    + 'transaction costs are not covered by the maker either'
+                    + ' (estimated add. fee: ' + str(additional_cost_maker) + ', paid for by maker: '
+                    + str(self.active_orders[nick]['txfee']) + '). '
+                    + 'Ignoring. Will select another maker shortly.')
+                return                      # timeout marks this maker nonresponsive
 
         total_input = sum([d['value'] for d in utxo_data])
         real_cjfee = calc_cj_fee(self.active_orders[nick]['ordertype'],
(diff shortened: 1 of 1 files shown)


jamesphillipturpin:
I'm not yet familiar enough with the tumbler/taker scipts to get into that level of detail yet, but I understand that there is a potential problem with snoops looking at UTXOs from makers and then not signing the transactions, and one of the potential future countermeasure is to allow makers to blacklist takers for not signing transactions after seeing the maker UTXOs - and reserving some UTXOs for only trusted takers once a Sybil attack of snoops is detected. It would be quite annoying for everyone involved if noobs manually rejecting offers from makers trying to gradually collect dust triggered a blacklisting of not only themselves, but also newcomers for fear of a Sybil attack as well.

raedah:
I know this is just being sketched out, but I feel called to bring up that the taker behavior should be consistent with the maker code behavior. We dont want the taker code to reject participants that have many utxos, when its jm's own maker code that is whats serving lots of utxos. Its not an attack if its the default behavior of the project. That being said, it makes sense that a taker would be able to optimize for transaction cost, though cost is not the most important factor. Another factor for example would be the desirability of utxos that are of an older age. I would get rid of the current_max_inputs var. It doesnt make sense to me to drop a participant for increasing the cost of the transaction fee without considering the total cost including the coinjoin fee they are charging.

AlexCato:
The fees do create a problem for takers and need to be dealt with somehow.

The vast majority of coinjoins is between 0.01 and 0.2 BTC in my experience from running a (cheap) yieldgen for several weeks. But which order are taken is only evaluated by the cj fee, which is orders of magnitude lower than the tx fees. Just one single maker with 7 inputs can create tx fees > 0.001 btc, which are already a significant part of the low value coinjoins.
So currently a user gets asked to confirm total coinjoin-fees of 2376 satoshi, but the makers and scripts silently add tx fees of 50000-150000 which the user usually does not even get to see.
I do agree that the total fee tag in the end (cj fee + tx fee) should be taken into consideration instead of just counting inputs. But this is intended as a first step, and already actively allows more inputs than the current_max_inputs var if the maker pays for the additional ones.

What adds to the problem is that most makers have the possibility to avoid this problem, but actively choose not to do it:
- they could add miner fees. Nearly nobody does, even the default usually is changed, to 0.
- they could use merge_algorithm gradual, which should lead to an average of ~2 inputs/maker in the long run and avoid this problem alltogether

But I agree that the maker's code should be adjusted as well.

raedah:
Yes, the cost is hidden right now. Probably just out of a perhaps too simple design. The miner fee contribution is set to zero because it is a feature that should be phased out. Setting a static contribution fee doesn't take into consideration the utxo count for the specific transaction, so its not much different then just modifying the cjfee being charged. My feeling is that short term would be to display more information to the taker, such as prompting to confirm the txfee and additional information about the transaction before it is sent. Probably also good to display the send amount to the user in case of typos. I am weary of implicit behavior that is going to exclude valid participants which could reduce the overall amount of joins happening, but if the code is quite refined then maybe it wont be an issue. I haven't tested it yet. It should really just be a matter of the taker deciding if they are willing to spend the extra satoshis or not.. though I realize we need to figure out what is workable. Longer term, I think we need to work towards a shared txfee negotiation between participants.

adlai:
I'm in favor of a different approach to the same general idea, but this specific approach gets my **NACK**. It's locking us in to keep using the `txfee` parameter, which needlessly complicates market structure [as discussed previously](https://github.com/JoinMarket-Org/joinmarket/issues/120).

As I've described on IRC, a more flexible approach is to initiate the transaction negotiation with a few extra makers, and drop the one(s) with the least favorable offers. The selection metric of coin age divided by input size has the double benefit of avoiding the dustiest makers while also diversifying the set of participants in a sybil-resistant manner (as the only way to "game" this parameter is to invest more coins in your yield generator).

AlexCato:
There's better approaches to this problems as discussed above. Closing this.

Issue #543, opened 2016-05-27
Use Tor stream isolation


raedah:
The current configuration is just for JM to use the default circuit that all of the users other tor traffic is traveling over. JM should be using the tor stream isolation feature to create a distinct connection for JM to use so that there wont be correlation with other applications the user is using over tor.

raedah:
https://www.torproject.org/docs/tor-manual.html.en
"IsolateSOCKSAuth - Don’t share circuits with streams for which different SOCKS authentication was provided. (On by default)"

What are we currently using for SOCKS authentication? Nothing? If it were to be randomized, that should create a new circuit. 

https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/irc.py#L628-L630

This should have worked.

                    """
                    setdefaultproxy(PROXY_TYPE_SOCKS5,
                        self.socks5_host, self.socks5_port,
                        True)
                    """
                    setdefaultproxy(proxytype=PROXY_TYPE_SOCKS5,
                        addr=self.socks5_host,
                        port=self.socks5_port,
                        rdns=True,
                        username=''.join(random.choice(string.lowercase) for i in range(8)),
                        password=''.join(random.choice(string.lowercase) for i in range(8)))

but all I am getting is

  File "/home/x/run/joinmarket/joinmarket/irc.py", line 646, in run
    self.sock.connect(self.serverport)
  File "/home/x/run/joinmarket/joinmarket/socks.py", line 391, in connect
    self.__negotiatesocks5(destpair[0], destpair[1])
  File "/home/x/run/joinmarket/joinmarket/socks.py", line 210, in __negotiatesocks5
    chr(len(self.proxy[5])) + self.__proxy[5])
AttributeError: 'socksocket' object has no attribute 'proxy'

chris-belcher:
Yes,
I've put medium difficulty for now because I don't know how hard or easy it is.

raedah:
Looks like the error is here, https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/socks.py#L209-L210

File "/home/joinmarket/joinmarket/socks.py", line 210, in __negotiatesocks5

AttributeError: 'socksocket' object has no attribute 'proxy'

So im just assuming that socks5 code is broken where user/pass auth is performed.

chris-belcher:
Looks like it. Someone else found that socks5.py file from stack exchange I think (I had also used it previously for coinjumble) and it's had stuff fixed in it.
[...]

Issue #503, opened 2016-04-26
Broadcast through tor


chris-belcher:
An added option to #56 where on broadcasting, the taker connects to a random peer on the p2p network (maybe found with a DNS seed) over tor and pushes it's transaction, then disconnects. For the best most-untracable transaction broadcasting.

Also have an option where it randomly chooses a joinmarket peer to push to, as well as the p2p network over tor.

Probably worth waiting until #470 is made since code can be reused.

raedah:
Could also connect to web services (via tor) to push transactions.
https://live.blockcypher.com/btc/pushtx/
https://blockchain.info/pushtx

AdamISZ:
Closing for merge of #662 ; feel free to reopen if you want @chris-belcher

PR #560, opened 2016-06-04
socks.py: randomize socks5 credentials for tor stream isolation

AlexCato:
As suggested in #543 this PR allows TOR stream isolation by using random username/passwords.

Streams are only isolated when connecting to the clearnet IP address of the irc server though (tested: works).

If the hidden irc service is used, streams are NOT isolated. This is not supported by TOR according to fluffypony from the JM irc channel. Basically: connections to **different** hidden services are isolated by default. Connections to the **same** hidden service **cannot** be isolated.

This will therefor also work nicely for things like connecting to random bitcoin nodes to push transactions anonymously as belcher suggested in #503 .

I think therefor, that there is no downside to this fix.

### DIFF
--- a/joinmarket/socks.py
+++ b/joinmarket/socks.py
@@ -32,6 +32,7 @@ for tunneling connections through SOCKS proxies.
 
 import socket
 import struct
+import random
 
 PROXY_TYPE_SOCKS4 = 1
 PROXY_TYPE_SOCKS5 = 2
@@ -113,8 +114,8 @@ def setdefaultproxy(proxytype=None,
                     addr=None,
                     port=None,
                     rdns=True,
-                    username=None,
-                    password=None):
+                    username=str(random.randrange(10000000, 99999999)),
+                    password=str(random.randrange(10000000, 99999999))):
     """setdefaultproxy(proxytype, addr[, port[, rdns[, username[, password]]]])
 	Sets a default proxy which all further socksocket objects will use,
 	unless explicitly changed.
@@ -207,7 +208,7 @@ class socksocket(socket.socket):
             # Okay, we need to perform a basic username/password
             # authentication.
             self.sendall("\x01" + chr(len(self.__proxy[4])) + self.__proxy[4] +
-                         chr(len(self.proxy[5])) + self.__proxy[5])
+                         chr(len(self.__proxy[5])) + self.__proxy[5])
             authstat = self.__recvall(2)
             if authstat[0] != "\x01":
                 # Bad response


raedah:
tACK

chris-belcher:
Looks great. Untested by me so far.

Connections to the same hidden service probably cant be isolated because the client cant choose the rendezvous point which is what chooses the path to the hidden service.

raedah:
What would be the Pros/Cons of connecting on the clearnet address with isolated channels, vs connecting to the onion hidden service?

AlexCato:
Empirical: the clearnet seems to be more stable... at the last IRC outage, it was accessible hours before the hidden service came back up.
Can't comment on privacy aspects though. From my understanding, hidden services are meant to protect the server owner, not the participant. From that understanding it wouldnt make a difference for JoinMarket purposes.

raedah:
I have been trying to learn more about how this works, and though this code looks good, it may not be what we want to implement as the default behavior. It may need to be commented out. The good thing about this patch is that it will hide information about who owns the yg's to anyone who is monitoring the irc channel, by creating different circuits between instances. I dont know where this cyberguerilla hostname syntax actually comes from. The potential bad side of this patch is that, from some of the people I have been able to talk to, it seems that increasing the number of circuits, while using the same origin and destination, may be increasing the number of attack vectors leading to you. More research can be done here.

AlexCato:
We could still suggest/leave the default in the joinmarket.cfg file to connect to the hidden irc service.
If someone misconfigures something and accidently connects to the IRC clearnet server with the socks5 tor proxy enabled, I'd still argue that this is an improvement for them over not having their streams isolated.

I'm running with this intentionally to the clearnet for the last 3 days now, and whenever I get disconnected from IRC, I like the fact that not only my irc-nick changes, but also my displayed cyberguerrilla-identifier (that username@...) is completelty different.

AdamISZ:
Sorry not to comment on this earlier but I don't know much about it and don't have the time/bandwidth to get involved. @chris-belcher do you consider it mergeable?

Would be possible to add some kind of test for this? AFAIR we have no testing of the socks functionality at all yet. It occurs to me that miniircd can't support such a connection, but I haven't looked into it.

raedah:
I am fine with it being merged. This code is an improvement. I just want it known that we still need more research on exactly what Tor is doing behind the scenes. I have tried reaching out to some Tor experts, but no one really seems to have the answer. I will keep looking into it.

chris-belcher:
I'm happy with merging this because it's been tested somewhat and doesn't obviously break.

AdamISZ:
OK, based on a tested ack and consensus, merging.

Issue #562, opened 2016-06-07
yield-generator-oscillator crash: "list index out of range"


dan-da:
```
2016-06-06 14:43:38,637 [MainThread  ] [DEBUG]  Traceback (most recent call last):
  File "yield-generator-oscillator.py", line 715, in main
    irc.run()
  File "/home/joinmarket/joinmarket/joinmarket/irc.py", line 657, in run
    self.__handle_line(line)
  File "/home/joinmarket/joinmarket/joinmarket/irc.py", line 540, in __handle_line
    self.__handle_privmsg(_chunks[0], _chunks[2], get_irc_text(line))
  File "/home/joinmarket/joinmarket/joinmarket/irc.py", line 491, in __handle_privmsg
    self.__on_privmsg(nick, parsed) 
  File "/home/joinmarket/joinmarket/joinmarket/irc.py", line 369, in __on_privmsg
    self.on_order_fill(nick, oid, amount, taker_pk)
  File "/home/joinmarket/joinmarket/joinmarket/maker.py", line 246, in on_order_fill
    taker_pubkey)
  File "/home/joinmarket/joinmarket/joinmarket/maker.py", line 63, in __init__
    utxo_list, utxo_data = populate_utxo_data()
  File "/home/joinmarket/joinmarket/joinmarket/maker.py", line 52, in populate_utxo_data
    self, oid, amount)
  File "yield-generator-oscillator.py", line 508, in oid_to_order
    lmd = sorted(filtered_mix_balance, key=lambda x: x[1],)[-1]
IndexError: list index out of range 

done
```

raedah:
Please post the output of 'wallet-tool.py yourwal.json summary' if you can.
I am guessing this is an edge case where there is not enough funds in any mixdepth to fulfill the offer.

dan-da:
Your guess appears correct.  The last mix_depth messages before the crash were:

```
2016-06-06 14:43:38,611 [MainThread  ] [DEBUG]  mix depths that have enough = []
2016-06-06 14:43:38,611 [MainThread  ] [DEBUG]  mix depths that have enough with output_size_min, []
```

AlexCato:
PR #551 fixes this for the other yield generators.
I'll fix this for this oscillator later today.

AlexCato:
Fixed in #566 . If you download the new oscillator-file from the development-branch, this shouldnt happen any more.

dan-da:
thx!  trying it now..

PR #564, opened 2016-06-08
yg-oscill: fix empty filtered_mix_balance crash

AlexCato:
Fixes issue #562
Log tried to access variable before it was checked

### DIFF
--- a/yield-generator-oscillator.py
+++ b/yield-generator-oscillator.py
@@ -493,15 +493,15 @@ class YieldGenerator(Maker, OrderbookWatch):
         filtered_mix_balance = [m
                                 for m in mix_balance.iteritems()
                                 if m[1] >= total_amount + output_size_min]
-        log.debug('mix depths that have enough with output_size_min, ' + str(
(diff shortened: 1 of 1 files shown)


yield-generator-oscillator.py
```diff
@@ -493,15 +493,15 @@ def oid_to_order(self, cjorder, oid, amount):
         filtered_mix_balance = [m
                                 for m in mix_balance.iteritems()
                                 if m[1] >= total_amount + output_size_min]
-        log.debug('mix depths that have enough with output_size_min, ' + str(
-            filtered_mix_balance))
         try:
             len(filtered_mix_balance) > 0
```
AdamISZ: This line evaluates to False for empty filtered_mix_balance, but it doesn't throw an Exception, does it?

AlexCato:
Do not merge yet.
Will check this when I'm home in a few hours and test.

AdamISZ:
You can use the same syntax I used in #551 presumably. Haven't checked if this is different.

PR #565, opened 2016-06-08
Alternative Bitcoin Core integration

veqtrus:
Apparently @chris-belcher has been working on something similar in #555 but I had this ready before I noticed the PR so here it is.


chris-belcher:
Really we developers should be talking to each other more so work like this isn't repeated.

veqtrus:
I did discuss this on IRC 7 days ago and waxwing told me you were working on it but...

WyseNynja:
What was the main goal of this change? What do we still need from it?

Issue #567, opened 2016-06-09
error: unable to fill order constrained by dust avoidance


johnjacksonbtc:
Thanks to bot I hit an issue using yield-generator-mixdepth.py. It seems it is related with TODO in maker.py. Would be this error triggered if private key for most dusty address `1Pw4sq7DtZTrN3yQR1fV6WPX5TM6Erva54` would not be included in joinmarket wallet?

```
2016-06-09 10:01:41,174 [MainThread  ] [DEBUG]  <<pubmsg nick=Nojacivas message=!orderbook
2016-06-09 10:01:41,440 [MainThread  ] [DEBUG]  <<privmsg nick=Nojacivas message=!fill 0 266289 284d7963363e9f4074f40471596fa55ccc2fe15e68bbfad55f8c1b90072e4705
2016-06-09 10:01:41,440 [MainThread  ] [DEBUG]  new cjorder nick=Nojacivas oid=0 amount=266289
2016-06-09 10:01:41,456 [MainThread  ] [DEBUG]  get_utxos_by_mixdepth = 
{0: {u'021eebe09a94bf929b2cecc70d35c12565bb3c438f773bcdda0298a2241013ab:30': {'address': u'13kQxEXVUfSFpFbe96gkxhoNZvSp8mSyDY',
                                                                              'value': 9940},
     u'07d1220e4584f4eadfd69055e4a72de298560b841fd54515c5fe178870494f47:107': {'address': u'1MJCGKHYmxxNs39xWxUiaZ3zoZZvFjSWsM',
                                                                               'value': 10643},
     u'0f150126bb84bfc53d2ec575ef44e49840f29f6dcbc572a5f57329dbdc7f57f5:25': {'address': u'1Pw4sq7DtZTrN3yQR1fV6WPX5TM6Erva54',
                                                                              'value': 3367},
     u'188c631d8c8c15bf28b2ccae7cb5b0ae321d42ede6218fb38ea79ff3af8e45b6:555': {'address': u'13kQxEXVUfSFpFbe96gkxhoNZvSp8mSyDY',
                                                                               'value': 9620},
     u'1966ce9c0204d9970688cfe7f7be31d944db1c916e6bafcece026ddb2a4a5e74:6': {'address': u'1Pw4sq7DtZTrN3yQR1fV6WPX5TM6Erva54',
                                                                             'value': 3928},
[...]

PR #569, opened 2016-06-11
yieldgens: fix issue 567 for max_size orders

AlexCato:
Fixes issue #567 

Problematic were these lines: https://github.com/JoinMarket-Org/joinmarket/blob/master/yield-generator-mixdepth.py#L157-161

If a txfee > the dust_threshold was set and maximum size order was requested, no mixdepth could contain enough coins to cover amount+txfee. This mostly wasnt an issue until now because most makers have a txfee=0 or did not even notice, because this does not crash the yieldgen-bot.

So this PR sets the announced maximum order to:
max_coins_in_one_mixdepth - max(dust_threshold, txfee)

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -85,7 +85,7 @@ class YieldGenerator(Maker):
         order = {'oid': 0,
                  'ordertype': ordertype,
                  'minsize': minsize,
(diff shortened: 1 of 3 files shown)


raedah:
okay by me. code looks fine.

chris-belcher:
looks fine to me

CohibAA:
tACK, yield-generator-basic.py

Issue #522, opened 2016-05-10
repetitive 'importing 300 addresses into account'


raedah:
Steps
1. Start with a fresh bitcoin wallet.dat to hold your watch addresses on the rpc server.
2. Launch joinmarket with a jm wallet you have already used for sometime.
3. see:  too few addresses in [(0, 1), (1, 1), (2, 1), (3, 1)]
             importing 300 addresses into account joinmarket-wallet-abc123
4. says, restart Bitcoin Core with -rescan, do that, takes forever
5. repeat steps 2,3,4 over and over. Does it ever end?

AdamISZ:
I think we've discussed this before, right? Is https://www.reddit.com/r/joinmarket/comments/3te8lm/too_few_addresses_in_3_1_4_0/cx5frsx not solving it? (not that I'm saying it shouldn't be fixed up, just wondering if you're talking about something new or the same thing)

raedah:
Yes, we have discussed this before. Filed so that we can look for a permanent fix that doesn't require changing the code manually.

raedah:
On some runs, such as the first run, it doesnt say "too fee addresses". It just goes straight to the importing step.  The code that causes this is these 15 lines here: https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/blockchaininterface.py#L591-L607

It just imports 20 addresses for internal and external and for each mixdepth level, so 200 total.
Instead of doing that, it should be using the wallet index_cache numbers.

chris-belcher:
How about an option in the .cfg file which controls how many addresses to import per branch ?

raedah:
the .cfg could be done, but I think what is most important here is that the address importing recognizes the wallet index_cache

AdamISZ:
> But most important here is that the address importing recognizes the wallet index_cache so that recoveries do not fail.

Having recently done a lot of work on this (see #574 ) I am of the opinion that it *should* indeed be the case that simply repeating over and over again until it works is enough. And this includes the case of a miscreant counterparty requesting lots of addresses. The index_cache field in the wallet file is needed for the sync to happen correctly, and it is not changed between runs of sync_wallet. (It might be a bit "in the weeds" but note this comment: https://github.com/JoinMarket-Org/joinmarket/pull/574#discussion_r67148491)

[...]

PR #574, opened 2016-06-15
modify wallet sync to account for requested but unused addresses, add…

AdamISZ:
… testing features for sync

This was motivated by problems people were having with sync, but the main point is not really that: the change in `blockchaininterface` here is addressing this condition: addresses are requested in a yield generator in `CoinJoinOrder.__init__` which are not used, the index cache is updated. On restart, the syncing process as-was would correctly wait until the index_cache value was reached, but at the end would assign the index to the `last_used_addr+1`, where `last_used_addr` is coming from the rpc `listtransactions`, i.e. addresses actually used in transactions, not accounting for the fact that the index (and cache) had been updated to a higher number due to the requests. This change now sets the index to the index cache value in cases where it's higher than the `last_used_addr`.

I'll cross apply the change to blockr if there's agreement. Probably just add one more commit.

There are other points to mention here, I'll make separate comments.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -161,9 +161,9 @@ class BlockrInterface(BlockchainInterface):
                 if last_used_addr == '':
                     wallet.index[mix_depth][forchange] = 0
                 else:
-                    wallet.index[mix_depth][forchange] = wallet.addr_cache[
-                                                             last_used_addr][
-                                                             2] + 1
+		    next_avail_idx = max([wallet.addr_cache[last_used_addr][2]+1,
+		                        wallet.index_cache[mix_depth][forchange]])
+		    wallet.index[mix_depth][forchange] = next_avail_idx
 
     def sync_unspent(self, wallet):
         # finds utxos in the wallet
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -637,8 +637,9 @@ class BitcoinCoreInterface(BlockchainInterface):
                 if last_used_addr == '':
                     wallet.index[mix_depth][forchange] = 0
                 else:
-                    wallet.index[mix_depth][forchange] = \
-                        wallet.addr_cache[last_used_addr][2] + 1
+		    next_avail_idx = max([wallet.addr_cache[last_used_addr][2]+1,
+		                        wallet.index_cache[mix_depth][forchange]])
+		    wallet.index[mix_depth][forchange] = next_avail_idx
 
         wallet_addr_list = []
         if len(too_few_addr_mix_change) > 0:
--- a/joinmarket/wallet.py
+++ b/joinmarket/wallet.py
@@ -10,7 +10,7 @@ from getpass import getpass
 
 import bitcoin as btc
 from joinmarket.slowaes import decryptData
-from joinmarket.blockchaininterface import BitcoinCoreInterface
+from joinmarket.blockchaininterface import BitcoinCoreInterface, RegtestBitcoinCoreInterface
 from joinmarket.configure import jm_single, get_network, get_p2pk_vbyte
 
 from joinmarket.support import get_log, select_gradual, select_greedy, \
@@ -143,7 +143,7 @@ class Wallet(AbstractWallet):
         for i in range(self.max_mix_depth):
             self.index.append([0, 0])
 
-    def read_wallet_file_data(self, filename):
+    def read_wallet_file_data(self, filename, pwd=None):
         self.path = None
         self.index_cache = [[0, 0]] * self.max_mix_depth
         path = os.path.join('wallets', filename)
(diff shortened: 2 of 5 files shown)


joinmarket/wallet.py
```diff
@@ -143,7 +143,7 @@ def __init__(self,
         for i in range(self.max_mix_depth):
             self.index.append([0, 0])
 
-    def read_wallet_file_data(self, filename):
+    def read_wallet_file_data(self, filename, pwd=None):
```
AdamISZ: This non-intrusive addition to wallet setup for testing (see TestWallet class in commontest.py) was to allow automatically opening an encrypted wallet. Seems not to be an issue, but of course do review.

test/test_wallets.py
```diff
@@ -10,28 +10,139 @@
...
+    while not jm_single().bc_interface.wallet_synced:
+        #Wallet.__init__() resets index to zero.
+        wallet.index = []
+        for i in range(5):
+            wallet.index.append([0, 0])
+        #Wallet.__init__() also updates the cache index
+        #from file, but we can reuse from the above pre-loop setting,
+        #since nothing else in sync will overwrite the cache.
+
+        #for regtest add_watchonly_addresses does not exit(), so can
```
AdamISZ: I'm inclined to believe that, as this comment suggests, we can do this for the main user level scripts. Just re-do sync several times until it works. Understood that it's not a trivial question. I do understand why the stop-restart approach is currently used: we can only pick up transactions with listtransactions from addresses that are already imported, so once we find out we have not imported enough, we need to go back to the beginning in some sense. But I believe doing it like this - resetting wallet.index to all zeros, picking up the cache from the file once only and keeping it (since it isn't modified during sync), and then just looping until wallet_synced is True (or some sanity iteration number for failure), might be viable.

joinmarket/blockchaininterface.py
```diff
@@ -637,8 +637,9 @@ def sync_addresses(self, wallet):
                 if last_used_addr == '':
                     wallet.index[mix_depth][forchange] = 0
                 else:
-                    wallet.index[mix_depth][forchange] = \
-                        wallet.addr_cache[last_used_addr][2] + 1
+		    next_avail_idx = max([wallet.addr_cache[last_used_addr][2]+1,
+		                        wallet.index_cache[mix_depth][forchange]])
+		    wallet.index[mix_depth][forchange] = next_avail_idx
```
AdamISZ: The several tests (most commented out for build) in test_wallet_sync in test_wallets.py have indicated that this gives the "correct" result in spamming and non-spamming case: example: you do some transactions and use 3 addresses in a branch, then a spammer requests 4 more, this will return index 7 (i.e. the 8th, after the first 7 were used up).

test/test_wallets.py
```diff
@@ -10,28 +10,139 @@
...
+        #just repeat as many times as possible. This might
+        #be usable for non-test code (i.e. no need to restart the
+        #script over and over again)?
+        sync_count += 1
+        log.debug("TRYING SYNC NUMBER: " + str(sync_count))
+        jm_single().bc_interface.sync_wallet(wallet)
+        #avoid infinite loop on failure.
+        assert sync_count < 10
```
AdamISZ: I'm inclined to believe that, as this comment suggests, we can do this for the main user level scripts. Just re-do sync several times until it works. Understood that it's not a trivial question. I do understand why the stop-restart approach is currently used: we can only pick up transactions with listtransactions from addresses that are already imported, so once we find out we have not imported enough, we need to go back to the beginning in some sense. But I believe doing it like this - resetting wallet.index to all zeros, picking up the cache from the file once only and keeping it (since it isn't modified during sync), and then just looping until wallet_synced is True (or some sanity iteration number for failure), might be viable.

chris-belcher:
I've read the changes and they look pretty good.

AdamISZ:
Cross referencing [this](https://github.com/JoinMarket-Org/joinmarket/issues/522#issuecomment-226244522) comment.

Short version, if you are not recovering, i.e. you have the wallet.json with index_cache in place, I believe that simply enough repeats of sync_wallet ought to always be enough (no rescan, no gap limit changes etc.)

AdamISZ:
I'm quite keen to merge this as it is closing a potential (fairly small admittedly) privacy hole. And the change to running code is small and isolated. However, it is also pretty important, so being a little cautious. Any other comments?

raedah:
I havent been able to review it, but this is currently the part of the code that gives me the most annoyance. Doing anything non-standard seems to end up with an endless barrage of import and restart msgs.

AlexCato:
unsure if it is this pull's fault, but my wallet-tool seems to fail now:

  File "wallet-tool.py", line 103, in <module>
    jm_single().bc_interface.sync_wallet(wallet)
  File "/home/joinmarket/blockchaininterface.py", line 78, in sync_wallet
    self.sync_addresses(wallet)
  File "/home/joinmarket/blockchaininterface.py", line 641, in sync_addresses
    wallet.index_cache[mix_depth][forchange]])
IndexError: list index out of range

Backstory: tried to use sendpayment, ended up in an infinite loop of importing more addresses and if it finally was done (-g 120), it still failed because of "not enough funds" even though the wallet-tool did display way more bitcoins in the mixdepth to send from. Now the wallet-tool displays the error above after i've used sendpayment once with this PR checked out. Could also be the fault of something else though.

raedah:
I can mention that one issue I have noticed with the existing code base is that if you use the wallet-tool.py importprivkey  to import a private key from another jm wallet, then whenever you run a wallet-tool.py summary, it will have to reimport what seems to be all the addresses of the wallet again, probably because it is reassigning the privkey to a new account, and that throws everything out of sync somehow. This causes a condition where you have to reimport all addresses whenever you switch which wallet you are loading.

Issue #573, opened 2016-06-14
Add option to display master xpub for each mix level


dan-da:
Tools such as [mybitprices.info](https://mybitprices.info/hd-wallet-addrs.html) (or [this](https://github.com/dan-da/hd-wallet-addrs) for more privacy) make it easy to audit a wallet given only a master xpub.   This is useful for validating correctness of the joinmarket wallet and for generating reports and information that it does not display.

wallet-tool.py presently displays the xpub for external addresses eg m/0/4/0, but for auditing purposes the parent xpub is needed, ie: m/0/4.  Typically bitcoin wallets make the xpub key at this level user-accessible.

Also ideally the master xpub and xpriv for the entire wallet should be user-accessible somehow.  For purposes of backup and disaster recovery.

If joinmarket devs do not wish to add this functionality to public code I understand, but I would appreciate a pointer on how to do it, for my own private use.

For my purposes, I already modified the code in wallet-tool.py to display the m/0/4/1 xpub.  That was very easy because it already prints m/0/4/0 so it was a single line change.  However, I am unclear how to display the parent xpub.   Or also the wallet master xpub/xpriv.

thanks.

chris-belcher:
To print out the xpub master key you add this

    print(btc.bip32_privtopub(master))

after this line

https://github.com/JoinMarket-Org/joinmarket/blob/master/joinmarket/wallet.py#L143


I'll leave it to someone else to think about the privacy and security aspects and decide whether it's a good idea to do this or not.

dan-da:
hey thanks.   that's simple, I get it now.  I'll submit a pull request for consideration later.

OverlordQ:
Like ''rm'' could expose it behind an explicit long form option called something like ''--expose-keys", to give some negative connotations to including it when running the script.

PR #575, opened 2016-06-15
adds --show-xpub option to display wallet xpub and for each mix level.

dan-da:
This pull request addresses issue #573.

User facing changes:
- new flag --show-xpub  "Display master xpub key for wallet and each mix level"
-  if flag is present:
     1. wallet xpub is displayed first with label: "wallet xpub:"
     2. mixlevel xpub is displayed before addresses with label: "xpub:".
- if flag is not present, no xpub are displayed, not even for external addresses.

note: This pull request removes the display of xpub for external addresses (m/0/x/0).  In my judgement, that info is mostly redundant to m/0/x xpub and clutters up the screen.   A possible enhancement could be to add a verbosity level to the --show-xpub flag, in which case it could print xpub for both m/0/x/0 and m/0/x/1.   Or we could simply keep the previous behavior.

note:  This pull request intentionally does not address display of xpriv keys as there are potentially more privacy issues around that, and it could easily be implented with a separate flag, eg --show-xpriv.

Let me know your thoughts.

### DIFF
--- a/joinmarket/wallet.py
+++ b/joinmarket/wallet.py
@@ -139,13 +139,13 @@ class Wallet(AbstractWallet):
         if extend_mixdepth and len(self.index_cache) > max_mix_depth:
             self.max_mix_depth = len(self.index_cache)
         self.gaplimit = gaplimit
-        master = btc.bip32_master_key(self.seed, (btc.MAINNET_PRIVATE if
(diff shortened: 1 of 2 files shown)


AdamISZ:
PRs should be against develop, not master.

I'd point people generally at #496 if submitting a PR without tests, but this is a very minor change in display, so that doesn't *really* apply (you could in theory add a test, but honestly there are more important things to be fixing up). I'll let others decide :)

Issue #493, opened 2016-04-07
optionally show the xpub bip32 key for external branches of the wallet


chris-belcher:
Allows people to receive payments to their wallet with a new address each time.

From: https://www.reddit.com/r/joinmarket/comments/4dn78i/possible_to_generate_display_an_xpub_key/

Because address reuse leads to privacy leaks, this improvement is well within JoinMarket's aim of improving bitcoin privacy.

PR #577, opened 2016-06-17
add --show-xpub option

dan-da:
note: I closed pr #575 and opened this instead based on develop branch instead of master because I don't know how to modify the original pr.

This pull request addresses issue #573.

User facing changes:

    new flag --show-xpub "Display master xpub key for wallet and each mix level"
    if flag is present:
        wallet xpub is displayed first with label: "wallet xpub:"
        mixlevel xpub is displayed before addresses with label: "xpub:".
    if flag is not present, no xpub are displayed, not even for external addresses.

note: This pull request removes the display of xpub for external addresses (m/0/x/0). In my judgement, that info is mostly redundant to m/0/x xpub and clutters up the screen. A possible enhancement could be to add a verbosity level to the --show-xpub flag, in which case it could print xpub for both m/0/x/0 and m/0/x/1. Or we could simply keep the previous behavior.

note: This pull request intentionally does not address display of xpriv keys as there are potentially more privacy issues around that, and it could easily be implented with a separate flag, eg --show-xpriv.

Alternatively, the --show-xpub flag could be removed, so that xpubs are always displayed and there is is one less flag/option.   ( per discussion in #496 )

Let me know your thoughts.

### DIFF
--- a/joinmarket/wallet.py
+++ b/joinmarket/wallet.py
@@ -130,13 +130,13 @@ class Wallet(AbstractWallet):
         if extend_mixdepth and len(self.index_cache) > max_mix_depth:
             self.max_mix_depth = len(self.index_cache)
         self.gaplimit = gaplimit
-        master = btc.bip32_master_key(self.seed, (btc.MAINNET_PRIVATE if
+        self.master_key = btc.bip32_master_key(self.seed, (btc.MAINNET_PRIVATE if
             get_network() == 'mainnet' else btc.TESTNET_PRIVATE))
-        m_0 = btc.bip32_ckd(master, 0)
-        mixing_depth_keys = [btc.bip32_ckd(m_0, c)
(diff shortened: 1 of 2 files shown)


dan-da:
comments?

AdamISZ:
I hadn't looked at this at all, so this is additional to #493 right, i.e. add xpubs for whole mixdepth and for whole wallet?

AFAIU #493 was based on the idea that you can give out an xpub for people to send funds to your wallet, while this is motivated differently. Quoting @chris-belcher from #573 "I'll leave it to someone else to think about the privacy and security aspects and decide whether it's a good idea to do this or not." The problem is I don't have time to look at everything, and this one seemed to slip through the gaps. Does anyone else have an opinion about it?

WyseNynja:
Do we still need this? It looks like wallet-tool already has something to print xpubs

PR #586, opened 2016-07-01
Wallet sync imports at least what is in index_cache, reducing number …

AdamISZ:
…of restarts needed for initial sync in some contexts.

Basically, the initial section of the sync_addresses code no longer just imports addr_req_count for each branch, but rather the minimum multiple of addr_req_count that is larger than what is in the index_cache (if it exists - if not, the behaviour is unchanged).

Generally this should improve the specific situation in which the user has a wallet.json file with an index_cache, but does not have the required addresses imported into the Bitcoin Core wallet. I expect that if using an existing wallet.json file on a fresh Core instance, 2 steps should be required only: run wallet-tool, do rescan, then run wallet-tool again only once (because all necessary addresses should be imported in one step.

Some extra tests added. @raedah or anyone else, if you have a scenario you can easily test, please do so, thanks.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -586,7 +586,7 @@ class BitcoinCoreInterface(BlockchainInterface):
 		#In cases where the Bitcoin Core instance is fresh,
 		#this will allow the entire import+rescan to occur
 		#in 2 steps only.
-		if wallet.index_cache != [[0,0]]*5:
+		if wallet.index_cache != [[0,0]]*wallet.max_mix_depth:
 		    #Need to request N*addr_req_count where N is least s.t.
 		    #N*addr_req_count > index_cache val. This is so that the batching
 		    #process in the main loop *always* has already imported enough
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -575,12 +575,37 @@ class BitcoinCoreInterface(BlockchainInterface):
             return
         log.debug('requesting wallet history')
         wallet_name = self.get_wallet_name(wallet)
+	#TODO It is worth considering making this user configurable:
         addr_req_count = 20
         wallet_addr_list = []
         for mix_depth in range(wallet.max_mix_depth):
             for forchange in [0, 1]:
+		#If we have an index-cache available, we can use it
+		#to decide how much to import (note that this list
+		#*always* starts from index 0 on each branch).
+		#In cases where the Bitcoin Core instance is fresh,
+		#this will allow the entire import+rescan to occur
+		#in 2 steps only.
+		if wallet.index_cache != [[0,0]]*5:
+		    #Need to request N*addr_req_count where N is least s.t.
+		    #N*addr_req_count > index_cache val. This is so that the batching
+		    #process in the main loop *always* has already imported enough
+		    #addresses to complete.
+		    req_count = int(wallet.index_cache[mix_depth]
+		                    [forchange]/addr_req_count) + 1
+		    req_count *= addr_req_count
+		else:
+		    #If we have *nothing* - no index_cache, and no info
+		    #in Core wallet (imports), we revert to a batching mode
+		    #with a default size.
+		    #In this scenario it could require several restarts *and*
+		    #rescans; perhaps user should set addr_req_count high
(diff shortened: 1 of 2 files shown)


joinmarket/blockchaininterface.py
```diff
@@ -575,12 +575,37 @@ def sync_addresses(self, wallet):
...
         wallet_addr_list = []
         for mix_depth in range(wallet.max_mix_depth):
             for forchange in [0, 1]:
+		#If we have an index-cache available, we can use it
+		#to decide how much to import (note that this list
+		#*always* starts from index 0 on each branch).
+		#In cases where the Bitcoin Core instance is fresh,
+		#this will allow the entire import+rescan to occur
+		#in 2 steps only.
+		if wallet.index_cache != [[0,0]]*5:
```
adlai: This line assumes that ```wallet.max_mix_depth == 5```. That's the default but is still an unsafe assumption.

AdamISZ: Good catch, thanks. One of those "this'll do for a test, fix it up later..." things :)

joinmarket/blockchaininterface.py
```diff
@@ -633,12 +658,26 @@ def sync_addresses(self, wallet):
...
-		    next_avail_idx = max([wallet.addr_cache[last_used_addr][2]+1,
-		                        wallet.index_cache[mix_depth][forchange]])
+		#index setting here depends on whether we broke out of the loop
+		#early; if we did, it means we need to prepare the index
+		#at the level of the last used address or zero so as to not
+		#miss any imports in add_watchonly_addresses.
+		#If we didn't, we need to respect the index_cache to avoid
+		#potential address reuse.
+		if breakloop:
+		    if last_used_addr == '':
```
AdamISZ: This "if breakloop" section is as before #574 . we will always start the addr_req_count*3 final step from either zero or the last used address; in the latter case, since no address could be marked "last used" without already having been imported, I believe it's impossible to end up with a "gap" of missed imports.

AdamISZ:
Making a note to give more context to the most recent commit; I put this on IRC, so just copying for anyone who didn't see that:

I'm now sure that #574 introduced a bug: in order to cover a privacy edge case, i introduced a scenario where it was possible that imports would be missed during sync, and never get picked up in reruns.

Specifically, the point is that the wallet.index value used when calling `add_watchonly_addresses` at the end of sync, when there were too few addresses imported, was, before #574 , starting with the `last_used_addr`. Whereas after #574 in that situation, the index would be set to `next_avail_idx` which can be (in that case, would be) the `index_cache` value, which will usually be a much higher number. 

Example:

index_cache [(100,0),....]
last used address: [ (71, 0), ...]

Suppose addr_req_count = 20.

Then after the first run we have imported 20 on the first branch (it quits early). On the second run, we import another 60 (ARQ*3) from the final step, because as soon as we pass the first batch of 20, we immediately hit `breakloop` condition. On the third run, we already have the first 80 (20+60) imported, but hit breakloop as soon as we try address 81 (i.e. the 5th batch of 20). At this point, due to the #574 code, the next_avail_idx was set to max(71, 100) where 71 was the last used address, and so the wallet.index for the first branch was set to 100. This means that the last (ARQ*3) imports would start from 100, so 100-160. This leaves us in a situation where 1-80 are imported, and 100-160 are imported, i.e. a gap between 80 and 100. This gap would repeat on any subsequent run, resulting in an infinite loop.

AlexCato:
tACK

Tested on an old wallet.json file which needed quite a few addresses to be imported. Tested scripts: wallet-tool.py, yield-generator-basic.py

Pre-#586: neverending restarting with partial imports
Post-#586: one start for import, one start for the thing I actually wanted to do.

Cant think of an edge case to test.

AdamISZ:
@AlexCato thanks very much.

I'll do a bit more testing today and then merge it. It's not an ideal situation, ideally various different tests and testers, but we are currently in a rather botched situation, so even if this is not right it's better to move the develop branch to this code I think.

AdamISZ:
A few more manual tests with different wallet inserts and index_caches seem to work OK. Merging with the above caveats. Still welcoming other comments!

raedah:
Tested this and it worked great. thx Adam.

PR #589, opened 2016-07-04
Fix #588 Wallet list index out of range error

### LINKED ISSUE #588: wallet-tool.py crash: IndexError: list index out of range
```python wallet-tool.py wallet.json
2016-07-03 02:34:16,254 [MainThread  ] [DEBUG]  hello joinmarket
2016-07-03 02:34:20,821 [MainThread  ] [DEBUG]  requesting wallet history
2016-07-03 02:34:21,045 [MainThread  ] [DEBUG]  rpc: getaddressesbyaccount ['joinmarket-wallet-censored']
2016-07-03 02:34:21,066 [MainThread  ] [DEBUG]  rpc: listtransactions ['joinmarket-wallet-censored', 1000, 0, True]
  File "wallet-tool.py", line 103, in <module>
    jm_single().bc_interface.sync_wallet(wallet)
  File "censored/joinmarket/blockchaininterface.py", line 78, in sync_wallet
    self.sync_addresses(wallet)
  File "censored/joinmarket/blockchaininterface.py", line 641, in sync_addresses
    wallet.index_cache[mix_depth][forchange]])
IndexError: list index out of range
```
Tried the same with an other wallet and it worked as expected, so there might be something special with this one wallet that gave me the error.
Running current dev branch

AlexCato:
Uses the number of mixdepths from the wallet file


AlexCato:
I think I found a problem with this solution:

Given there's a standard wallet with 5 mixdepths, but you start a yieldgenerator with a max_mixdepth-configuration of fewer than 5: then there will be  wallet.json-file will be changed to contain the fewer index caches.
This PR then will make it impossible to ever be able to use the original number of mixdepths again.

Proposed solution:
Changing the default value of max_mixdepth in the configuration to -1 (from 5) and only read the number of mixdepths from the wallet file if the user does not specify the max_mixdepth.
Will code that later if noone sees a problem with that.

AlexCato:
Changed default in wallet-tool.py to -1 now. This will auto-detect the number of mixdepths only if the user does not explicitly configure it.

Therefor the default behavior is improved without any change in behavior if the user intentionally sets the parameter.

chris-belcher:
utACK on this. Anyone else have any thoughts?

AdamISZ:
Agreed. No conceivable negative I can see. Merging.

PR #594, opened 2016-07-25
fix crash on short/null privmsg

AdamISZ:
After reports via IRC: receiving privmsgs with 0 length can cause crash; also noticed it will also be possible if message after parsing trailers may cause a crash in on_privmsg in messagechannel. Added checks for this.


AdamISZ:
Going to merge directly to develop, as it's somewhat of a 'hot fix'. Can revert of course if there's a reason.

PR #596, opened 2016-07-29
Log testnet as well, to different file

nkuttler:
I would like to see testnet transactions logged as well. In this patch I use a property, but I guess setting the `statement_file` attribute in `__init__` would be considered cleaner? I can do it either way and update the other yigen scripts as well if this change is OK.

### DIFF
--- a/yield-generator-basic.py
+++ b/yield-generator-basic.py
@@ -40,7 +40,11 @@ log = get_log()
 # spent from utxos that try to make the highest balance even higher
 # so try to keep coins concentrated in one mixing depth
 class YieldGenerator(Maker):
-    statement_file = os.path.join('logs', 'yigen-statement.csv')
+    @property
+    def statement_file(self):
+        suffix = '-test' if get_network() == 'testnet' else ''
+        filename = 'yigen-statement{0}.csv'.format(suffix)
+        return os.path.join('logs', filename)
 
     def __init__(self, msgchan, wallet):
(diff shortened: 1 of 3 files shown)


chris-belcher:
I have nothing against this. In my mind the yield-gen log file's features are mostly replaced by wallet-tool.py's history method.

chris-belcher:
Before merging could you squish the commits into one.

nkuttler:
I haven't done the `yield-generator-oscillator.py` yet, the logging works a little differently. But feel free to merge anyway, I don't really understand what that one does and don't use it (yet).

AdamISZ:
> Ah, if it works, could you cherry-pick this into the 0.2.0 branch as well? Or should I make another PR?

There's already a small divergence; will just have to find the right way when the time comes. cherry-pick might be involved, not sure, would welcome advice.

AdamISZ:
Re the copy paste, I did a small refactor already in the 0.2.0 branch to avoid this (see `joinmarket/yieldgenerator.py`). Maybe we can just put this edit in there, and drop this PR?

nkuttler:
I only need to log testnet for the 0.2.0 branch, so yes, we can drop this pr and I can update the code. I've also started to use the oscillating yigen, I'll revisit that code as well.

AdamISZ:
OK, but as I say i'm hoping you only need to put it in the base yieldgenerator class after refactor. ("hoping" means I haven't looked at it, but I think so)

PR #605, opened 2016-08-26
Manage error 429 of blockr

the9ull:
When a wallet becomes too big the blockr service could raise this error on syns process:
{"status":"error","data":null,"code":429,"message":"Too many requests. Wait a bit..."}


AdamISZ:
This looks good, it would be nice to have a test (added to `test/test_blockr.py`), but that might not be practical given the nature of this functionality.

Code looks fine; if it's been tested locally, then ACK. Anyone else have a thought?

the9ull:
For the tests we could use a mock. I'll try to add it

AdamISZ:
Any progress on test? Anybody else have a thought?

the9ull:
I rewrote the code in order to cover all the calls to blockr.io and I added the test in test_blockr.py.

AdamISZ:
Looks very good, thanks.

PR #606, opened 2016-08-26
Mixdepth orders improvement

the9ull:
I found 2 issues:
1. if I change some parameters (e.g. DUST_THRESHOLD) the orders generator creates invalid orders. Now the sanity check deletes the invalid orders instead break all with an assert;
2. If more mixdepths have the same fee (e.g. `cjfee = ['0.0001', '0.0001', '0.0001', '0.00005', '0.00005']`) it is possible to join the orders with the same cjfee, improving the privacy


AdamISZ:
All ygs except -basic and -pe have now been removed (basically a rationalization). Awaiting whether someone wants to create a separate repo for them. Sorry about that. Closing.

PR #610, opened 2016-09-11
Add gaplimit option for yield generators.

domob1812:
Add a new option `-g` (as with `wallet-tool.py`) for the yield generators that allows controlling the gap limit used in the underlying wallet.

### DIFF
--- a/joinmarket/yieldgenerator.py
+++ b/joinmarket/yieldgenerator.py
@@ -77,7 +77,7 @@ class YieldGenerator(Maker):
 
 
 def ygmain(ygclass, txfee=1000, cjfee_a=200, cjfee_r=0.002, ordertype='reloffer',
-           nickserv_password='', minsize=100000, mix_levels=5):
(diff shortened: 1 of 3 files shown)


AdamISZ:
Please open pull requests against the develop branch. Thanks. Looks good (I'm actually a bit bewildered, didn't realise we didn't have that!).

domob1812:
Ah, thanks for the hint!  I've rebased now and changed this PR to be for `develop`.

I've been changing the gaplimit in the version I run for quite some time now (just hardcoded in the code), but I think it makes sense to have a proper option for that. ;)

AdamISZ:
Yes it does! Thanks .. :confused: (can't remember if we didn't have it or it was botched up somewhere). I'll take another look in a bit.

AdamISZ:
I'm not sure why the travis test didn't trigger, but it probably has something to do with the way you switched from master to develop. Can you re-create it as a new PR based off develop and close this one? I think that should fix it. I think I'll just merge it then.

domob1812:
Done in #613.  Closing this one.

PR #613, opened 2016-09-17
Add gaplimit option for yield generators.

domob1812:
Add a new option `-g` (as with `wallet-tool.py`) for the yield generators that allows controlling the gap limit used in the underlying wallet.

Recreated PR #610 as per the discussion there.

### DIFF
--- a/joinmarket/yieldgenerator.py
+++ b/joinmarket/yieldgenerator.py
@@ -77,7 +77,7 @@ class YieldGenerator(Maker):
 
 
 def ygmain(ygclass, txfee=1000, cjfee_a=200, cjfee_r=0.002, ordertype='reloffer',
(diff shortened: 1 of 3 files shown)


AdamISZ:
Yeah, it still doesn't trigger. No idea, oh well will just merge. If the build test fails can deal with it then.

AdamISZ:
Oh you opened against master again!

domob1812:
Oh sorry - seems the interface doesn't like me.  I selected "develop", then "compare across forks" and then my fork - but it seems that the "compare across forks" reset the base branch....  I'll do it again, sorry for the spam.

domob1812:
Ok, seems it worked in #614 now.

Issue #619, opened 2016-09-18
Additional IRC server suggestions.


qubenix:
Extending the discussion from here: https://www.reddit.com/r/joinmarket/comments/537uur/additional_irc_servers/.

OFTC:

+ Site: https://www.oftc.net
+ Open Tor Policy: https://www.oftc.net/Tor/
+ Contact: support@oftc.net
+ Github: https://github.com/oftc

CRYTO:

+ Site: http://cryto.net/irc/
+ Open Tor Policy: "TOR/VPN/proxy users are welcome on all servers, but abusive IPs will be banned."
+ Contact: `/msg joepie91` on irc.

AdamISZ:
I tried OFTC about 2-3 weeks ago, they (ironically) had shut down their Tor service a few weeks before that. I don't know if it's reinstated or not.

Other suggestions that have been mentioned: 

agora (cfyfz6afpgfeirst.onion/6667, agora.anarplex.net/14716) - operator is OK with us using, it auto-joins to \#agora which is an annoyance that'd have to be fixed, other than that it works. 

I tried volatile.ch but they strictly conform to the IRC RFC nick length restriction which screws up our new multi-message-channel code, I'm not sure how feasible a rewrite is, so putting that aside for now.

QuakeNet. Seems to support Tor but was told "they discourage it"or something by someone. Haven't tested or looked into it further.

CohibAA:
Slightly off topic, and maybe I missed something simple, but is there any way to determine from within JM the status of the IRC connections to the various hosts?  I've been out of the loop for awhile so I'm still not totally familiar with all the code changes, but I am trying to catch up and help out with the new version.

More to the topic at hand, I would like to test more host options and gather some type of data to determine connect-ability, reliability, etc for the different servers over a period of time.   I have been testing various servers, reviewing debug logs, and even made some adjustments to the debug logging to provide better information about which host's activity is being logged, but I am still having a hard time sorting out the info to see which servers are the best candidates for having multiple default connections.  Any ideas are appreciated.

Lastly, I'll submit a PR for the debug logging changes at some point soon, because I think it's useful, and the logs are not good for multiple IRC connections currently, unless someone has addressed this elsewhere.

AdamISZ:
> Slightly off topic, and maybe I missed something simple, but is there any way to determine from within JM the status of the IRC connections to the various hosts?

[...]

PR #622, opened 2016-09-24
part #agora if connected

CohibAA:
https://github.com/JoinMarket-Org/joinmarket/issues/619#issuecomment-249370294


AdamISZ:
Thanks!

(About the coverage thing, you probably already understood but it's the non-determinisms of the tests, some of which'll be hard to get rid of. It's still a useful tool even so, so not disabling it for now).

PR #623, opened 2016-09-25
add hostid to irc debug log

CohibAA:
add hostid to irc debug logging


AdamISZ:
Probably we can wrap up logging messages so this is done automatically, but no big deal, can be done when logging overall is improved.

PR #624, opened 2016-09-25
add agora-irc to config

CohibAA:
https://github.com/JoinMarket-Org/joinmarket/issues/619

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -94,19 +94,19 @@ rpc_user = bitcoin
 rpc_password = password
 
 [MESSAGING]
-host = irc.cyberguerrilla.org
-channel = joinmarket-pit
-port = 6697
-usessl = true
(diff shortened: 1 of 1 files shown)


AdamISZ:
ACK from me, although the server op reserved his right to change his mind, which is the only reason I'm not super-enthusiastic about using agora, just that, since it works today, and since he in principle gave permission, I think on balance we should give it a go.

@chris-belcher @adlai thoughts? Or anyone else?

AdamISZ:
On the other hand, this is (kind of) what develop is for; so merging as-is, and let's see if anything crops up as the number of participants grows.

AlexCato:
Running with these last 3 commits from @CohibAA  for the last 24 hrs, agora-irc seems to work just fine! (in combination with Rizon and CgAn). No errors, but also no coinjoins to report though.

PR #627, opened 2016-09-27
Autopep8 yg-pe.py

WyseNynja:
I forked the "develop" branch and then ran `autopep8 -a -a -a -a -i yg-pe.py` and then removed imports that pyflakes told me were unused.


AdamISZ:
Merging, the import removal in particular is a good tidy up. As for more general formatting/standardising I feel like we should hold off on that, perhaps include it in a broader refactoring (in which case we should make sure to decide what standards we're following hopefully). Thanks again.

PR #630, opened 2016-10-09
Add different log levels

AlexCato:
Currently, this is only intended for users to better understand which log messages are relevant to them.
Change all log.debug() lines in every .py file to the appropriate log level (log.debu(), log.info(), log.warn() or log.error()).
No changes in any program logic. DEBUG also still default log level.


AlexCato:
Hm, theres more commits than expected. Closing and trying again soon.

PR #631, opened 2016-10-09
Add different log levels

AlexCato:
Currently, this is only intended for users to better understand which log messages are relevant to them.
Change all log.debug() lines in every .py file to the appropriate log level (log.debug(), log.info(), log.warn() or log.error()).
No changes in any program logic. DEBUG is also still default log level for both the console and logfiles.

### DIFF
--- a/joinmarket/message_channel.py
+++ b/joinmarket/message_channel.py
@@ -81,8 +81,8 @@ class MessageChannelCollection(object):
                 #but should not kill the bot. So, we don't raise an
                 #exception, but rather allow sending to continue, which
                 #should usually result in tx completion just timing out.
-                log.debug("Couldn't find a route to send privmsg")
-                log.debug("For counterparty: " + str(cp))
+                log.warn("Couldn't find a route to send privmsg")
+                log.warn("For counterparty: " + str(cp))
         return func_wrapper
 
     def __init__(self, mchannels):
@@ -193,7 +193,7 @@ class MessageChannelCollection(object):
             time.sleep(1)
             i += 1
             if self.give_up:
-                log.debug("Shutting down all connections")
+                log.info("Shutting down all connections")
                 break
             #feature only used for testing:
             #deliberately shutdown a connection at a certain time.
@@ -255,7 +255,7 @@ class MessageChannelCollection(object):
             self.active_channels[nick].privmsg(nick, cmd, message)
             return
         else:
-            log.debug("Failed to send message to: " + str(nick) + \
+            log.warn("Failed to send message to: " + str(nick) + \
                           "; cannot find on any message channel.")
             return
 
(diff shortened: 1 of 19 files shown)


AlexCato:
As a next step, I think there should be an option in joinmarket.cfg to set the desired log level for console outputs (file logs should always use DEBUG, imho). It could default to the INFO level, unless explicitly changed.
This should be especially helpful for new users, who then do not get flooded with messages which are not relevant to them.

AdamISZ:
It's great to see someone kicked this off finally! :)

Now let's start bikeshedding :)

I had the same thought about the logfile. I'll try to make some comments tomorrow.

joinmarket/message_channel.py
```diff
@@ -255,7 +255,7 @@ def privmsg(self, nick, cmd, message, mc=None):
             self.active_channels[nick].privmsg(nick, cmd, message)
             return
         else:
-            log.debug("Failed to send message to: " + str(nick) + \
+            log.warn("Failed to send message to: " + str(nick) + \
```
AdamISZ: This one could happen very often e.g. if there is a spy connecting/disconnecting. I think info or debug would be fine, but I don't mind.

joinmarket/support.py
```diff
@@ -294,7 +294,7 @@ def choose_orders(db, cj_amount, n, chooseOrdersBy, ignored_makers=None):
     else:
         orders_fees = sorted(orders_fees, key=feekey)  #sort by ascending cjfee
 
-    log.debug('considered orders = \n' + '\n'.join([str(o) for o in orders_fees
+    log.info('considered orders = \n' + '\n'.join([str(o) for o in orders_fees
```
AdamISZ: I think keep the considered orders as debug? Seems like the class of stuff that is filling up the terminal a bit too much.

AdamISZ:
Shouldn't we add to this PR the code that filters out the logging that comes out to the terminal?

joinmarket/message_channel.py
AlexCato: Don't mind either way, assumed this doesnt happen often since I have not seen that message yet. I'd favor info.

joinmarket/support.py
AlexCato: Agreed.

AlexCato:
I've already worked on the code to filter the console messages, but from what i've researched it cant be done in the same part of the code, if it is supposed to be a config-file-option (guess you know this, but for reference: https://docs.python.org/2/howto/logging.html#logging-advanced-tutorial ). The PR I have in mind will make this more clear.
But I felt that the modified log messages can stand on their own, still improving user experience even if not filtered out by default. At least it lets them judge the severity of messages. Can add another commit to the PR with the log message filtering, if you guys want to see it here.

AlexCato:
Added the code now as well, to tweak logging to the console to INFO by default. Added this as a config option to joinmarket.cfg.
As usual there's some fights of git branches vs. me, so there's one more commit than necessary. The code for the logging changes is only in the files support.py and configure.py

Reasoning for moving the console logging initialization from support.py to configure.py:
Before my changes, the console logging was initialized before the configuration was loaded. Accoding to my 'research', it's not possible to change the log level of an already attached loghandler. So it needs to be initialized correctly the first time. That's why I moved it.

I've tested this code with YG-PE and sendpayment. Worked just fine.

PR #636, opened 2016-10-17
add Dockerfile and entrypoint.py

WyseNynja:
This isn't ready for merging yet but I am done working on it for tonight and wanted to see what you all thought.

I also have a script to setup a Tor and bitcoind container that is mostly done but I'm waiting for my node to sync to make sure everything works there.

With Docker installed, a single command gets you a running yield generator or tumbler. Although so far I've only tested it with the orderbook watcher.

I can switch the FROM to use upstream ubuntu (or something smaller like alpine), but at least for developing this branch, it is nice to have my scripts.

Let me know what you think.


AdamISZ:
Seems like a great direction to go in :) There were various discussions about this *kind* of thing, it seems like a good idea; but personally I've never even investigated Docker, specifically, so not exactly qualified to comment on the details. Hopefully someone else can take a look.

WyseNynja:
I have a partially complete rocker-compose file that will launch Tor, then Bitcoin, then an ob_watcher, and then a configurable number of wallets. It isn't ready yet, but you can see what I have so far at https://github.com/WyseNynja/joinmarket-suite

WyseNynja:
This still isn't ready yet, but the local tests I've had time to do so far are going really well. I just need to make some changes to how /etc/hosts is setup and I think this will be ready.

PR #637, opened 2016-10-17
add direct send feature from mixdepth using -N 0 in sendpayment scrip…

AdamISZ:
…t; includes sweep

Done a fair amount of checking/testing on regtest (there is an extra test function included in the commit also).

Basically just use -N 0 and the rest of the syntax is as usual for sendpayment. Prompts with transaction raw/serialized before you choose to push.

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -32,6 +32,63 @@ def check_high_fee(total_fee_pc):
         print('WARNING   ' * 6)
         print('\n'.join(['=' * 60] * 3))
 
+def direct_send(wallet, amount, mixdepth, destaddr, answeryes=False):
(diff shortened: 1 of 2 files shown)


AdamISZ:
Tested OK on realnet with 1 transaction, I might do another in a bit; made a small tweak to show the txid clearly. I think it has appropriate info coming out on the terminal and the sanity checks are sufficient. I'm happy with it, let me know if you see anything.

AdamISZ:
Tested again, OK (this time sweep). Unfortunately only I have tested but have tested (sweep, non-sweep) across regtest, testnet, mainnet quite a bit now, so going to merge.

PR #638, opened 2016-10-18
ignore .DS_Store

WyseNynja:
These are annoying on macs.


AdamISZ:
not worth bothering now, but please squash in future if it's easy

Issue #632, opened 2016-10-11
the yield generator should have by default a random fee


greenaddress:
To make it harder to track things on the network it seems to make sense to use a random fee (within a range/max value?)

AlexCato:
Can easily be done, though arguably, as long as there's others not changing the default, individual yieldgens are still hard to be distinguished by the fee alone. But it might add one more layer, so it's possibly a slight improvement.
Could offer a PR to randomize the default relative cj fee to be between 0.018% and 0.022% (currently: fixed 0.02% ) or something along those lines?

AdamISZ:
Yeah except this is to be added to a reduction in the default imo. Obv people like to set their own prices (that's the point) but having a *default* tweak I thought, on balance after discussion with @greenaddress was a sound idea. People are lazy and run with the same fee all the time, mostly. Well, at least with the default scripts.

juscamarena:
Wouldn't it be better to have a random fee set in a specific range after every cj or every x cjs?

AlexCato:
That would complicate things, as you'd have to announce changed orders in public whenever the cj fee changes.
The current yield generator's goal is to announce his offerings in public as rarely as possible to avoid some information leakage about coinjoins (thats why it touches the maximum mixdepth only if absolutely necessary, because this leads to a changed maximum offer size). Could be changed only if an new announcement had to be made anyways to avoid this problem.

But I dont see what's gained by changing this as well: a snoop would monitor the channel 24/7 anyways, and random cj fees dont help at all with that, as your bot's nickname stays the same and offers can therefor easily be linked just by that.

AdamISZ:
I thought about this more carefully. TLDR I think we should strongly enforce a different fee value each time (of course "enforce" is not possible, but you get it).

First let's be clear that there's 2 different threat models; one is real time channel monitoring (which admittedly is easy to do), and the other is blockchain-only analysis. 

Now, consider the latter - suppose after the event, blockchain-only analysis. Considering one tx only, the size of the fee doesn't really matter, since the inputs and change outputs can (almost) always be linked using subset-sum analysis, while the coinjoin outs cannot, no matter what the fee is. But a single entity always using a very unique fee, let's say 0.0618%, gives a watermark to his own (inputs, change) set for each tx he takes part in, assuming that within some tolerance no one else is near to that value, which will effectively link his transactions that send to different mixdepths (by our design this is what we're trying to hide).

[...]

PR #640, opened 2016-10-19
add yg-randomizer: YG to randomize all aspects of cj offers

AlexCato:
Randomization of all aspects of coinjoin offers after every cj to make blockchain analysis harder. Code based on yg-pe.py with changes only in user configuration and the method create_my_orders(self).

This could have been made easier with normal distributions, letting users configure mu and sigma. But for less "mathy" people, this way to configure it is easier to understand imho.

Adam mentioned that more YGs probably should get their own repo. It's up to you, however you prefer this.

Idea stolen from #632 

Edit: updated the PR with the newly suggested defaults in #641

### DIFF
--- a/yg-randomizer.py
+++ b/yg-randomizer.py
@@ -1,0 +1,208 @@
+#! /usr/bin/env python
+from __future__ import print_function
+
+import datetime
+import time
+import random
+
+from joinmarket import jm_single
+from joinmarket import get_log, calc_cj_fee
+from joinmarket import YieldGenerator, ygmain
+
+# User settings
+
+ordertype = 'reloffer'  # [string, 'reloffer' or 'absoffer'] / which fee type to actually use
+cjfee_a = 5000          # [satoshis, any integer] / absolute offer fee you wish to receive for coinjoins (cj)
+cjfee_r = '0.0002'      # [percent, any str between 0-1] / relative offer fee you wish to receive based on a cj's amount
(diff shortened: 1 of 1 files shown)


raedah:
I will just say that personally Id like to see the market price feedback behavior come back a la oscillator too.

AdamISZ:
I'd hoped people would maintain custom scripts separately.

As an experiment, today, I'll set up a separate repo in the Org and try to give @AlexCato  and @raedah commit permissions, then you could add them in there. The other alternative is to have people just maintain their own, separate repos in their own github accounts, perhaps.

Then there would be a pointer in the README saying basically "here's 2 (or better, just 1) default yg, feel free to use one of the custom ones in <joinmarket-org/custom-scripts>, see <link> here for details".

What I'm trying to avoid is having to review and test (even crude tests take a while, and the tests aren't really good enough on their own) the several different options that we want people to have. Anything in the main repo has to be signed off by one of us three; recently that has just been me, realistically, and so my time is very squeezed. There are huge issues to address outside of the choice of yield generator algo, so I have to focus on those.

If we create a new repo for this, albeit it's clunky, it avoids that issue if other people can sign off on them, review them, and do the bugfixing for them.

AdamISZ:
@AlexCato  @raedah I sent a collaborator invite, let me know if it doesn't get to you

AlexCato:
Thanks @AdamISZ . I've added this to the new custom-scripts repo instead ( https://github.com/JoinMarket-Org/custom-scripts ). Closing this.

PR #641, opened 2016-10-19
Modify default fee to 0.02%

AdamISZ:
Modify default maker count to rand(3,5) from (4,6)
Modify default maker_timeout_sec to 45 from 30 (bigger messages)

Rationale: default relative fee is too high, in tumbler is likely to result in fees ~ 4 (num mixdepths) x 5 (num makers) x .2% which is 4% before bitcoin tx fees which can easily be another percent, for smallish amounts. This is clearly too high (although it's an overestimate, but still). Market experience suggests fees rapidly fall to much lower values.

Note that with default txfee=1000 (admittedly this is almost completely arbitrary nowadays until a new tx fee contribution model is devised), we get a minsize with these defaults of ~ 7 million satoshi; although this is fairly large I am not unhappy with dissuading any trades < 5 million since at that point the bitcoin fee starts to eat a large chunk (e.g. 100k satoshi being 2% at that point). (Note that a commitment is not used if your transaction doesn't fit that condition, it's a filter in the order selection.)

Slight reduction from (4,6) to (3,5) for default makercount: more makers is not only more expensive, but more likely to fail. In my experience doing "lightweight" coinjoin, e.g. just retail sendpayment, I am not particularly inclined to pay very large transaction fees by using 6+ makers, and I know it helps the chance of a quickly effective tx if I choose less (also of course picking with -P). This one is of course very arguable, certainly less is going to be worse for privacy, I just think 3-5 is a more reasonable tradeoff.

Increase in maker_timeout_sec from 30 to 45: this is just a reflection of the observation that throttling is much more likely to be triggered now even before the big !tx messages, in particular in !auth. Haven't worked out the details but it seems wise to bump it somewhat. The "default if not in file" of 60 is of course even more conservative (unchanged here).

I realise that there is plenty of room for bikeshedding here; if you have a strong opinion feel free to voice it, but otherwise will merge after a few days.

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -116,7 +116,7 @@ socks5_port = 9050, 9050
 console_log_level = INFO
 
 [TIMEOUT]
-maker_timeout_sec = 30
+maker_timeout_sec = 45
 unconfirm_timeout_sec = 90
 confirm_timeout_hours = 6
 
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -283,9 +283,9 @@ def main():
                       type='int',
                       dest='makercount',
                       help='how many makers to coinjoin with, default random '
-                           'from 4 to 6; use 0 to send *direct* to a destination '
+                           'from 3 to 5; use 0 to send *direct* to a destination '
                            'address, not using Joinmarket',
-                      default=random.randint(4, 6))
+                      default=random.randint(3, 5))
     parser.add_option(
         '-C',
         '--choose-cheapest',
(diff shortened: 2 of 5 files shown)


WyseNynja:
utACK

I wish we had data on failure rates and tx sizes and privacy gained for different maker counts so we could be sure these lower maker counts were worth it.

I also wonder if we should change the absolute fee.

AdamISZ:
Rebasing and merging in the absence of other comments. We can always change it back :)

PR #643, opened 2016-10-21
Allow notifies coming from external hosts

paoloaga:
Handy to use on a computer that doesn't host the bitcoind node (like a RPi).

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -526,7 +526,7 @@ class BitcoinCoreNotifyThread(threading.Thread):
         self.btcinterface = btcinterface
 
     def run(self):
-        notify_host = 'localhost'
+        notify_host = '0.0.0.0'
         notify_port = 62602  # defaults
         config = jm_single().config
         if 'notify_host' in config.options("BLOCKCHAIN"):


AdamISZ:
All PRs against develop branch please.

paoloaga:
Ok, next time I will push there. Anyway this is just a very brief patch (just a string in the code). Not sure if it's worth it to make another PR. If you think it's valid, just insert it in the next release. Thanks.

chris-belcher:
It should be possible to change the PR destination without closing it and opening a new one.

paoloaga:
I found the feature, thanks. Is it all right now?

AdamISZ:
OK, cool, @chris-belcher do you ACK? (sorry for trivial question but I know you actually thought about this, I didn't :)) If you merge please use github-merge.py to sign the merge commit, else I can do it.

adlai:
utACK

This does make scripts slightly more vulnerable to grinding induced by other computers, but I think that people running JM with ports forwarded should know what they're doing. Also, I don't see that this creates dangers beyond merely consuming extra resources, in which case the operator could stop forwarding ports and use an SSH tunnel instead.

Further thoughts, @chris-belcher ?

chris-belcher:
Yes I think its good. utACK.

About a year ago we had a vulnerability which involved passing strings from the http server to system(). Since then that part was changed to use urllib which doesn't pose such a serious security risk.

adlai:
I'm amending my position to NACK, because this does open up more avenues for grinding or even worse vulnerabilities. Off the top of my head I can't think of how to exploit this, but why make life easier for enemies?

I'll ACK a version of this which lets the user configure the binding host, but leaves the default as localhost (or 127.0.0.1, which I believe has the same effect).

PR #647, opened 2016-10-24
initial draft allowing takers to optionally complete transactions wit…

AdamISZ:
…h less than the initially requested number of makers; sweep function is unchanged

It occurred to me that this is a much simpler, faster, and possibly more robust way of dealing with unresponsive makers who don't even provide utxos or signature up to !ioauth.

This is not really intended to supercede the idea of more intelligent selection, something like outlined [here](https://gist.github.com/AdamISZ/79457f1c20bc0702d299d873c899676a) but is a more basic idea that is simpler.
 
If the number of full respondants in the `!fill` - `!ioauth` phase is less than that specified (by -N or similar), but is at least equal to `jm_single().config.getint("POLICY", "minimum_makers")`, then continue anyway - the change will be calculated correctly, higher than originally anticipated, and the txfee will be less than originally anticipated but it is already re-calculated at this point.

The timeout code in `recover_from_nonrespondants` thus checks the above arithmetic, if sufficient, it calls `recv_txio` with dummy values, which flags the code to skip the first step and pass straight into calculation of our change output and construction of `!tx` messages (nothing for the makers changes; their outputs are already set by their own fees).

Nothing changes for sweep; there was previously no way to handle any change in the input set of orders, and there still isn't (at least I haven't thought of one yet).

I've done various tests with randomly malicious makers (using `test/ygrunner.py` and inserted random failure-to-ioauth into `maker.py`) and it seems to work as intended.

User can flag not to use this feature by setting `POLICY.minimum_makers` to 0.

Review greatly welcomed. This needs careful testing, even if it is a fairly simple change (at least in comparison to other ways of dealing with the problem!)

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -126,6 +126,11 @@ confirm_timeout_hours = 6
 # for most rapid dust sweeping, try merge_algorithm = greediest
 # but don't forget to bump your miner fees!
 merge_algorithm = default
+# For takers: the minimum number of makers you allow in a transaction
+# to complete, accounting for the fact that some makers might not be
+# responsive. Should be an integer >=2 for privacy, or set to 0 if you
+# want to disallow any reduction from your chosen number of makers.
+minimum_makers = 2
 # the fee estimate is based on a projection of how many satoshis
 # per kB are needed to get in one of the next N blocks, N set here
 # as the value of 'tx_fees'. This estimate is high if you set N=1, 
--- a/joinmarket/taker.py
+++ b/joinmarket/taker.py
@@ -147,58 +147,65 @@ class CoinJoinTX(object):
         return True
 
     def recv_txio(self, nick, utxo_list, auth_pub, cj_addr, change_addr):
-        if nick not in self.nonrespondants:
-            log.debug(('recv_txio => nick={} not in '
-                       'nonrespondants {}').format(nick, self.nonrespondants))
-            return
-        self.utxos[nick] = utxo_list
-        utxo_data = jm_single().bc_interface.query_utxo_set(self.utxos[nick])
-        if None in utxo_data:
-            log.error(('ERROR outputs unconfirmed or already spent. '
-                       'utxo_data={}').format(pprint.pformat(utxo_data)))
-            # when internal reviewing of makers is created, add it here to
-            # immediately quit; currently, the timeout thread suffices.
-            return
-        #Complete maker authorization:
-        #Extract the address fields from the utxos
-        #Construct the Bitcoin address for the auth_pub field
-        #Ensure that at least one address from utxos corresponds.
-        input_addresses = [d['address'] for d in utxo_data]
-        auth_address = btc.pubkey_to_address(auth_pub, get_p2pk_vbyte())
-        if not auth_address in input_addresses:
-            log.error("ERROR maker's authorising pubkey is not included "
-                      "in the transaction: " + str(auth_address))
-            return
+        if nick:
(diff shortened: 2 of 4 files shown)


AdamISZ:
Tried on main net, chose 2 known malicious out of 6 using -P, got a 4 party tx completed.

joinmarket/configure.py
```diff
@@ -126,6 +127,11 @@ def jm_single():
 # for most rapid dust sweeping, try merge_algorithm = greediest
 # but don't forget to bump your miner fees!
 merge_algorithm = default
+# For takers: the minimum number of makers you allow in a transaction
+# to complete, accounting for the fact that some makers might not be
+# responsive. Should be an integer >=2 for privacy, or set to 0 if you
+# want to disallow any reduction from your chosen number of makers.
+minimum_makers = 2
```
AlexCato: I also believe this should be opt-out. "First user" experience is important, also default-settings should just work. Which they currently dont, due to these unresponsive bots.

AlexCato: Recommendation for user though: at the point where they confirm that they will pay 0.xx % of cjfees in total, there should be a note if minimum_makers > 0. Something like:
If some makers do not respond, we will still create a coinjoin with at least (minimum_makers) makers. If you do not desire this behavior, you can set "minimum_makers = 0" in joinmarket.cfg.

AdamISZ: @AlexCato I agree. Will push and squash that edit shortly.

AdamISZ:
Additional notes as of 4fe8e25 (will squash before merge, don't want to bork comments for now)

This commit does 3 things:
1. Trivial, remove `CoinJoinTx.actual_respondants` which was a holdover from working on another branch, it is completely redundant with the existing `active_orders.keys()` here.
2. I had removed the existing smart order-reselection on initial failure for the case `minimum_makers=0` which was unnecessary; that code is reinstated in the block `elif minmakers == 0` in `recover_from_nonrespondants()`. Thus, the code leaves functionality unchanged if people disable this option.
3. Added an informative message as per @AlexCato 's suggestion, in `sendpayment.py` so that if choose-from-subset is enabled, the user is warned that the maker count can go down.

I did a bunch of tests of cases: (1) minmakers 0, no malicious bots, (2) minmakers 0, randomly malicious bots, (3) minmakers 2, no malicious bots, (4) minmakers 2, malicious bots and (5) sweeps with malicious bots (minmaker setting not relevant).

The tests seem to behave as expected. I'll run a couple more tumbler tests today, this time including malicious behaviour in bots to see if it handles it OK.

AdamISZ:
Tumbler tests with inserted maker maliciousness seem to be working fine. I'd anticipate merging this in the next few days - and making a new release - unless anyone objects; additional tests by others hugely appreciated.

joinmarket/taker.py
```diff
@@ -405,43 +412,59 @@ def self_sign_and_push(self):
...
-            #Re-source commitment; previous attempt will have been blacklisted
-            self.get_commitment(self.input_utxos, self.cj_amount)
-            self.msgchan.fill_orders(new_orders, self.cj_amount,
-                                     self.kp.hex_pk(), self.commitment)
+                #Re-source commitment; previous attempt will have been blacklisted
+                self.get_commitment(self.input_utxos, self.cj_amount)
+                self.msgchan.fill_orders(new_orders, self.cj_amount,
+                                         self.kp.hex_pk(), self.commitment)
+            else:
+                log.info("Two few makers responded to complete, trying again.")
```
adlai: s/Two/Too/

Issue #470, opened 2016-03-16
Bloom filter blockchain interface


chris-belcher:
Fleshing out this idea in more detail: https://github.com/JoinMarket-Org/joinmarket/issues/55#issuecomment-165994649

The user would give a single ip address and host to JoinMarket that would connect to a network node controlled by the user and synchronize it's wallet using BitcoinJ-style bloom filters.
This would have the same security, privacy and trustless properties as running a full node on the same machine but would solve other problems too. Such as:

1. Tails users find it very difficult to sync a full node over tor. This feature would allow them to point their JoinMarket to a .onion full node controlled by them either in their machine or a VPS.

2. Full nodes are very annoying to start up and shut down. After having been switched off even overnight it takes an irritating amount of time to sync to the network again and be able to make transactions. With this feature a user could have a full node permanently on in their own home (on a raspberry pi?) and point their JM to it over LAN.

Two problems/issues I can think of.

1. The bitcoin p2p protocol doesn't expose the estimatefee function.

2. Care must be taken when broadcasting transactions so attackers that connect to every single node on the network don't see the IP address of the full node.

However a better alternative to all this may be to simply tunnel JSON-RPC over ssh or tor and get the same thing.

chris-belcher:
Third benefit: user experience with sync'ing is much better with bloom filters.
The user can download blocks quickly containing only their transactions, the script can increment the address index with respect to the gap limit, so only one pass is needed.

Fact is, if there's something Mike Hearn knew about that was user experience.

chris-belcher:
Here is some code I was playing with that could be helpful to anyone working on this

https://gist.github.com/chris-belcher/493551782452ff7cf6336dec760e1f1e
https://gist.github.com/chris-belcher/1f1fee0b39f023d7bd04dc6e5b722359

The p2p networking code should be somewhat modular so it can be reused for #503 and https://github.com/JoinMarket-Org/joinmarket/issues/55#issuecomment-184165679

chris-belcher:
This and #653 have the serious problem that they can't query the UTXO set, which is required to be a taker.

In recv_txio() and add_signature() any arbitrary UTXO that the makers send must be queried for their script and amount. In make_commitment() only the UTXOs already in the taker's own wallet must be queried, in fact that code could be modified to use wallet.unspent instead of query_utxo_set().

[...]

Issue #55, opened 2015-03-11
SPV blockchain interface.


chris-belcher:
Setting up bitcoind is annoying and I bet many people wont do it. Using blockr.io is a poor alternative.

A good idea might be to use BitcoinJ's SPV verification to obtain the transaction history and UTXOs without privacy violations.

chris-belcher:
Actually.. https://jonasnick.github.io/blog/2015/02/12/privacy-in-bitcoinj/

BitcoinJ is not a solution to our privacy problem.

chris-belcher:
From reading around a bit, it seems there is not solution for a private blockchain interface that isnt running a full node.

Having people rely on blockr.io is still unacceptable, perhaps a small workaround could be to implement many different web blockchain explorers and have each yield generator choose between them

chris-belcher:
What if there was some way to prove you had a local copy of the blockchain and were not using a blockchain explorer?

chris-belcher:
If issue https://github.com/chris-belcher/joinmarket/issues/57 works, this problem may end up being solved by economics and incentives instead of technology. And increase the number of full nodes on the network to boot!

chris-belcher:
insight is better than blockr.io as in #64 as there are many services running it so it would be more decentralized

chris-belcher:
Connecting to Electrum servers could be a good idea.

chris-belcher:
Synchronizing tx and address data with privacy is an area of active research http://lists.linuxfoundation.org/pipermail/bitcoin-dev/2015-July/009510.html

chris-belcher:
Another idea could be to use BitcoinJ style SPV with an option to specify the IP address of the node you want to connect to. This would allow users to connect to only nodes they control and trust.

chris-belcher:
Yet another way is to create a client with SPV security (trusting the miners) which does not use bloom filters. In other words it downloads every block from the wallet creation date onwards and searches it for its own addresses. This would give SPV security but almost-full privacy (only leaking the approximate wallet creation date)

nopara73:
I've gathered some information on BIP37 discussions.  
BIP37, bloom filers discussion privacy history (not chronological):  
  
1. BIP37 - Mike Hearn, Matt Corallo: https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki  
2. Privacy in BitcoinJ - Nick Jonas: https://jonasnick.github.io/blog/2015/02/12/privacy-in-bitcoinj/  
[...]

PR #652, opened 2016-11-08
Adds ability to change encryption password of wallet.json

### LINKED ISSUE #651: Add ability to change encryption password of wallet.json
The encryption password can't be changed.
It wouldn't be too hard to add an option to wallet-tool.py which lets it be changed.

juscamarena:
This resolves #651
Suggestions welcome.
TODO: Add tests

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -20,7 +20,8 @@ description = (
     'balances. (displayall) Shows ALL addresses and balances. '
     '(summary) Shows a summary of mixing depth balances. (generate) '
     'Generates a new wallet. (recover) Recovers a wallet from the 12 '
-    'word recovery seed. (showutxos) Shows all utxos in the wallet, '
+    'word recovery seed. (changepassphrase) Change encyption password '
+    'for specified wallet (showutxos) Shows all utxos in the wallet, '
     'including the corresponding private keys if -p is chosen; the '
     'data is also written to a file "walletname.json.utxos" if the '
     'option -u is chosen (so be careful about private keys). '
@@ -89,7 +90,7 @@ noseed_methods = ['generate', 'recover', 'listwallets']
 methods = ['display', 'displayall', 'summary', 'showseed', 'importprivkey',
     'history', 'showutxos']
 methods.extend(noseed_methods)
-noscan_methods = ['showseed', 'importprivkey', 'dumpprivkey']
+noscan_methods = ['changepassphrase', 'showseed', 'importprivkey', 'dumpprivkey']
 
 if len(args) < 1:
     parser.error('Needs a wallet file or method')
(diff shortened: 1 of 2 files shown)


wallet-tool.py
```diff
@@ -229,6 +230,35 @@ def cus_print(s):
...
+    password_key = btc.bin_dbl_sha256(password)
+    encrypted_seed = encryptData(password_key, seed.decode('hex'))
+    timestamp = datetime.datetime.now().strftime("%Y/%m/%d %H:%M:%S")
+    walletname = args[0]
+    walletpath = os.path.join('wallets', walletname)
+    fd = open(walletpath, 'r')
+    walletfile = fd.read()
+    fd.close()
+    walletdata = json.loads(walletfile)
+    walletdata['creation_time'] = timestamp
```
AdamISZ: Probably better not to change creation_time.

AdamISZ:
tACK (just basic check).

I think this is fine (apart from the minor nit above). It would be better to factor out things like the password input and probably the wallet opening code, but no reason for that to block here (that kind of refactoring is needed in a few places).

Re: tests: the tests in test/test_wallets.py were written ages ago (I started out using class-based tests but that isn't necessary with the py.test approach we switched to, just functions is fine). Although the tests for wallets are quite crappy and neglected, the basic idea of using pexpect as [here](https://github.com/JoinMarket-Org/joinmarket/blob/master/test/test_wallets.py#L225-L231) should be fine for creating a test where you run the `changepassphrase` command, and then check that the wallet still opens. What's really needed there (and same for existing tests) is to have a sanity check of balance after applying the command. If I had time I would improve it myself.

wallet-tool.py
chris-belcher: Yes, that wallet creation time one day needs to be used for #470 and https://github.com/JoinMarket-Org/joinmarket/issues/55#issuecomment-184165679

juscamarena: Ah, good to know! I can change this.

juscamarena:
I wanted to touch as little of the rest of the code as I could for a first time pull request, but yeah I agree with  factoring most of the wallet related io out.

chris-belcher:
Looks great to me.

Squish these two commits together (here's a tutorial in case it helps http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

And also add tests. I can help you on IRC since I remember getting them to run is a little tricky. There's also this tutorial on the wiki https://github.com/JoinMarket-Org/joinmarket/wiki/Testing

chris-belcher:
Shall we merge this? (Reminder to self that the merge.py script needs to be used so every commit is signed)

juscamarena:
I had the test working and passing with the correct outputs being checked with logs.. When I run them locally now I'm getting weird issues with bitcoind in regtest requesting wallet history instead of executing my changepassphrase function.. I'm stumped.

AdamISZ:
The test is expecting ": passphrase has been updated" but that is not printed on the console, and I can't find it in the code? Was it something you changed/removed perhaps?

juscamarena:
Okay, now I think it's ready to merged.

PR #654, opened 2016-11-12
Fix install deps

WyseNynja:
Inspired by https://www.reddit.com/r/joinmarket/comments/5cgy40/pyconfigh_no_such_file_or_directory/

This is a quick change to match that. However, for my own install on a fresh system, I only need:
 - gcc
 - libsodium13
 - python-dev
 - virtualenv

I highly recommend that we stop recommending installing to the system python and always recommend using virtualenv instead. Its just too easy to break your system install with pip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ The installation is slightly changed, with the secp256k1 python binding no longe
     ```
 
 + Install Python dependencies:
-    You need `pip` (e.g. `sudo apt-get install python-pip`, recent Python usually ships with pip).
+    You need `pip` >= 1.5.6 (e.g. `sudo apt-get install python-pip`, recent Python usually ships with pip).
 
     (Setuptools version must be >= 3.3).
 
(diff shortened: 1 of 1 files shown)


OverlordQ:
Probably should keep both python and python-dev since I'm pretty sure there's not a dependency between the two

OverlordQ:
I lied, there is from dev package to base.

AdamISZ:
Slight annoyance, sorry, but anduck just put out a PR to fix "apt-get python-pip" to "apt-get install ..", pls rebase. Re the virtualenv, I was slightly cautious about pushing people to use virtualenv as it's another hoop for an "untrained" user, but I do agree to be honest. So I appreciate these changes to the instructions.

WyseNynja:
I definitely think that we should promote the use of virtualenvs. I've seen too many systems broken by pip installing to the system. We should probably do that in a separate (larger) PR.

Issue #653, opened 2016-11-10
Full block SPV sync


chris-belcher:
Issue for this https://github.com/JoinMarket-Org/joinmarket/issues/55#issuecomment-184165679

> Yet another way is to create a client with SPV security (trusting the miners) which does not use bloom filters. In other words it downloads every block from the wallet creation date onwards and searches it for its own addresses. This would give SPV security but almost-full privacy (only leaking the approximate wallet creation date)

chris-belcher:
This would be a great default blockchain interface. If your wallet creation date is _now_ then no blocks have to be downloaded. Also it doesn't require any configuration by the user.

chris-belcher:
This idea had the problem that takers wouldn't have access to the UTXO set, so makers could send them fake UTXOs to DOS.

Here is a possible solution to that from gmaxwell.

```
<gmaxwell> belcher: exactly what is needed in these queries? can we make a public API that gives the required data without trashing privacy?
<belcher> not really, it needs a mapping from address to utxo and history
<belcher> im planning to write something where the user points joinmarket to their own full node and it sync's from there, which would be private
<gmaxwell> What does history mean precisely? 
<belcher> whether an address was used on the blockchain before
<gmaxwell> and it needs this only as of the current height, right?
<belcher> yes
<gmaxwell> so, if I make a table of 128 bit hashes of txid|vout|scriptpubkey|amount|used_flag  for the whole UTXO set the database would be 738 MB, maybe half that size if dust addresses were filtered. and maybe 10x smaller if you just instead eliminate all reuse (which I suspect you might be able to ban?).  
<gmaxwell> I think it would not be completely unreasonable to have a thing where users just download the whole thing, check its hash against varrious observers and get diffs-- especially if you could eliminate dust and reuse.  Alternatively, it wouldn't be completely unrealistic for people to have PIR servers that can query this data.
<belcher> so that i understand, is that hash(txid|vout|scriptpubkey|amount|used_flag), and then if you know a UTXO you can check whether it exists?
<gmaxwell> Yes, so someone you're trying to join with would need to already know their UTXO, and you could query this to tell if it exists.  If you can't do that then you need a real database and not a set-- which could be done but would be a couple times bigger-- not out of the question, but getting less attractive vs a full node.
<belcher> yes this scheme would help certainly
[...]

Issue #56, opened 2015-03-13
Against bitcoin p2p network sybil attackers.


chris-belcher:
Regarding https://www.reddit.com/r/Bitcoin/comments/2yvy6b/a_regulatory_compliance_service_is_sybil/

Obviously coinjoin doesn't directly help against this IP tracking. But this project can still help.

We should introduce a method where the taker can send the fully-signed tx hex to one of the makers who will then broadcast it. The taker would choose to broadcast the txhex himself, or send it to one randomly chosen maker to broadcast. In this way the coinjoin tx could come from many IPs instead of just the taker's.

Naturally this kind of bad actor could also be a sybil in the joinmarket. They will be limited in ways already discussed: To coinjoin everybody's transactions the sybil would need to own a large amount of bitcoins. Also their coinjoining activity will stop other sybils learning what's happening. Because of the feature in #14 the sybil will not be guaranteed to coinjoin with everyone just because they offer the lowest price.

It's a partial solution.

AdamISZ:
> We should introduce a method where the taker (coinjoin-initiator) can send the fully-signed tx hex to one of the makers (market-makers who wait-to-coinjoin) who will then broadcast it.

Agreed, that can only help. Of course, network level obfuscation *might* be another step up, e.g. Tor, but that's another can of worms, and a bit outside the scope here.

chris-belcher:
What does network level obfuscation mean in practice?

Regarding the edit. True.
The taker has to be in a race to claim the UTXO inputs, otherwise malicious takers could negotiate many coinjoins with makers and never broadcast them to use up all the maker's UTXOs.
The person controlling the IRC server will be easily able to make that link. Hopefully it shouldn't be an issue for another messaging channel like subspace.

chris-belcher:
Wrote the code for the maker side. https://github.com/chris-belcher/joinmarket/commit/919e97fd2d58a9758f60fec350fd6888210945f9

Disabled on taker side for now.

chris-belcher:
There could be a script called pushtx.py which connects to the network and sends a tx to a maker, either chosen by random or chosen by the user.

A similar kind of thing can be in the ob-watcher.py html page

chris-belcher:
Could requesting a maker to push lots and lots of transactions could get that maker banned by other bitcoin nodes? This may be a DOS opportunity.

chris-belcher:
An example of a paper that actually does this: https://arxiv.org/pdf/1612.06747v3.pdf

[...]

PR #662, opened 2016-11-19
Tor broadcast method

chris-belcher:
Use `tx_broadcast = tor` in joinmarket.cfg to connect to a random bitcoin node on the network, to broadcast a transaction and then disconnect. The DNS seeds are used to find these nodes.

Right now on the bitcoin network, there are an unknown number of sybil nodes which aggressively announce themselves in an effort to attract more people to connect to them. Then they spy on every newly-broadcasted transaction, tracking it as it propagates through the p2p network and giving them a good idea of the IP address that originally broadcasted that transaction.

JoinMarket already had a feature of asking a maker bot to broadcast your transaction, although this isn't as good as using tor. See issue #56 for discussion.

This PR implements #503. It includes code for connecting to a bitcoin node over p2p network which can be re-used for other features like #470 and #653.

The test uses the new `generatetoaddress` RPC call from 0.13.0 and hopefully works quite well.

I haven't tested on mainnet.

### DIFF
--- a/joinmarket/peertopeer.py
+++ b/joinmarket/peertopeer.py
@@ -1,0 +1,464 @@
+#! /usr/bin/env python
+from __future__ import absolute_import, print_function
+
+import socket, time, random, sys
+from struct import pack, unpack
+from datetime import datetime
+
+from joinmarket.configure import load_program_config, get_network
+from joinmarket.socks import socksocket, setdefaultproxy, PROXY_TYPE_SOCKS5
+from joinmarket.support import get_log
+
+import bitcoin as btc
+log = get_log()
+
+PROTOCOL_VERSION = 70012
+DEFAULT_USER_AGENT = '/JoinMarket:0.2.3/'
+
+##protocol versions above this also send a relay boolean
+RELAY_TX_VERSION = 70001
+
+##length of bitcoin p2p packets
+HEADER_LENGTH = 24
+
+##how many times to connect to peer before giving up
+MAX_CONNECTION_ATTEMPTS = 10
+
+##if no message has been seen for this many seconds, send a ping
+KEEPALIVE_INTERVAL = 2 * 60
+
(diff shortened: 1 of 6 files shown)


joinmarket/peertopeer.py
```diff
@@ -0,0 +1,307 @@
...
+        self.socks5_hostport = socks5_hostport
+        if not self.testnet:
+            self.magic = 0xd9b4bef9 #mainnet
+        else:
+            if testnet == True:
+                self.magic = 0x0709110b #testnet
+            else:
+                self.magic = 0xdab5bffa #regtest
+        self.closed = False
+        self.connection_attempts = 4
```
chris-belcher: Increase this number to about 20 or 30 since 4 is far too low, also make it a defined constant MAX_CONNECTION_ATTEMPTS

chris-belcher:
Came across some archival full nodes which run with -blocksonly=1 which would mean it would never ask for the transaction so my code would hang.

chris-belcher:
Travis says X because there's the code in taker.py which isn't tested. It would be hard to test because it requires a socks5 host and port. Perhaps I could find some simply socks5 script that just redirects it to localhost.

chris-belcher:
Right now this uses a user-agent of /JoinMarket:0.2.3/. I don't see any reason to try to hide by using a user agent of /Satoshi:0.13.2/ or something if all the connections will be over tor.

weex:
Tested this out with Bitcoin Core v0.13.2 and get some failures. https://gist.github.com/weex/b64fa1a685efc45e1df2e6be21ba63c8

I think it's because importaddress changed.

My other question is does this diff just connect to dns seed nodes or should it be requesting nodes from the seeds by which to broadcast?

chris-belcher:
Thanks for testing @weex 

Squished commits. I believe this is ready to merge.

Issue #657, opened 2016-11-14
tx_fees in joinmarket.cfg used to configure fee rate


chris-belcher:
The `tx_fees` field in joinmarket.cfg could also be used to configure the fee rate.

For example if `tx_fees` < 144 then interpret it as a block confirm target, otherwise interpret it as satoshi-per-kilobyte.

chris-belcher:
Closing as this was done in PR #665

PR #665, opened 2016-11-20
Allow users to set their own tx fees, as suggested in #657

AlexCato:
While at it, I also fixed the "don't use N=1 with bitcoin core" (which returns -1 and leads to a hardcoded fee instead of honoring the user's wish to target a REALLY fast confirmation block). It's now handled automatically.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -135,6 +135,26 @@ class BlockchainInterface(object):
         required for inclusion in the next N blocks.
 	'''
 
+    def get_fee(self, N):
+        '''Get the fee the user wishes to use. This can either be a manually set
+        fee if the 'block' target is higher than 144 or an estimation by the
+        selected blockchaininterface, if available
+    '''
+        selected_fee = -1 # just to be safe
+
+        if N > 144:
+            selected_fee = N
+        else:
+            selected_fee = self.estimate_fee_per_kb(N)
+
+        # this can happen if e.g. Bitcoin Core has not enough data to estimate
+        if selected_fee < 0:
+            # TODO anything better than a hardcoded default?
+            log.warn('Got no good fee estimation from the blockchain source. Using 30sat/Kb as fallback.')
+            return 30000
+
+        return selected_fee
+
 
 class BlockrInterface(BlockchainInterface):
     BLOCKR_MAX_ADDR_REQ_COUNT = 20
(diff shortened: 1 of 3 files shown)


AlexCato:
Do not merge yet. As suggested on IRC, i'll avoid code duplication by moving the check of N>144 to the base class

AlexCato:
This is my new implementation suggestion. If acceptable, I'll squash this into one commit.

Considering adding a test for that as well, but havent ever touched those. So might take a bit. Want one?

AdamISZ:
utACK. Seems good.

chris-belcher:
Miscommunication on my part but for the avoiding code duplication point I meant putting it in this function https://github.com/JoinMarket-Org/joinmarket/blob/78470c3aeeff31595f21eb70de179edb235e096d/joinmarket/wallet.py#L21

It seems thats the only place estimate_fee_per_kb() is actually called. I was thinking the check 144 thing would go there. It would be more OOP that way and these two lines wouldn't have to be repeated for every new BlockchainInterface.

        if super(BlockrInterface, self).fee_per_kb_has_been_manually_set(N):
            return N

Though this objection is only about code still, we could still merge it today and it would all work.

chris-belcher:
> \<waxwing\> belcher: i pushed alexcato in that direction because it makes more sense (imo) to keep the change within the class, rather than assuming the function will only be called in one place
> \<belcher\> i see
> \<belcher\> right well the word "estimate" implies it would somehow guess rather than take a user input?
> \<waxwing\> true
> \<belcher\> its a style/aesthetic thing ofc
> \<waxwing\> yeah. i guess it comes down to whether the behaviour should be considered global or not, i.e. should it always work like that. but, i take your point for sure.
> \<belcher\> im happy with it going either way
> \<belcher\> maybe its better to keep it this way just because its already written?
> \<belcher\> then maybe replace the word "estimate" with "get" ?
> \<belcher\> i think the wording of stuff is quite important since after a few months they can really help with remembering what certain parts actually do
> \<waxwing\> yeah i could see both being OK, maybe i lean more towards your original POV now, which i'm reluctant to say because i was the one who suggested changing it :) i just didn't like the cognitive load of, if in future you need to call estimate fee somewhere else, you have to remember to add the >144 interpretation.
> \<waxwing\> but maybe not, if things like this are well encapsulated in the wallet.
> \<belcher\> right so you could call wallet.estimate_tx_fee()
> \<belcher\> well one day we will have maker offers that talk about fee rate instead of just fee, so this estimate_fee_per_kb() would be used there..
> \<waxwing\> yes. working on electrum helps bring this into focus. there is so much advantage when everything bitcoin related (e.g. signing) is deferred to the wallet. but that broadens the discussion a bit too much :)
> \<belcher\> how about a new function inside BlockchainInterface called something like get_fee_per_kb() which is in the abstract base class and has the 144 logic inside, so then all code just calls that and nobody calls estimate_fee_per_kb()
> \<belcher\> and that way new implementations of blockchaininterface just implement estimate_fee_per_kb() however they know
> \<waxwing\> i don't follow exactly how that's different from what alexcato has now?
> \<belcher\> 1) it makes the name estimate_fee_per_kb() wrong and 2) every new implementor of blockchaininterface has to have the two lines involving fee_per_kb_has_been_manually_set()
> \<waxwing\> definitely agree for the name. i just didn't quite follow how it's avoiding that duplication? oh, you mean you call the method of the base class?
> \<waxwing\> if so, yes, got it, that's cleaner

AlexCato:
This should be a generalized, future-proof implementation now. Will squash into one commit if deemed acceptable.

AdamISZ:
utACK, please do squash. @chris-belcher I seem to remember you also reviewed the latest here positively? If so I'm happy to do the merge if that's more convenient.

chris-belcher:
Yes, I reviewed it, utACK.

AlexCato:
Already squashed

PR #666, opened 2016-11-20
Add total profit to history command in wallet-tool

Empty2k12:
**This adds the total profit from all successfull coinjoins to the history command in wallet-tool**

This is useful because before you had to manually add all deposits and subtract them from the wallet balance. It displays prior to the annual interest rate and will work without numpy installed.

To see the profit enter this command:

> python wallet-tool.py wallet.json history

It looks like this:

> total profit = 0.00020829 BTC
> continuously compounded equivalent annual interest rate = 6.79291165341 %

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -471,6 +471,7 @@ elif method == 'history':
         now = jm_single().bc_interface.rpc('getblock', [bestblockhash])['time']
     print('     %s best block is %s' % (datetime.datetime.fromtimestamp(now)
         .strftime("%Y-%m-%d %H:%M"), bestblockhash))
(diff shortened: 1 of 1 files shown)


chris-belcher:
Great idea. Percentages are nice but an absolute number of bitcoins is more direct, even if harder to compare between different investments.

Empty2k12:
@chris-belcher not only this but you can easily copy this value to preev.com or wherever and check your profits in fiat. if you then know your electricity price you can find out how and if this is profitable for you.

CohibAA:
tACK

Good idea, thanks.

Issue #656, opened 2016-11-14
wallet-tool showutxos should show tries


eduard6:
wallet-tool `showutxos` should include the number of tries for each utxo. This allows knowing that a wallet has used all of its tries and needs to be split or add external utxos.

Example:
```
$ python wallet-tool.py wallet.json showutxos -p
{
    "<txid>:8": {
        "privkey": "<privkey>", 
        "value": <value>, 
        "address": "<address>",
        "tries": 0
    }, 
    "<txid>:2": {
        "privkey": "<privkey>",
        "value": <value>,
        "address": "<address>"
        "tries": 3
    }, 
}
```

AdamISZ:
Great thought, thanks.

If anyone else would like to do this that'd be fantastic.

chris-belcher:
I'd recommend we actually make something like this the default. Bitcoin addresses should only be shown if they're unused and ready to receive money.

I've written about it here #259 that it would also discourage address reuse.

AdamISZ:
@chris-belcher aren't you talking about something else there though? This is about sourcing commitments. At the moment commitments.json stores used commitments, but it doesn't help a user to know which utxos those commitments were sourced from, nor how many tries are left on each utxo.

chris-belcher:
Yes you're right. I thought OP was proposing a new wallet-tool method called `showutxos`, didn't realise it already existed.

eduard6:
I have the code and will do PR in couple days.

AdamISZ:
@eduard6 great, thanks, if you have any Qs don't hesitate to ask.

AdamISZ:
Implemented and merged, thanks.

PR #672, opened 2016-11-28
Enhance showutxos (JoinMarket-Org/joinmarket#656)

eduard6:
Enhance showutxos (JoinMarket-Org/joinmarket#656)

Include external utxos
Add `mixdepth`
Add `tries`
Add `tries_remaining`
Add `external` (true/false)

To avoid duplicating the code that checks for the number of tries, added `btc.podle.get_podle_tries()`. This is called from `wallet-tool.py` and `btc.podle.generate_podle()` is refactored to use it.

Renamed `tries` parameter for `btc.podle.generate_podle()` to `max_tries` to avoid confusion.

### DIFF
--- a/bitcoin/podle.py
+++ b/bitcoin/podle.py
@@ -327,13 +327,35 @@ def update_commitments(commitment=None, external_to_remove=None,
     with open(PODLE_COMMIT_FILE, "wb") as f:
         f.write(json.dumps(to_write, indent=4))
 
-def generate_podle(priv_utxo_pairs, tries=1, allow_external=None):
+def get_podle_tries(utxo, priv=None, max_tries=1, external=False):
+    used_commitments, external_commitments = get_podle_commitments()
+
+    if external:
+        if utxo in external_commitments:
+            ec = external_commitments[utxo]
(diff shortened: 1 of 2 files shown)


AdamISZ:
Just wanted to mention thanks for doing this; I'll make time to review in the next few days.

AdamISZ:
OK:

1. Reviewed the code in detail. This was the most important bit. Refactor looks good, and doesn't break or change anything (the most sensitive thing here is accidentally "using up" commitments, obviously it doesn't do that).
2. For sanity locally ran test_podle.py; those tests are quite a mess, but do have the virtue of fully putting the podle/commitments logic through its paces. Well, we already knew that was passing, but good.
3. Manually ran the showutxos command against real wallets. Results look good, and changed correctly when I changed the taker_utxo_retries in the config.

So, tACK and I'm going to merge as it's unlikely anyone else will look at this in detail.

Two additional things might be nice in future: 1/an additional tool that shows available utxo commitments for specific transaction sizes; either as an option in wallet-tool or perhaps make a more fully featured analysis tool in cmttools/ , and 2/test of this showutxos method (but this is not very important).

Issue #676, opened 2016-11-30
Feature request: Number of confirmations displayed in wallet-tool.py


qubenix:
It would be nice for me to see how many confirmations each address has while I'm waiting for commitments to become available.

Empty2k12:
Implemented. See #677

Issue #259, opened 2015-10-04
Change wallet-tool display to discourage address reuse


chris-belcher:
Printing out lists of addresses, especially addresses already been used encourages address reuse. Which is bad for privacy, as a privacy-focused project we should try to use the best possible practices.

Instead, display the contents of the wallet as list of UTXOs, with unused addresses available for depositing. Maybe not in detailed form but have UTXOs be the list item granularity instead of addresses. A single UTXO is the closest analogy bitcoin has to an actual "coin".

Perhaps the first few characters of the UTXO hex could be shown, along with their value. There would also need to be a transaction log where people could see when they've received payments. although the majority of people wont be using joinmarket as an actual day to day wallet.

The distinction between UTXOs living on receive addresses and change addresses doesn't matter.

Also have an option that people can use to get back to the old way.

chris-belcher:
These results show the urgency of discouraging address reuse. https://www.reddit.com/r/Bitcoin/comments/3poxoo/obpp_org_on_twitter_how_common_is_address_reuse/

I may reach out to wallet developers to ask if they have a plan to help stop users believing address-reuse is fine.

chris-belcher:
BTW newer version of Electrum don't display the Addresses tab by default anymore, probably for this reason of discouraging address reuse.

chris-belcher:
This feature would also make running wallet-tool faster because the ECC calculations for finding public keys only needs to be done once.

PR #677, opened 2016-11-30
Show confirmations for addresses in wallet-tool (Implements #676)

Empty2k12:
**This pull requests add the ability to see confirmations for the addresses as requested in #676**

The new output looks like this:
```
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00000001 btc 1000 confs 
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00000010 btc 900 confs 
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00000100 btc 800 confs 
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00001000 btc 700 confs 
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00010000 btc 600 confs 
  m/0/0/1/000 1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2 used 1.00100000 btc 500 confs 
```

For unused addresses it does not show anything.

### DIFF
--- a/wallet-tool.py
+++ b/wallet-tool.py
@@ -6,6 +6,7 @@ import json
 import os
 import sys
 import sqlite3
+import json
 from optparse import OptionParser
 
 from joinmarket import load_program_config, get_network, Wallet, encryptData, \
@@ -140,6 +141,15 @@ if method == 'display' or method == 'displayall' or method == 'summary':
         if method != 'summary':
             print(s)
 
+    def json_obj_by_value(haystack, needle, value):
+        for item in haystack:
+            if item[needle] == value:
+                my_item = item
+                break
+        else:
+            my_item = None
+        return my_item
+
     total_balance = 0
     for m in range(wallet.max_mix_depth):
         cus_print('mixing depth %d m/0/%d/' % (m, m))
@@ -152,6 +162,7 @@ if method == 'display' or method == 'displayall' or method == 'summary':
             cus_print(' ' + ('external' if forchange == 0 else 'internal') +
                       ' addresses m/0/%d/%d' % (m, forchange) + ' ' + xpub_key)
 
+            addressinfo = jm_single().bc_interface.rpc('listreceivedbyaddress', [0, False, True])
             for k in range(wallet.index[m][forchange] + options.gaplimit):
                 addr = wallet.get_addr(m, forchange, k)
                 balance = 0.0
@@ -160,6 +171,11 @@ if method == 'display' or method == 'displayall' or method == 'summary':
                         balance += addrvalue['value']
                 balance_depth += balance
                 used = ('used' if k < wallet.index[m][forchange] else ' new')
+                addr_data = json_obj_by_value(addressinfo, 'address', addr)
+                if addr_data:
+                    confs = '{0} confs'.format(addr_data['confirmations'])
+                else:
+                    confs = '' #Leave empty since we don't want this showing for unused addresses
                 if options.showprivkey:
                     privkey = btc.wif_compressed_privkey(
                     wallet.get_key(m, forchange, k), get_p2pk_vbyte())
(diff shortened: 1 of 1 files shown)


chris-belcher:
What if an address contains more than one UTXO ? (because the address was reused)

Also what if the user is using blockr.io as a blockchain interface instead of bitcoin core? And if we code up new blockchain interfaces it also won't work.

I think for this kind of thing #259 would be better

Empty2k12:
> What if an address contains more than one UTXO ? (because the address was reused)
I thought about that, too but since address reuse is discouraged this seems like the best solution going forward (especially considering the future implementation of #259)
> Also what if the user is using blockr.io as a blockchain interface instead of bitcoin core? And if we code up new blockchain interfaces it also won't work.
This is something I did not consider, however. I have been thinking a little bit about this and came to the conclusion that this feature should only be activated when Bitcoin Core is being used for querying the blockchain. Maybe there is a way to integrate this with block explorers as well when they are returning the confirmations for a address (or more specifically the first transaction in the address) with the query, but I have not been able to come up with a data format that allows loading Confs from the server and from Core since the requests are so fundamentally different. Additionally I think this should NOT query a block explorer for every Address. If the data is returned in the initial request we can use it, if not the servers shouln't be spammed with unesseccay requests.
This could further incentivize users to switch to running a full node.
> I think for this kind of thing #259 would be better
What do you mean spcifically by that? I have been thinking about implementing #259 for a while, I can try to tackle that next if you want.

chris-belcher:
What I meant by #259 is that since it would show individual UTXOs then it would make sense to show the number of confirmations there, instead of next to addresses which might have more than one UTXO on them.

It would be fine to skip support for blockr.io, but one day when #653 and #470 are implemented then they also won't work. I agree blockr.io and querying web blockchain explorers was a really bad idea which is why I'm thinking about how #653 and #470 would fit into joinmarket.

This is the reason that the BlockchainInterface class exists, to abstract away the blockchain/bitcoin network code from everything else. I'd say the best way going forward is to have sync_unspent() in all the BlockchainInterfaces also record the block in which the UTXO was mined into. So the unspent dict would look like this

> {0: {u'0000000000000000000000000000000000000000000000000000000000000000:0': {'address': u'1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2',
>                                                                             'value': 100000000,
>                                                                             'blockheight': 410000},
>     u'0000000000000000000000000000000000000000000000000000000000000000:0': {'address': u'1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2',
>                                                                             'value': 100100000,
>                                                                             'blockheight': 410100},
>     u'0000000000000000000000000000000000000000000000000000000000000000:7': {'address': u'1OoOOOOOooOoooOOo9Oo9o9OOo9oOoOOO2',
>                                                                             'value': 101000000,
>                                                                             'blockheight': 410000}},

Then a new method in BlockchainInterface could return the current block height, from which confirmation count can be calculated from for each UTXO.

This would also be needed in the solution to this problem https://github.com/JoinMarket-Org/joinmarket/issues/470#issuecomment-260671512

> In make_commitment() only the UTXOs already in the taker's own wallet must be queried, if the joinmarket wallet kept track of UTXO confirmations then that code could be modified to use wallet.unspent instead of query_utxo_set().

Empty2k12:
I updated the PR to use blockchaininterface, show UTXO instead of address and confirmations for UTXOs

Empty2k12:
I should propably implement some caching for the block-height since simply running `python wallet-tool.py wallet.json` can cause the block height to be requested from Blockr/RPC around 50 times. This is unnessesarily much.

chris-belcher:
There's similar timeout code in the blockr blockchain interface code (look up BlockrInterface.last_sync_unspent)

Since wallet-tool is the only place that issue comes up, it might be better to only call get_block_count() once at the beginning of the 'display' function in wallet-tool and store the result in a local variable.

chris-belcher:
Could you finish this please and then we can merge it?

Empty2k12:
@chris-belcher I will when I fully understood how the tests work and when achool isn't as busy. In the mean time feel free to do alternate implementations that are not as error riddled as mine.

Empty2k12:
@chris-belcher Will do when school work permits. (Most likely this evening (in 12+ hours))

qubenix:
@Empty2k12 Any progress on this?

Issue #674, opened 2016-11-29
Allow empty password / no wallet encryption


chris-belcher:
Some people using joinmarket scripts for internal scripting find the encrypted wallet pointless and difficult to code automatically because it uses get_pass()

One way to do this would to try the empty password "" when decrypting and use that if it works.

Another way is to have all scripts have a command-line option to use as the password. This is bad because it could lead to the password ending up `.bash_history` and would need each script to have this option added.

When creating a wallet with no encryption, a lot of warnings should be given to the user.

PR #679, opened 2016-12-01
Allow no/empty password (Implements #674)

Empty2k12:
**This allows wallets to have a zero length (`''`) password. (As requested in #674)**

When generating a new wallet and pressing enter both times when you are asked to set a password it will present you with the following warning
```
============= WARNING =============
Using no password is very dangerous
===================================
Abort? [Y/n]
```
To proceed with a empty password you have to confirm with a `n`, pressing enter will stop the generation process.

The wallet checks `''` and if it does not unlock the wallet it will ask the user to input the password.

### DIFF
--- a/joinmarket/wallet.py
+++ b/joinmarket/wallet.py
@@ -177,11 +177,16 @@ class Wallet(AbstractWallet):
                 self.index_cache += [[0,0]] * (
                     self.max_mix_depth - len(self.index_cache))
         decrypted = False
+        trieddefault = False
         while not decrypted:
-            if pwd:
-                password = pwd
+            if trieddefault:
+                password = ''
+                trieddefault = True
             else:
-                password = getpass('Enter wallet decryption passphrase: ')
+                if pwd:
+                    password = pwd
+                else:
+                    password = getpass('Enter wallet decryption passphrase: ')
             password_key = btc.bin_dbl_sha256(password)
             encrypted_seed = walletdata['encrypted_seed']
             try:
@@ -196,7 +201,8 @@ class Wallet(AbstractWallet):
                 else:
                     raise ValueError
             except ValueError:
-                print('Incorrect password')
+                if not trieddefault:
+                    print('Incorrect password')
                 if pwd:
                     raise
                 decrypted = False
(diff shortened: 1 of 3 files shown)


chris-belcher:
Thanks for doing this!

Please change the target branch to be joinmarket:develop not joinmarket:master

Empty2k12:
@chris-belcher Sorry, should know by now which branch is the development branch. Changed it!

juscamarena:
tACK

joinmarket/wallet.py
```diff
@@ -177,11 +177,16 @@ def read_wallet_file_data(self, filename, pwd=None):
                 self.index_cache += [[0,0]] * (
                     self.max_mix_depth - len(self.index_cache))
         decrypted = False
+        trieddefault = False
         while not decrypted:
-            if pwd:
-                password = pwd
+            if trieddefault:
+                password = ''
+                trieddefault = True
```
AdamISZ: I can't see the logic of this (lines 180-184, plus 204-5); `trieddefault` is set to False before the decryption-attempting loop, and the password is set to '' if `trieddefault` is True; but it is not set to True anywhere except within that condition where it has already evaluated as True. I'm guessing what you intended is, line 182 to read `if not trieddefault`, thus to enforce an attempt to unlock with a null password once only.

AdamISZ: Yes, after testing with null and non-null password wallets, the functionality seems to be correct with that single edit.

AdamISZ:
ping @Empty2k12 with that one edit ^ i'm happy to merge this.

Empty2k12:
@AdamISZ I implemented the change. You were right. I'm still wondering how I tested both cases and it worked.

AdamISZ:
Getting failures in test_wallets test script. Sorry I don't have time to figure it out now, if someone else did, that'd be great. You can see the errors in the travis log (it's a way up because of the logfile printouts at the bottom).

Empty2k12:
Thank you for the implementation of the test case @peterzen

Empty2k12:
@peterzen [This](https://travis-ci.org/JoinMarket-Org/joinmarket/builds/208319308) is the log output for the last test run.

chris-belcher:
Added my own commit which fixes a small thing that makes the test now work (although coverage reduced so it says test failed)

Will merge with merge.py soon

peterzen:
Hi Chris - re your message; my address is indeed peter@prioritylane.com.  Thanks!

PR #683, opened 2016-12-05
Retry Blockr requests on network error

eduard6:
Retry Blockr requests on network error (DNS, timeout, etc)

Does not retry HTTP errors (to be conservative to avoid retrying invalid requests)

### DIFF
--- a/bitcoin/bci.py
+++ b/bitcoin/bci.py
@@ -11,8 +11,10 @@ if platform.system() == "Windows":
 else:
     try:
         from urllib.request import build_opener
+        from urllib.error import URLError, HTTPError
     except:
         from urllib2 import build_opener
+        from urllib2 import URLError, HTTPError
 
 log = get_log()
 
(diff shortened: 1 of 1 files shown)


AdamISZ:
Needs a rebase, should be trivial. It might be nice if someone tests this.

eduard6:
Rebased

AdamISZ:
Have we thought through what happens if a long sleep like this (up to 600 s) is occurring in a thread while other work continues?

eduard6:
> Have we thought through what happens if a long sleep like this (up to 600 s) is occurring in a thread while other work continues?

I had not. I have not experienced any issues with it though even when it sleeps for that long. The time.sleep() call was based on the one from make_blockr_request() which has a maximum of 60 s. If 600 s seems too long I can change this pull request to 60 s too.

eduard6:
Closing since blockr is obsolete now.

PR #684, opened 2016-12-05
Change user agent to Firefox ESR for improved privacy

eduard6:
Change user agent to Firefox ESR for improved privacy

Currently user agent is set to a string that (in practice) identifies a request as being from JoinMarket. This is bad for privacy.

This PR changes the user agent to a standard Firefox ESR user agent string. This is what is used by Tor Browser and will help JoinMarket requests blend in more. New ESR releases are made once a year so this will not need to be changed often.

### DIFF
--- a/bitcoin/bci.py
+++ b/bitcoin/bci.py
@@ -25,7 +25,7 @@ def make_request(*args):
     else:
         opener = build_opener()
     opener.addheaders = [('User-agent',
-                          'Mozilla/5.0' + str(random.randrange(1000000)))]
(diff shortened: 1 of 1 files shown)


AlexCato:
Fantastic idea.
utAck (not much to test anyways)

AdamISZ:
utACK

(other things could be done no doubt but i see no negative side to this change)

CohibAA:
utACK, good catch

AdamISZ:
Merging in absence of other comments.

PR #685, opened 2016-12-08
Improve wallet tool help format

eduard6:
Improve wallet tool help format

Ugly hack but I think it is better than what was there.


chris-belcher:
Wow I didn't think to try if this worked. Great!

CohibAA:
tACK - Much easier to read now, thanks.

PR #686, opened 2016-12-29
Add link to "High level design"

meeDamian:
It took me a good while to find it, and was really helpful in making me understand fundamentals.


chris-belcher:
ACK

AdamISZ:
Also ACK but please base off develop, also I think it'd probably be better to phrase it as "go to (link to wiki) to get started, or for more in depth technical information,see (design doc)", something like that. Also worth noting that while the messaging protocol *.md file in the Docs repo has been updated for 0.2, then high level design has not as yet (not a huge deal, but would be nice for it to be updated).

meeDamian:
Rebased to `develop` and wording changed as suggested. 

I think, even slightly outdated, design doc helps to understand the fundamental concepts a lot. And it can be updated any time later as well.

AdamISZ:
Yes, thanks, will pop that in now.

PR #687, opened 2016-12-30
improved patientsendpayment now that the new protocol allows it to work

chris-belcher:
Added some improvements to patientsendpayment, including:

1. Can have many send jobs. It will send an amount of bitcoin to one destination and when that's done it can start sending to another.

1. Can send to many addresses per destination to avoid address reuse, can also accept a xpub BIP32 key.

1. Has all the timeout callback improvements that tumbler and sendpayment have when being a taker.

1. Can be configured to wait forever and never become a taker. (Simple idea but a bit of an oversight before)

Currently in the process of making a wiki page with more details of how to use it.

### DIFF
--- a/patientsendpayment.py
+++ b/patientsendpayment.py
@@ -3,17 +3,38 @@ from __future__ import absolute_import
 import sys
 import threading
 import time
+import random
 from datetime import timedelta
 from optparse import OptionParser
-# data_dir = os.path.dirname(os.path.realpath(__file__))
-# sys.path.insert(0, os.path.join(data_dir, 'joinmarket'))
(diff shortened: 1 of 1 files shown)


chris-belcher:
Added tutorial to wiki https://github.com/JoinMarket-Org/joinmarket/wiki/Sending-payments-with-CoinJoin#patient-send-payment

AdamISZ:
I've tested this in only one mode: a run that waits and then switches to Taker and completes successfully.

A full test of the functionality is pretty involved (maker behaviour + xpub). I'd be happy to merge based on your own tests/work on it @chris-belcher , let me know if you want to do that.

PR #688, opened 2016-12-31
fixup README installation instructions

instagibbs:


AlexCato:
This is not really a bug: if you downloaded the .tar.gz file from the first link given, the filename is correct as it is written right now.
If 2nd (direct) link was used, your correction would be the way to do it.

Since I assumed that the first link is used by most people, I used the extract command to fit that filename.

instagibbs:
hm, perhaps I can change the other line to `wget https://github.com/JoinMarket-Org/joinmarket/archive/v0.2.2.tar.gz -O joinmarket-0.2.2.tar.gz`?

AlexCato:
Since we're on the command line anyways at that point, that's imo an improvement over the status quo. Fine with me!

chris-belcher:
Can you merge into the develop branch not the master branch

PR #689, opened 2016-12-31
Clear up installation using virtualenv

juscamarena:


AdamISZ:
I don't know the right way to look at this (hence the original tentative language), because virtualenv is not something non-expert users really understand, while on the other hand *not* using it can be a recipe for mess. As for this edit, not sure if it really helps (of course in *some* cases it would).

weex (changes requested):
Why not consolidate this text with the bit above? The (this) \n (that) format looks weird as well.

WyseNynja:
I use virtualenv for my install without any issue, and I definitely think we should recommend it. I've seen virtualenv save many a system python install from breaking. However, OSX and matplotlib and virtualenv don't work very well together: http://matplotlib.org/faq/virtualenv_faq.html#working-with-matplotlib-in-virtual-environments

PR #695, opened 2017-01-04
Removed configurable mixdepths from yieldgens, also reworded error message

chris-belcher:
The configurable mixdepths never made sense, and tripped up some people. Also the error message lead to lots of confusion in the IRC channel.

### DIFF
--- a/joinmarket/yieldgenerator.py
+++ b/joinmarket/yieldgenerator.py
@@ -16,6 +16,8 @@ from joinmarket import get_irc_mchannels
 
 log = get_log()
 
+MAX_MIX_DEPTH = 5
+
 # is a maker for the purposes of generating a yield from held
 # bitcoins
 class YieldGenerator(Maker):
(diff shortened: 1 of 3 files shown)


AdamISZ:
Reviewed, seems clear, merging.

yg-pe.py
```diff
@@ -43,14 +42,15 @@ def create_my_orders(self):
             # minimum size bumped if necessary such that you always profit
             # least 50% of the miner fee
             self.minsize = max(
-                int(1.5 * self.txfee / float(self.cjfee_r)), self.minsize)
+                int(1.5 * self.txfee / float(self.cjfee_r)), max_minsize)
```
WyseNynja: are you sure this is supposed to be max_minsize? Isn't self.minsize already set to that? I think this was right as it was

chris-belcher: I changed the name because `self.minsize` and `minsize` had the same name and I believed that made it less readable.

It shouldn't make a difference, `self.minsize` was either set to `minsize` (which is now called max_minsize) or to the `int(1.5 * self.txfee / float(self.cjfee_r))` expression which is the same every time its evaluated.

joinmarket/yieldgenerator.py
```diff
@@ -16,6 +16,8 @@
 
 log = get_log()
 
+MAX_MIX_DEPTH = 5
```
periodic1236: Removing the configurable mixdepths is troublesome for wallets that already use more than the default of 5 mixdepths, as every time you quit the yield generator it truncates the index cache, having not read more than 5 levels into memory.

I worked around this for the time being by adjusting this value. Maybe I'll start over with a new wallet at some point.

PR #700, opened 2017-01-09
max_minsize is actually min_minsize, rename it

ian-kelling:
c = max(a, b), a and b are minimums of c, not maximums.


chris-belcher:
Yes, that was an oversight.

PR #702, opened 2017-01-28
yg-pe: Improve yg comments

AlexCato:
Since there's sometimes confusion about what the parameter actually mean, I've added verbose comments to it.

Also:
- remove nickserv_password parameter (not needed any more)
- rename max_minsize to min_minsize as in #700


chris-belcher:
Looks fine to me.

AdamISZ:
OK, thanks for eyes on it, was waiting in case the min_minsize thing was a problem, but it fits earlier discussion. Merging.

PR #703, opened 2017-01-28
Delete yield-generator-basic.py

AlexCato:
Rationale:
yg-pe.py does the same thing, but slightly better (better mixlevel selection). Imho we should just maintain one yield generator script and offer other ones in the "custom script" repo.


undeath:
Is it a feature of yg-pe.py to not always use output mixlevel = input mixlevel + 1? If it is a feature (and not a bug) I think yield-generator-basic.py should be kept to allow not screwing up mixlevels.

AdamISZ:
Yes, it's a feature, the idea is to minimize order re-announcement by targetting a mixdepth which doesn't change the offer size. I agree this should not be the only option. There is a custom-scripts repo which I'd expected people to put their alternative yieldgen algos in, but not much has been happening (there is a -randomizer script which I prefer to yg-pe actually).

Having said all that I'm not sure the deterministic movement +1 should be that important, but maybe it's useful in ways I haven't thought of.

undeath:
Not respecting mixlevels means rendering trust assumptions for each mixlevel void. If a maker is a passive taker that can be a severe issue.

AdamISZ:
What do you mean by "trust assumptions for mixlevels"?

"If a maker is a passive taker" - don't understand that either; a taker isn't passive.

Guessing, but I'm wondering if you're thinking about a scenario of doing a send (being a Taker) with a wallet that is currently running as a Maker. I don't recommend doing that, you should shut down the yieldgenerator before doing a sendpayment, then restart it.

(Re "taker isn't passive": there is patientsendpayment mixing maker and taker roles, but don't think you'd mean that? Well, I'll stop guessing and let you clarify)

undeath:
I mean cases where a maker is interested in anonymizing their own bitcoins, too. Sorry about the confusing choice of words.

AdamISZ:
I think I get what you're saying, perhaps: that during the initial phase where coins are moving strictly only forwards (from 0 to 4 mixdepth), there is no "revisiting" of the same mixdepth. It's not a particularly notable difference in my view, given that each receiving address is new. With the -basic algo, you'll still very often be making several transactions with the same from-to mixdepth pattern (e.g. you may do several transactions from 0 to 1 before anything else). As the coins start to spread "downwards" this "reuse of same mixdepth" will continue to happen often. Is it really that different? Well perhaps the discussion is a bit pointless, there may be some specific pattern of activity in which that "forwards only" is preferable, not sure. After a large number of transactions such that wrap-around is occurring I'm not sure I see any notable difference. If there's a particular scenario you have in mind, feel free to explain it, it might give ideas for other people too.

undeath:
With the -basic generator you can be sure whatever is in your highest mixlevel has been through sufficiently many coinjoins. With -pe you can not make such an assumption. Withdrawing from any address becomes an anonymity-gamble.

AdamISZ:
Ah, I see. Very simple, I missed it :) Yes, you have a guaranteed minimum number of joins for coins in mixdepth N, =N, although the maximum could be anything. Well, that's another quite good reason not to remove it that I didn't think of, thanks.

AdamISZ:
I'm going to close this on the basis there isn't agreement to do it. Feel free to reopen.

PR #704, opened 2017-01-28
Switch blockr to electrum as default blockchain interface

AlexCato:
Rationale:
blockr as default does not work for TAILS and other TOR users any more. This will reduce potential confusion, while at the same time not reducing privacy (blockr potentially has worse privacy implications, electrum is at least somewhat decentralized).

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -84,7 +84,7 @@ required_options = {'BLOCKCHAIN': ['blockchain_source', 'network'],
 defaultconfig = \
     """
 [BLOCKCHAIN]
-blockchain_source = blockr
+blockchain_source = electrum
 #options: blockr, bitcoin-rpc, regtest, bc.i, electrum
 # for instructions on bitcoin-rpc read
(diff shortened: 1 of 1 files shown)


chris-belcher:
Yes but we should test the electrum blockchain interface first.

I just found this bug in the pushtx() function while reading it https://github.com/JoinMarket-Org/joinmarket/blob/develop/joinmarket/electruminterface.py#L246

AdamISZ:
On being default, I had two reasons not to want to do that: first, it's slow without some improvements, two, it's only had minimal testing.

Thanks for that reviewing/bug find @chris-belcher . I have only tested this a few times, on main net (no other way sadly), and I only tested it from Taker side. As I mentioned on reddit, this is intended as a help for Takers, not Makers (because I think it will be unworkable with big wallets until some improvements are made). So that's why I didn't encounter that particular bug (return value only checked for truthiness in sendpayment).

Re: testing strategy. It's not going to be easy, but I guess a 'mock' style approach will work OK, it'll just be a bit of a pain making sure of the correct return values for the various electrum server calls, and then setting up some dummy data to return.

Overall it needs work ... just seems better to have it there as a fallback right now, than nothing ... I thin k bc.i is very nearly unusable.

PR #705, opened 2017-01-29
fixed pushtx return value

chris-belcher:
Fixed return value of electrum blockchain interface.

Untested by me, but probably works given its simplicity :fingers crossed:


AdamISZ:
utACK, but it should be fine, I took a mo' to check the bc.i interface, it returns txid/None instead of True/False so requires similar simple fix. Thanks, will merge now.

Test setup remains a need, if anyone wants to do what i did and try a couple of taker transactions, great, also trying yield gen would help (although i was quite explicit on reddit that it isn't intended for that use case due to slowness). afaict from testing on the *clientserver repo the handling of confirm/unconfirm callback is correct but for sure needs more attention on it.

PR #706, opened 2017-02-07
avoid re-importing already imported addesses to bitcoind

undeath:
When joinmarket notices un-imported addresses while synchronizing the wallet history with bitcoind it will re-import all used addresses again. Depending on the age and usage of the joinmarket wallet that can take several hours to complete.

This patch prevents already imported addresses from being re-imported.

### DIFF
--- a/joinmarket/blockchaininterface.py
+++ b/joinmarket/blockchaininterface.py
@@ -1025,7 +1025,7 @@ class BitcoinCoreInterface(BlockchainInterface):
         wallet_name = self.get_wallet_name(wallet)
         #TODO It is worth considering making this user configurable:
         addr_req_count = 20
-        wallet_addr_list = []
+        wallet_addr_set = set()
         for mix_depth in range(wallet.max_mix_depth):
             for forchange in [0, 1]:
                 #If we have an index-cache available, we can use it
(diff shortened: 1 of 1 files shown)


AdamISZ:
My first reaction is "great, finally someone else is looking at this!"

Second is: you are aware I hope that you can use the `--fast` option to avoid running this code once you have synced the first time? It avoids all these re-imports on restart.

Third thing is, and I'm probably just being dumb, but I can't see how this changes the number of imports: [this line](https://github.com/JoinMarket-Org/joinmarket/pull/706/files#diff-3b408ed85f23411e81db7c0ae451edfdR1053) gets executed once per mixdepth, once per internal/external branch, so why does it make a difference if it's a .update on a set rather than an append to a list?

undeath:
The relevant change is in this line: https://github.com/JoinMarket-Org/joinmarket/pull/706/files#diff-3b408ed85f23411e81db7c0ae451edfdR1067

The other changes are only caused by switching the types from list to set. (to avoid needless casts)

AdamISZ:
Doh! Yes, will take another look but makes sense. Do use --fast though.

chris-belcher:
Thanks for this. I looked at the code quickly now and it seems okay but I'll have to look again since this part of the code was always quite subtle and confusing (to me).

It would be good to test this in different situations. Like test when sync'ing a fresh new wallet that isn't in your full node yet, test when sync'ing a wallet where your full node has all the addresses, test when sync'ing a wallet where your full node has half the addresses and must import the other half.

AdamISZ:
@chris-belcher yep but do note: a variety of sync scenarios are tested in test_wallets now. ACK from me.

AdamISZ:
I am going to merge this. If anyone could please try it out ASAP, would be greatly welcome! I will run it myself immediately (easy way is wallet-tool without --fast and then with). It's difficult to see how it could cause a problem, given sanity testing has happened already.

Issue #707, opened 2017-02-07
sendpayment.py error SELECT COUNT(DISTINCT counterparty) FROM orderbook;


Onefox:
i just want to send some coins out of my wallet with:
 python sendpayment.py -m 4 -N 4 wallet.json AMOUNT ADDRESS

I did this plenty of times before but now it stoppt working, 

`2017-02-07 22:41:26,409 [MainThread  ] [INFO ]  hello joinmarket
2017-02-07 22:41:26,409 [MainThread  ] [INFO ]  Joinmarket directory is: /home/joinm/joinmarket-0.2.2
2017-02-07 22:41:26,423 [MainThread  ] [INFO ]  Estimated miner/tx fee for each cj participant: 37742
2017-02-07 22:41:26,424 [MainThread  ] [INFO ]  starting sendpayment
Enter wallet decryption passphrase:
/usr/local/lib/python2.7/dist-packages/secp256k1/__init__.py:228: UserWarning: implicit cast from 'char *' to a different pointer type: will be forbidden in the future (check that the types are as you expect; use an explicit ffi.cast() if they are correct)
  self.ctx, res_compressed, outlen, self.public_key, compflag)
/usr/local/lib/python2.7/dist-packages/secp256k1/__init__.py:506: UserWarning: implicit cast from 'char *' to a different pointer type: will be forbidden in the future (check that the types are as you expect; use an explicit ffi.cast() if they are correct)
  res = func(inst.ctx, key, scalar)
2017-02-07 22:41:29,541 [MainThread  ] [INFO ]  requesting detailed wallet history from Bitcoin Core client
2017-02-07 22:41:36,417 [MainThread  ] [INFO ]  starting sendpayment
2017-02-07 22:41:36,419 [MainThread  ] [INFO ]  starting message channels
2017-02-07 22:41:36,423 [MCThread    ] [INFO ]  connecting to host 6dvj6v5imhny3anf.onion6698
2017-02-07 22:41:36,432 [MCThread    ] [INFO ]  connecting to host cfyfz6afpgfeirst.onion6667
2017-02-07 22:41:40,656 [MCThread    ] [INFO ]  Connected to IRC and joined channel on agora-irc
waiting for all orders to certainly arrive
2017-02-07 22:41:40,866 [MCThread    ] [INFO ]  Connected to IRC and joined channel on CgAn
Exception in thread PaymentThread:
Traceback (most recent call last):
[...]

PR #708, opened 2017-02-08
lock db access from taker scripts to avoid cursor exceptions

AdamISZ:
See #707 and my comment on it for explanation.

Testing: I'm finding it very difficult to reproduce, I believe this is because it fundamentally depends on the speed with which the db call returns; even with 6 makers x 300 orders each (done by slight tweaks to scripts and running ygrunner/sendpayment) I was only able to reproduce the "Recursive use of cursor" exception once, so not reliably, at least yet.

On the other hand, the test suite passes, so in the case that this *doesn't* solve the observed problem in the live pit (well, I hope it does), I don't see it causing a problem.

### DIFF
--- a/sendpayment.py
+++ b/sendpayment.py
@@ -101,8 +101,9 @@ class PaymentThread(threading.Thread):
         self.ignored_makers = []
 
     def create_tx(self):
-        crow = self.taker.db.execute(
-            'SELECT COUNT(DISTINCT counterparty) FROM orderbook;').fetchone()
+        with self.taker.dblock:
+            crow = self.taker.db.execute(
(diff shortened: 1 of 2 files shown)


AdamISZ:
I guess a couple of subsidiary notes: there are other places where db access occurs, it may make sense to go with a global lock for the whole application, and use it everywhere. Until that possibility is addressed, the only danger should be exception throws like #707, since obviously sqlite3 is not allowing concurrent access.

AdamISZ:
Yeah, that's what I meant by the "subsidiary" comment. It's not that I didn't notice the others.

PR #710, opened 2017-02-11
For CJs: ignore makers with overlapping offer ranges

AlexCato:
support.py: checkes if makers have overlapping offers and removes them from being considered for joins

This is done efficiently in O(n) by adding up the ranges of a maker's offers.
Example:
Maker has 3 offers:
a) min=1, max=3
b) min=1.5, max=4
c) min=3, max=5

First, the "total" maximum range of that maker is highest_max - lowest_min (in this case: 5-1=4).
Now the single ranges are iteratively added up and checked against the "total" max.

 If there is no overlap, then the total max will always be >= the added up ranges of the individual offers.

Currently, about 25% of all makers use overlapping ranges and would be filtered out by this.

### DIFF
--- a/joinmarket/support.py
+++ b/joinmarket/support.py
@@ -249,6 +249,7 @@ def pick_order(orders, n):
 def choose_orders(db, cj_amount, n, chooseOrdersBy, ignored_makers=None):
     if ignored_makers is None:
         ignored_makers = []
+    remove_invalid_makers_from_db(db)
     sqlorders = db.execute(
         'SELECT * FROM orderbook WHERE minsize <= :cja AND :cja <= maxsize;',
         {'cja': cj_amount}).fetchall()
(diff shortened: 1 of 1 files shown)


AdamISZ:
Took another quick look at algo (not the code yet)

This *does* allow overlaps if entire range not covered, right? (example: (1,3), (7,9), (12,14), (13,15) -> total range 14, sum 8); are we saying therefore that since they're not exploiting the maximum range they could, it's OK? On reflection, that's probably not OK, since maybe the "action" is in the 12-15 range and they still get away with it by using the "dummy" low offers.

AlexCato:
Agreed, there can still be overlaps.
I'd also prefer to solve this once and for all and just check for actual overlaps; combined with #711 there is also no more performance issues.

Will change the algorithm as soon as I have some more time on my hands.

AlexCato:
Will build this on top of #712 once merged and improve/change it as discussed.

PR #711, opened 2017-02-12
Takers: filter out makers with many offers

AlexCato:
Takers: filter out makers with more than X offers (default 5), configurable.

Tested and works: sendpayment

Since I've never used the tumbler or patientsendpayment, I did not test those. Please have an extra-careful look at the changes there.

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -60,6 +60,7 @@ global_singleton.bc_interface = None
 global_singleton.ordername_list = ['absoffer', 'reloffer']
 global_singleton.commitment_broadcast_list = ['hp2']
 global_singleton.maker_timeout_sec = 60
+global_singleton.max_maker_offers_allowed = 5
 global_singleton.debug_file_lock = threading.Lock()
 global_singleton.debug_file_handle = None
 global_singleton.blacklist_file_lock = threading.Lock()
@@ -131,6 +132,11 @@ merge_algorithm = default
 # responsive. Should be an integer >=2 for privacy, or set to 0 if you
 # want to disallow any reduction from your chosen number of makers.
 minimum_makers = 2
+# For takers: to filter out potentially malicious makers, this is
+# the maximum number of offers a maker is allowed to have. If a
+# makers has more offers, then he will be disregarded as potential
+# counterparty for coinjoins
+max_maker_offers_allowed = 5
 # the fee estimate is based on a projection of how many satoshis
 # per kB are needed to get in one of the next N blocks, N set here
 # as the value of 'tx_fees'. This estimate is high if you set N=2,
@@ -346,6 +352,13 @@ def load_program_config():
             'TIMEOUT', 'maker_timeout_sec')
     except NoOptionError:
         log.info('TIMEOUT/maker_timeout_sec not found in .cfg file, '
+                  'using default value')
+
+    try:
+        global_singleton.max_maker_offers_allowed = global_singleton.config.getint(
+            'POLICY', 'max_maker_offers_allowed')
+    except NoOptionError:
+        log.info('POLICY/max_maker_offers_allowed not found in .cfg file, '
                   'using default value')
 
     # configure the interface to the blockchain on startup
--- a/joinmarket/support.py
+++ b/joinmarket/support.py
@@ -381,6 +381,17 @@ def choose_sweep_orders(db,
     return result, cj_amount, total_fee
 
 
+def remove_makers_with_too_many_offers(db, number_allowed_offers):
+    # removes makers with more than the number of allowed offers
+    crows = db.execute(
+        'SELECT DISTINCT counterparty FROM orderbook WHERE oid >= ?;', (int(number_allowed_offers),)).fetchall()
+
+    for row in crows:
+        nick = row["counterparty"]
+        db.execute('DELETE FROM orderbook WHERE counterparty=?;', (nick,))
+        log.debug('The following maker had too many offers and was removed: ' + nick)
+
+
 def debug_dump_object(obj, skip_fields=None):
     if skip_fields is None:
         skip_fields = []
--- a/patientsendpayment.py
+++ b/patientsendpayment.py
@@ -11,6 +11,7 @@ from joinmarket import Maker, Taker, load_program_config, IRCMessageChannel, \
 from joinmarket import validate_address, jm_single, get_p2pk_vbyte
 from joinmarket import get_log, choose_orders, weighted_order_choose, \
     debug_dump_object, sync_wallet
+from joinmarket.support import remove_makers_with_too_many_offers
 from joinmarket import Wallet
 
 import bitcoin as btc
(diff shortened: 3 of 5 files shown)


chris-belcher:
Some points:

It would be better to put the `remove_makers_with_too_many_offers()` function calls in one of the superclasses like OrderbookWatch or Taker. That way it only needs to be called once instead of in every application script, avoiding code-reuse is a benefit of OOP and we should use it.

Recommend to reword the 'potentially malicious makers' phrasing in joinmarket.cfg, from what I see these makers don't do any harm except make the sqlite database big.

The SQL query of checking the oid value isn't secure. Makers choose their own oid values, today the oid starts from zero and counts up but it can be anything. Makers could set their oid=9001, only have a single offer and get banned anyway. Recommend instead to use an SQL query like `'SELECT oid, counterparty FROM orderbook WHERE COUNT(oid) > ?;` which should return all the (oid, counterparty) that have more than the threshold count of offers. You could probably also do it in one statement using the sql DELELE but it's been a while since I did sql so don't remember how.

Perhaps the threshold of 5 offers per maker is too low? There might be some nice algorithms that involve a few more offers than that. For example I remember yield-generator-mixdepth announced 6 offers, one for each mixdepth and an absoffer from zero to the balance of its lowest balance mixdepth.

AdamISZ:
> Perhaps the threshold of 5 offers per maker is too low? There might be some nice algorithms that involve a few more offers than that. For example I remember yield-generator-mixdepth announced 6 offers, one for each mixdepth and an absoffer from zero to the balance of its lowest balance mixdepth.

Yes, in my reddit post I referred to 20 with this history in mind. If it's user configurable then the maker just takes his chances, but in any case I agree we can probably be *fairly* lax.

chris-belcher:
Though it's user configurable, the default will be the case for almost all users, especially with the joinmarket.cfg file having so many options now.

AlexCato:
Thanks for checking this out so quickly!

> It would be better to put the remove_makers_with_too_many_offers() function calls in one of the superclasses like OrderbookWatch or Taker. That way it only needs to be called once instead of in every application script, avoiding code-reuse is a benefit of OOP and we should use it.

I've tried to find a suitable place in OrderbookWatch or Taker, but didnt see any. I especially wanted to make sure that this is only done when necessary, i.e. not every time a new offer is announced (only if someone actually wants to take one). Open for suggestions!

> Recommend to reword the 'potentially malicious makers' phrasing in joinmarket.cfg, from what I see these makers don't do any harm except make the sqlite database big.

Can do. Though I meant "potentially" as in "hints that this is not necessarily the case". Will be changed.

> The SQL query of checking the oid value isn't secure. Makers choose their own oid values, today the oid starts from zero and counts up but it can be anything. Makers could set their oid=9001, only have a single offer and get banned anyway. Recommend instead to use an SQL query like 'SELECT oid, counterparty FROM orderbook WHERE COUNT(oid) > ?; which should return all the (oid, counterparty) that have more than the threshold count of offers. You could probably also do it in one statement using the sql DELELE but it's been a while since I did sql so don't remember how.

I did indeed think about this. Yes, makers can choose any OID they want, but why should they? This statement allows exactly OID 0 through 4 to be used; I can change this so that any combination of 5 OIDs works, but I really dont see what this is good for. I'd not consider it "not secure" though.
Will do if you prefer that still.

> Perhaps the threshold of 5 offers per maker is too low? There might be some nice algorithms that involve a few more offers than that. For example I remember yield-generator-mixdepth announced 6 offers, one for each mixdepth and an absoffer from zero to the balance of its lowest balance mixdepth.

Again, I'd prefer lower over higher, and even if it's only meant to make the orderbook more human-readable. >10 makers with 20 offers each violate this goal imho.

> Though it's user configurable, the default will be the case for almost all users, especially with the joinmarket.cfg file having so many options now.

Strongly agree. The importance of defaults can hardly be overstated.

chris-belcher:
> I've tried to find a suitable place in OrderbookWatch or Taker, but didnt see any. I especially wanted to make sure that this is only done when necessary, i.e. not every time a new offer is announced (only if someone actually wants to take one). Open for suggestions!

in taker.create_cj() perhaps

> I did indeed think about this. Yes, makers can choose any OID they want, but why should they? This statement allows exactly OID 0 through 4 to be used; I can change this so that any combination of 5 OIDs works, but I really dont see what this is good for. I'd not consider it "not secure" though.
Will do if you prefer that still.

Fair point, it just looked wrong on first impression. Maybe put a comment with a short explanation for anyone who makes the same assumption as me.

The aim of a human-readable orderbook probably can't happen long term if the number of makers goes up a lot. Exchange orderbook visuals like http://bitcoinwisdom.com/ collapse their orderbook into a depth chart.

AdamISZ:
Re: moving to superclass, yes, clearly desirable. I also can't see a more logical approach than having a function in `Taker`; doesn't really fit in OrderbookWatch currently given Makers are subclassing it too now.

Re: restricting oid values or not, I too can see both sides of that argument. I guess I agree with @chris-belcher 's comments, it's OK but needs to be mentioned somewhere as a new rule.

chris-belcher:
> If not that, then it does seem logical to add a function in Taker. But unless I missed something, that still requires added calls from user-level scripts.

I meant adding it to taker.start_cj() not taker.create_cj() (which doesn't exist)
https://github.com/JoinMarket-Org/joinmarket/blob/78470c3aeeff31595f21eb70de179edb235e096d/joinmarket/taker.py#L639

The only downside would be that it never gets called in the ob-watcher.py script, although that script could add it to the loop that periodically says !orderbook

In a situation like this it would be nice to use socket timeouts to implement a heartbeat, which already exists in the p2p network code https://github.com/JoinMarket-Org/joinmarket/blob/3756cc498b2cd6a050003da62590c31d3d13f3df/joinmarket/peertopeer.py#L297 and also in this experimental IRC code https://gist.github.com/chris-belcher/39033530fe550506b833e71c6f19f82d If we had that then this `remove_makers_with_too_many_offers()` function could go in the heartbeat function.

AdamISZ:
Yeah, start_cj makes more sense. Architecturally it does fit into OBW class though, do you see anything wrong with the on_order_seen suggestion? That way they would never make it into the db in the first place.

We already have a delete call into the db in that function, I guess an extra query in there is not an issue. Just need to have a variable like ignored_makers there so as to remember to ignore all after ban.

AlexCato:
Thanks for the comments! I hope I'll find some time this weekend to build a solution closer to these ideas.

AlexCato:
So, had a look at it now:

At the point in time when start_cj() is called, the maker-counterparties are already chosen. Sure, I could remove those that do not fit certain criteria (e.g. too many offers or overlapping offers) by cross-checking their other offers in the orderbook-db, but then the CJ would be performed with fewer counterparties than the taker expected/preferred.
I think thats not much of an improvement. Did I miss something?

on_order_seen() might work, if I keep add a blacklist of "bad" makers in its state.
Objections to me implementing it there?

AlexCato:
Implemented as discussed in #712

PR #712, opened 2017-02-17
takers: ignore makers with more than max_offers_allowed offers

AlexCato:
The option is configurable and defaults to 15.

Implemented as discusses in #711 and waxwing/belcher on IRC.

I'll build the overlapping offer checks on top of this soon.

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -60,6 +60,7 @@ global_singleton.bc_interface = None
 global_singleton.ordername_list = ['absoffer', 'reloffer']
 global_singleton.commitment_broadcast_list = ['hp2']
 global_singleton.maker_timeout_sec = 60
+global_singleton.max_maker_offers_allowed = 15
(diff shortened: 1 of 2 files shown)


AdamISZ:
Looks good to me, utACK

Also, I think 15 is reasonable. I wouldn't object to 5-10 either. But I think I'm more influenced by our low-bandwidth message channel (public IRC), it isn't necessary. Anyway ignore that, bikeshedding :)

We could consider a `self.msgchan.send_error` message to signal to the maker that they were banned?

WyseNynja:
Will the orderbook also ignore these makers with these changes, or do we need more code for that?

AdamISZ:
@WyseNynja do you mean ob-watcher? Yes, the on_order_seen callback is in the OrderbookWatch class, and the filter is applied there. Whether that's good or bad, don't know, but then ob-watcher isn't too huge of a deal really.

Issue #634, opened 2016-10-15
Make tumbler idempotent [$1,000 bounty]


eduard6:
Because the tumbler can crash or stall for many reasons being able to restart it and have it pick up where it left off is very useful.

Example:

> $ python tumbler.py -n mysession [options] wallet.json addr1 addr2 addrN
> $ sleep 3600; kill <pid>
> $ python tumbler.py -n mysession [options] wallet.json addr1 addr2 addrN

Options:

- `-n <session name>`: Declare session name.
- `-m <depth>`: If declared, first tumbler run starts from this mixdepth.
- `-D`: If declared, or if no -m is declared, first tumbler run starts from lowest mixdepth that has balance.
- `-y`: Answer yes to all questions. If not declared, user has to approve "tumble plan" on first run. If user does not approve "tumble plan" on first run then session file (see below) should not be created.

Details:

- When session name is declared (`-n`), tumbler will create a `sessions/<session name>.json` file (or `.txt` or anything).

- tumbler will store all options passed on command line including wallet and destination addresses.

- tumbler will store "tumble plan".

- When tumbler is run with session name (`-n`) and `sessions/<session name>.json` file exists, ignore all command line options/wallet/addresses and load from saved options. tumbler should be able to resume using either the exact same command line as the first run or just using `python tumbler -n mysession` and either command will work the same.

- tumbler continues "tumble plan" from where it left off. Obviously when one of `addr1`...`addrN` has been sent to it should not be used again.

- [Optional] tumbler stores PID in session file and uses it to lock session.

- [Optional] session file should be set to 0600/only owner can read.

Bounty:

* I will transfer bounty in advance to any core developer for them to pay to who ever finishes the feature. Please provide address here or by message.
* I will leave it to core developers to decide who deserves the bounty.
* If not finished in 30 days I may cancel bounty and ask for refund, or I may extend deadline.
* $1,000 in BTC when transferred to core developer (BTC may be worth more or less $ when finished of course).
* To qualify for bounty, code should be merged into main JoinMarket project (or be good code that was rejected by core developers for some reason unrelated to code quality)

eduard6:
Previous similar issues: #464 #467

AdamISZ:
Not addressing this proposal right now, but coincidentally, and relevant-ly:

[...]

PR #714, opened 2017-02-27
Session support as described in issue #634

piratelinux:
See the issue here: https://github.com/JoinMarket-Org/joinmarket/issues/634


piratelinux:
Hi, I'm not sure how to fix the conflict with the TestWallet class in test/commontest.py. I added support for passing the password to the wallet constructor, but there seems to be a conflict when this test runs.

joinmarket/support.py
```diff
@@ -398,3 +410,61 @@ def debug_dump_object(obj, skip_fields=None):
...
+    f.write(json.dumps(destaddrs)+',\n')
+    f.write('"tx_list":')
+    f.write(json.dumps(tumbler.tx_list)+',\n')
+    f.write('"next_tx": 0,\n')
+    f.write('"pushed_tx": "",\n')
+    f.write('"pushed_destaddr": ""}\n')
+    f.close()
+
+def update_session_tx_confirmed (sessionname,i_tx):
+    sessionfilename = os.path.join(script_dir,'../sessions/'+sessionname)
```
AdamISZ: Crashed here on a first test run due to sessionname not being set (it's 'None'). The help message says it's optional, so I guess can be fixed various ways, e.g. don't call this fn if sessionname option not chosen, or something similar. It's currently called unconditionally in TumblerThread.confirm_callback()

AdamISZ:
Tested: run on regtest without any adverse maker behaviour. Added sessionname as per minor bug above, test passed OK.

Second test: with malicious makers that fail to respond with small probability (one way of testing restart). Test passed OK, but a few things worthy of considering:

On restarting, might want to give users help about how, for example, restarting with `-n` means the options, wallet and destination addresses are picked up from the json, so any options set on command line are ignored, is that correct (or is it only a subset, haven't checked yet)?

I notice the wallet utxo printout goes to mixdepth 11 by default now, can imagine why that might be, but need to investigate.

Will need to do tests that explicitly cover the case of quit before confirm, so far only doing fast confirm tests with regtest.

Another tricky case (and relatively common) is commitment sourcing failure.

Overall it's looking promising in terms of achieving the intended goal. @chris-belcher @adlai @AlexCato  and anyone else, your input would be greatly appreciated.

AdamISZ:
(There are several more trivial points like : squashing before commits - I'm not super strict about that personally, but it's better not to at least not have 'remove debug' type of commits. there are also really minor formatting things like spaces in function arguments, but I'm again not that bothered, maybe others are.)

AdamISZ:
The changes you've made to wallet should not be there.

I know where it comes from - the 'Wallet' object having to receive a password interactively is a pain. I've been working on improving this kind of thing in the joinmarket-clientserver repo (this and a number of similar things) - but only by removing it from the Wallet class so that the necessary interactivity can be defined by callers (see the joinmarket-qt gui e.g.). But putting a password in command line -> `options` is not acceptable, since it exposes it in things like bash history.

If you're restricting yourself to a command line script (as you are here), then just leave it as interactive even on restart.

joinmarket/support.py
```diff
@@ -398,3 +410,61 @@ def debug_dump_object(obj, skip_fields=None):
...
+    if os.path.isfile(sessionfilename):
+        log.debug("update session tx confirmed "+sessionname+" "+str(i_tx))
+        with open(sessionfilename, "r") as sources:
+            lines = sources.readlines()
+            with open(sessionfilename, "w") as sources:
+                for line in lines:
+                    log.debug('line: '+line+'\n')
+                    if re.match(r'^["]next_tx["][:][ ][0-9]',line):
```
AdamISZ: General comment about this regexp matching approach: it should be far simpler to just json.loads() and then do the updates/changes purely in Python, then write back with json.dumps()

AdamISZ:
There is a bug in handling the case of quit after broadcast but before confirm:

If a user quits before confirmation has occurred, `support.update_session_tx_confirmed` has not been called, and so `options['next_tx']` is not updated. On restart, the tumbler will retry the same transaction.

I noticed this when trying quitting after unconfirm (I set the block generation time to 20 seconds in the regtest setup). I was seeing `that was 1 tx out of 18` (example) more than once after restart.

I haven't thought through how one would fix it in this version of the tumbler, I presume moving the update to the unconfirm_callback although one would have to think through the details.

AdamISZ:
An interesting example of the above scenario; user quit on unconfirm of sweep, then restart and generate a block, get:

    2017-03-08 20:50:48,802 [TumblerThrea] [INFO ]  choosing sweep orders for total_input_value = 0 n=5

which makes sense since the sweep already happened, there are no coins in the mixdepth. However, on restart *again* the problem self-corrects since the confirm callback now has been called and next_tx has been updated.

tumbler.py
```diff
@@ -344,23 +383,37 @@ def run(self):
...
+        pushed_tx = self.taker.options['pushed_tx']
+        if len(pushed_tx)>0 :
+            pushed_destaddr = self.taker.options['pushed_destaddr']
+            self.current_tx = next_tx
+            jm_single().bc_interface.add_tx_notify(
+                btc.deserialize(pushed_tx), self.unconfirm_callback,
+                self.confirm_callback, pushed_destaddr,
+                self.timeout_callback)
+            with self.lockcond:
+                            self.lockcond.wait()
```
AdamISZ: Extraneous extra indent here

AdamISZ:
Tested 'addrask' functionality: appears to be fine.

AdamISZ:
Note the code [here](https://github.com/piratelinux/joinmarket/blob/33046ff4e2978b188899d38da0e634d64ee141c4/tumbler.py#L417-L420). This is a way to set the mixdepth balance that's used to calculate the amount fraction, at the start of each mixdepth, only. So for example, if the amount fraction is 10%, it will be 10% of the amount when starting to process that mixdepth.

If you restart the tumbler, `self.balance_by_mixdepth` will be set to whatever *remains* in the mixdepth on restart. So if you've already processed say 90% of the coins, and the amount_fraction is now 0.05, you will not process 0.05 of the original starting value, but of whatever is left, i.e. 0.005.

This is not the end of the world, but there could be cases where it causes problems, and in any case it fails to create the intended amount distribution. And in case it does cause a problem that problem will be persistent since on restarting a second time the same situation occurs. I addressed this distribution preservation issue [here](https://github.com/AdamISZ/joinmarket-clientserver/blob/master/jmclient/jmclient/schedule.py#L177-L185) although of course it doesn't carry across really.

piratelinux:
Hi thanks for coming through quickly on this. I will review it within the next few hours. I am also willing to do any tests you need, but then someone will eventually have to read the code of the tests. I will probably give you 1% of the bounty for being the escrow/mediator. That's what usually what the mediator I use at Rein gets. You may deserve more, but at least it's something for the principle. I'm not in a super rush, as long as things are smoothly moving forward with a predictable timeframe, then it's fine with me.

piratelinux:
For the case of quitting before confirm, I thought this should handle this:

https://github.com/JoinMarket-Org/joinmarket/pull/714/commits/fcc6b7b6b8e2ddd1d1f1c8b3e5a594328db91d58#diff-e5ebbf3237752f0c78e5834760582020R386

I put a call to new function "update_session_tx_pushed" right before the tx is broadcasted, and this saves the info about the unconfirmed tx to the sessions file for later waiting here: https://github.com/JoinMarket-Org/joinmarket/pull/714/commits/fcc6b7b6b8e2ddd1d1f1c8b3e5a594328db91d58#diff-7dca4cc17829f93a2517b5a8dd3ef018R373

piratelinux:
I now increment the next_tx variable right after the confirm and before it is used in the tx loop.

Also fixed some minor things you mentioned, and one subtle thing: I now put the pushed_tx info before the next_tx in the session file, so that it does not stay marked as unconfirmed right after processing the next_tx line and before removing the unconfirmed tx info.

piratelinux:
For the wallet password, I can see some cases where it can be passed safely on the command line, such as in ram disk os or by disabling bash history. Alternatively, you can pass a password file from which you read the password. I think it would be useful to automate the password passing, such as in a script that keeps restarting the tumbler if it fails (many times I have liquidity or no response problems that get resolved when I restart the tumbler). Maybe there's some command line tools to simulate a user passing an argument to a program, but would be easier if it was built in.

chris-belcher:
Looked at this for the first time and have two comments/requests about moving code around.

Updating sessions is only a tumbler feature and has nothing to do with sendpayment, electrum plugin, bitcoin-qt integration or any other use of the taker class. So I believe the new field `self.sessionname` shouldn't be in taker.py. It looks like you need it to update the session file after a transaction broadcast. That can instead be done within the tumbler.py file by adding it here (https://github.com/JoinMarket-Org/joinmarket/blob/develop/tumbler.py#L150) right after `jm_single().bc_interface.add_tx_notify()`

Also I think the functions `save_session_to_file`, `update_session_balance_by_mixdepth`, `update_session_tx_confirmed` and `update_session_tx_pushed` are tumbler-specific and should be moved to tumbler.py instead of support.py

Apart from that the code looks alright. It was a good idea to use the balance of each mixdepth to figure out where the previous run stalled.

piratelinux:
ok I isolated the changes to the tumbler script and squashed the commits into 1, but did a force push (didn't know how to resolve it otherwise), so now I only have 1 commit in my history.

piratelinux:
Forgot to test that last change. Will do now.

piratelinux:
ok I think it is tested enough now so someone can merge or test further. Thanks.

Issue #667, opened 2016-11-21
AssertionError in yield-generator-basic.py


Empty2k12:
**This error happened after accepting a CJ while two other CJ transaction had not confirmed yet**

> Exception happened during processing of request from ('127.0.0.1', 53802)
> Traceback (most recent call last):
>   File "/usr/lib/python2.7/SocketServer.py", line 290, in _handle_request_noblock
>     self.process_request(request, client_address)
>   File "/usr/lib/python2.7/SocketServer.py", line 318, in process_request
>     self.finish_request(request, client_address)
>   File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
>     self.RequestHandlerClass(request, client_address, self)
>   File "/home/redacted/Desktop/joinmarket-0.2.2/joinmarket/blockchaininterface.py", line 440, in __init__
>     self, request, client_address, base_server)
>   File "/usr/lib/python2.7/SocketServer.py", line 652, in __init__
>     self.handle()
>   File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
>     self.handle_one_request()
>   File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
>     method()
>   File "/home/redacted/Desktop/joinmarket-0.2.2/joinmarket/blockchaininterface.py", line 483, in do_HEAD
>     unconfirmfun(txd, txid)
>   File "/home/redacted/Desktop/joinmarket-0.2.2/joinmarket/maker.py", line 188, in unconfirm_callback
>     self, txid, removed_utxos)
>   File "yield-generator-basic.py", line 114, in on_tx_unconfirmed
>     neworders = self.create_my_orders()
>   File "yield-generator-basic.py", line 65, in create_my_orders
>     assert order['minsize'] <= order['maxsize']
> AssertionError
> ----------------------------------------

**Scrubbed Logs from Today**
[J5CuPVYjcBRMy9R5.scrubbed.txt](https://github.com/JoinMarket-Org/joinmarket/files/604746/J5CuPVYjcBRMy9R5.scrubbed.txt)
[J52pzaaRaXo8XtkF.scrubbed.txt](https://github.com/JoinMarket-Org/joinmarket/files/604747/J52pzaaRaXo8XtkF.scrubbed.txt)

chris-belcher:
Well that shouldn't be an assert, it should just return an empty list [] if maxsize < minsize and the bot wont have any offers until it's transactions confirm.

PR #717, opened 2017-03-02
fix for minsize > maxsize AssertionError in yield-generator-basic.py

peterzen:
fix for #667


chris-belcher:
Thank you.

Would you be able to do the same for yg-pe ?

PR #718, opened 2017-03-02
Yieldgenerator minsize assert fix

peterzen:


chris-belcher:
Looks like a simple change that should work.

It would be great if you could squish these two commits together into one (here's a tutorial in case it helps http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

Seems the test failed because the estimated fee rate in bitcoin has gone above the "sanity" checks. LoL.
That's unrelated to this PR so I suggest we just merge after a one or two more people look at it.

PR #720, opened 2017-03-03
fix for minsize > maxsize AssertionError in yield generators

peterzen:
Combined change


chris-belcher:
I've reviewed this and I think it's good.

AdamISZ:
Agreed, merging

PR #721, opened 2017-03-09
fixed absurd miner fee check

chris-belcher:
What we previously thought was an absurd miner fee is now the reality.

Bumped up this absurd value to be an order of magnitude higher than the current market fee rate.

Also fixed the test which had hard-coded absurd values, it now uses the absurd values from the joinmarket.cfg file. This is breaking all automated test runs right now. See https://travis-ci.org/JoinMarket-Org/joinmarket/builds/207548542#L1336

### DIFF
--- a/joinmarket/configure.py
+++ b/joinmarket/configure.py
@@ -147,7 +147,7 @@ tx_fees = 3
 # check limit on the satoshis-per-kB to be paid. This limit
 # is also applied to users using Core, even though Core has its
 # own sanity check limit, which is currently 1,000,000 satoshis.
-absurd_fee_per_kb = 150000
+absurd_fee_per_kb = 2000000
 # the range of confirmations passed to the `listunspent` bitcoind RPC call
 # 1st value is the inclusive minimum, defaults to one confirmation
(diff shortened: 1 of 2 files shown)


chris-belcher:
Looks like unrelated code test_p2p_broadcast made the test fail :\

AdamISZ:
ACK. If you want to merge and leave the p2p broadcast test fix till later, no problem with me.

AlexCato:
ACK in general, but an improvement suggestion:
I'm in agreement that this needs to be bumped up a bit. But with the new "absurd", a tumbler run could easily costs tens of dollars just in miner fees. Plus: the fees are usually not even mentioned when the user confirms, even though they make up the majority of the fees the taker has to pay... I'd love to see at least a rough estimate when the user confirms his intended joins.

chris-belcher:
Actually even today a tumbler run could cost that much. Because right now the taker will pay for any of the maker's UTXOs to be mined. So a maker could send 100 UTXOs and the taker is coded to pay for all of them. The total miner fee is potentially unbounded. :( I've already written about this somewhere else that the miner fee incentives in joinmarket are all screwed up right now.

PR #741, opened 2017-06-21
Message signing with Joinmarket wallet addresses

AdamISZ:
Signature is compatible with Core (although not much tested yet).
User passes the HD path as an option (-H).
Verification not yet exposed (some complexity in secp256k1 api).


AdamISZ:
Going to merge, don't see a likely issue here. Note verification is not implemented, need to investigate something like `secp256k1.ecdsa_recoverable_deserialize`, get a valid sig object etc.

Issue #660, opened 2016-11-19
StackTrace when running yield generator


RHavar:
```
2016-11-19 05:00:04,417 [MCThread    ] [INFO ]  <<pubmsg on CgAn: nick=J5iweygvSSV378dO message=!orderbook
Exception in thread MCThread:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/root/joinmarket/joinmarket/message_channel.py", line 31, in run
    self.mc.run()
  File "/root/joinmarket/joinmarket/irc.py", line 448, in run
    self.__handle_line(line)
  File "/root/joinmarket/joinmarket/irc.py", line 326, in __handle_line
    self.__handle_privmsg(_chunks[0], _chunks[2], get_irc_text(line))
  File "/root/joinmarket/joinmarket/irc.py", line 278, in __handle_privmsg
    self.on_pubmsg(nick, message)
  File "/root/joinmarket/joinmarket/message_channel.py", line 889, in on_pubmsg
    self.on_orderbook_requested(nick, self)
  File "/root/joinmarket/joinmarket/message_channel.py", line 556, in on_orderbook_requested_trigger
    self.on_orderbook_requested(nick, mc)
  File "/root/joinmarket/joinmarket/maker.py", line 277, in on_orderbook_requested
    self.msgchan.announce_orders(self.orderlist, nick, mc)
  File "/root/joinmarket/joinmarket/message_channel.py", line 285, in announce_orders
    cmd = orderlist[0]['ordertype']
IndexError: list index out of range
```

AdamISZ:
Thanks for the report. Which yield generator script are you running? Also please confirm whether you are running the latest version (0.2.2 = current master branch).

Your bot's orderlist (we should really call it "offerlist" now, but same thing) was empty, this shouldn't ever happen. I haven't seen it before. The stack trace is useful, thanks, but unfortunately it can't show directly why the orderlist is empty.

First thought (apart from checking which script is involved) is it might be to do with a too-small amount of coins available. On startup this results in a "Not enough funds" message and quit. If, during running, you participated in a join which was a large enough percentage of your total available, and all the other mixdepths are empty perhaps, then while that transaction was in an unconfirmed state, those coins are not available, and the amount left might be too small for the bot to allow an offer/order to be created; this *might* cause such a condition, I'd have to look into it.

Anyone else have another theory?

[...]

Issue #639, opened 2016-10-19
Update and/or check scrub-log.py


AdamISZ:
The scrub-log.py script appears to work OK with the new version (0.2.*), but a more careful audit to ensure that all relevant private data is scrubbed in all cases would be useful.

Another issue with scrub-log is that all timestamps are removed for privacy, it would be very helpful (e.g. in identifying network issues) to have the time deltas from the start.

AdamISZ:
See additional details here: https://github.com/JoinMarket-Org/joinmarket/pull/744#issuecomment-317417854

Issue #523, opened 2016-05-11
yield-generator-basic Crash


Onefox:
Here is the debug log from last night.
Before that the generator worked fine for several hours and joins after a restart its working again.

http://pastebin.com/WLuLs2BM

chris-belcher:
could you use this file on the entire log and then post it https://github.com/JoinMarket-Org/joinmarket/blob/master/logs/scrub-log.py

Onefox:
Sure here you go:
http://pastebin.com/kDJTC9zJ

chris-belcher:
Thanks for the report.

Looks like what happened is one of the UTXOs that yield-gen thought it had was actually not real. It went to recalculate which UTXOs it should use to fulfill the order but found there was now not enough bitcoins. This wasn't expected so it crashed.

> TIMESTAMP_0000000000420 [MainThread  ] [DEBUG]  mix depths that have enough = []

The way to fix this is somewhere around this line https://github.com/JoinMarket-Org/joinmarket/blob/087921b50beb93faa8e9125611b6cde7d12d4a16/yield-generator-basic.py#L107

check if len(filtered_mix_balance) == 0 and if it is then return None, None, None.

Tagging this as Easy / Quick to fix because it is.

Onefox:
Just one more over here:
http://pastebin.com/SQAvvQeC

Also the oder is not being update in the oderbook after a trade could that be?

AdamISZ:
This has been fixed in develop: https://github.com/JoinMarket-Org/joinmarket/commit/f9fd64f0d2d1ac8d16bf26425607b75443a7208c

PR #744, opened 2017-07-24
fix scrub-log.py bug: also scrub unicode TXIDs

adlai:
The scrub-log.py script failed to delete TXIDs that got dribbled into the log as unicode. Why this happens is not yet fully clear, but it appears to only occur with the `gettxout` RPC call. Problem was spotted by @AdamISZ while [helping a redditor](https://www.reddit.com/r/joinmarket/comments/6oyx3n/strange_maker_logs_not_sure_how_to_explain_at/).

### DIFF
--- a/logs/scrub-log.py
+++ b/logs/scrub-log.py
@@ -50,7 +50,7 @@ patterns = [
     ('(\W)([0-9a-fA-F]{64}):(\d+)(\W)',
     lambda m: m.group(1) + 'TXID_INDEX_%056d' + m.group(4),
     'txid:index'),
-    ('(gettxout \\[\')([0-9a-fA-F]{64}\', \d+)(, (True|False)\\])',
+    ('(gettxout \\[u?\')([0-9a-fA-F]{64}\', \d+)(, (True|False)\\])',
     lambda m: m.group(1) + 'TXID_%059d\', IDX_%03d' +  m.group(3),
     'gettxout'),
     ('(\'value\': )(\d+)(\\})',


AdamISZ:
tested ACK, in that it's a strict improvement on the existing. Thanks for fixing that.

BUT: this is not the full story - the first existing privacy loss is in 'mix depths that have enough' - which still reveals balance.
Second is that inter-participant messages don't seem to be scrubbed, so just for one example, the fill amount is there, also the fees earned, also I think the authorising pubkey (which can be converted to an address) is leaked.

Also, I think it would make sense to have a safer approach based on catching all 32 or 33 byte hex and replace, rather than a template approach. The integer amounts will have to be thought about carefully to do this 'catch all' thing. (To state the obvious, the template approach's brittleness also crops up in any code updates that change any log messages).

Last point is something that I noted in previous attempts to debug people's problems - I would prefer to keep the timestamps, but with a delta taken off - so you can see how long certain operations took.

If we can't sort of "unconditionally" solve the problem of leaks, I think it'd be better to delete the file, but then again, a lot of people use release rather than develop branch, so the horse has kind of bolted ... so probably forget that. I mean, maybe just tell people not to publish it until we're sure it's safe. To be fair, I'm not sure we ever did "tell people to publish it"!

0xB44EFD8751077F97:
>To be fair, I'm not sure we ever did "tell people to publish it"!

From https://github.com/JoinMarket-Org/joinmarket/issues/523#issuecomment-218483831:

>chris-belcher: 
>could you use this file on the entire log and then post it https://github.com/JoinMarket-Org/joinmarket/blob/master/logs/scrub-log.py

From https://github.com/JoinMarket-Org/joinmarket/issues/660#issuecomment-261712205:

>chris-belcher:
>Could you use the scrub-log.py script found in logs/ (and also https://github.com/JoinMarket-Org/joinmarket/blob/master/logs/scrub-log.py) to remove private information from your log file and then post it here?

@AdamISZ You did voice concern/warning here: https://github.com/JoinMarket-Org/joinmarket/issues/639#issue-183938378

AdamISZ:
@0xB44EFD8751077F97 Yep, I was just plain wrong about that. Sorry.

AdamISZ:
In the absence of further updates, merging as an unconditional improvement, #639 remains open and I've labelled it bug. We should also probably delete it until it's properly fixed as above.

PR #748, opened 2017-08-08
Add executable bit and hashbangs for easier usage

mb300sd:


AdamISZ:
Thanks, that looks helpful.

Worth a thought: use python2 alias? this code is not yet ported to Python 3.

mb300sd:
Changed to python2.

Issue #728, opened 2017-03-30
Switch secp256k1 dependency


ofek:
Consider switching from [https://github.com/ludbb/secp256k1-py](https://github.com/ludbb/secp256k1-py) to [https://github.com/ofek/coincurve](https://github.com/ofek/coincurve) as the former is no longer maintained.

New features include:

- Support for Windows (Python 3 only though, unlike Linux/macOS)
- Linux, macOS, and Windows all have binary packages for both 64 and 32-bit architectures
- A fix to remove CFFI warnings
- A global context is used by default, drastically increasing performance
- Each release uses newest version of [libsecp256k1](https://github.com/bitcoin-core/secp256k1)
- Implements a fix for [https://bugs.python.org/issue28150](https://bugs.python.org/issue28150)  to support Python 3.6+ on macOS
- Linux & macOS wheels now use GMP
- Endomorphism optimization is now enabled

When you do switch, **use coincurve version 2.1.0 specifically!** Anything after that will have a modified (cleaner :) API and only support Python 3.

AdamISZ:
@ofek thanks.

If someone wants to pursue this: would suggest making a branch to make this as an optional alternative for building initially, then people can test it out easily.

mb300sd:
+1 

The old binding no longer runs with the latest build if libsecp256k1.

I tried and failed at converting, api changes too much for my minimal understanding of EC.

AdamISZ:
@mb300sd thanks for the heads up. I will take a look today.

Is it libsecp256k1 rather than just secp256k1-py that has changed? (I guess so from the way you worded it). If so yeah it'll take some work and be quite hard to deal with since they don't issue releases. If it's secp256k1-py then we can just fix a particular version to install, for now.

ofek:
@AdamISZ Yes libsecp256k1 has received many improvements since the last release of secp256k1-py, 6 performance enhancements just in the past few months https://github.com/ofek/coincurve#500

Have you taken a look at https://github.com/ofek/coincurve yet?

mb300sd:
libsecp256k1 removed some functions related to schnorr signatures a while ago, so it fails with a cffi not found error when using the latest versions.

It looks simple enough to convert to coincurve, (for someone that understands the EC math), I got about half way though in an hour and couldn't tell what needed to be PrivateKey.multiply vs PublicKey.multiply anymore and gave up.

AdamISZ:
@mb300sd 

[...]

PR #749, opened 2017-08-09
change from secp256k1-py to coincurve

mb300sd:
Switching from the no longer maintained secp256k1-py library to coincurve.

Removes the hacks for Windows compatibility and enable donations on Windows. (untested)

#728

### DIFF
--- a/bitcoin/podle.py
+++ b/bitcoin/podle.py
@@ -1,15 +1,13 @@
 #Proof Of Discrete Logarithm Equivalence
 #For algorithm steps, see https://gist.github.com/AdamISZ/9cbba5e9408d23813ca8
-import secp256k1
+import coincurve
 import os
 import hashlib
 import json
 from py2specials import *
 from py3specials import *
-from secp256k1_main import ctx
 PODLE_COMMIT_FILE = None
 N = 115792089237316195423570985008687907852837564279074904382605163141518161494337
-dummy_pub = secp256k1.PublicKey(ctx=ctx)
 
 def set_commitment_file(file_loc):
     global PODLE_COMMIT_FILE
@@ -18,13 +16,6 @@ def set_commitment_file(file_loc):
 def get_commitment_file():
     return PODLE_COMMIT_FILE
 
-def tweak_mul(point, scalar):
-    """Temporary hack because Windows binding had a bug in tweak_mul.
-    Can be removed when Windows binding is updated.
-    """
-    return secp256k1._tweak_public(point,
-                                   secp256k1.lib.secp256k1_ec_pubkey_tweak_mul,
-                                   scalar)
 class PoDLEError(Exception):
     pass
 
@@ -43,7 +34,7 @@ class PoDLE(object):
         if not priv:
             if P:
                 #Construct a pubkey from raw hex
-                self.P = secp256k1.PublicKey(safe_from_hex(P), raw=True, ctx=ctx)
+                self.P = coincurve.PublicKey(safe_from_hex(P))
             else:
                 self.P = None
         else:
@@ -52,10 +43,10 @@ class PoDLE(object):
             #any other formatting abnormality will just throw in PrivateKey
             if len(priv)==66 and priv[-2:]=='01':
                 priv = priv[:-2]
-            self.priv = secp256k1.PrivateKey(safe_from_hex(priv), ctx=ctx)
-            self.P = self.priv.pubkey
+            self.priv = coincurve.PrivateKey(safe_from_hex(priv))
+            self.P = self.priv.public_key
         if P2:
-            self.P2 = secp256k1.PublicKey(safe_from_hex(P2), raw=True, ctx=ctx)
+            self.P2 = coincurve.PublicKey(safe_from_hex(P2))
         else:
             self.P2 = None
         #These sig values should be passed in hex.
@@ -80,9 +71,9 @@ class PoDLE(object):
         """
         if not self.P2:
             raise PoDLEError("Cannot construct commitment, no P2 available")
-        if not isinstance(self.P2, secp256k1.PublicKey):
+        if not isinstance(self.P2, coincurve.PublicKey):
             raise PoDLEError("Cannot construct commitment, P2 is not a pubkey")
-        self.commitment = hashlib.sha256(self.P2.serialize()).digest()
+        self.commitment = hashlib.sha256(self.P2.format()).digest()
         return safe_hexlify(self.commitment)
 
     def generate_podle(self, index=0):
@@ -117,14 +108,14 @@ class PoDLE(object):
         #TODO nonce could be rfc6979?
         k = os.urandom(32)
         J = getNUMS(index)
-        KG = secp256k1.PrivateKey(k, ctx=ctx).pubkey
-        KJ = tweak_mul(J, k)
+        KG = coincurve.PrivateKey(k).public_key
+        KJ = J.multiply(k)
         self.P2 = getP2(self.priv, J)
         self.get_commitment()
         self.e = hashlib.sha256(''.join(
-            [x.serialize() for x in [KG, KJ, self.P, self.P2]])).digest()
+            [x.format() for x in [KG, KJ, self.P, self.P2]])).digest()
         k_int = decode(k, 256)
-        priv_int = decode(self.priv.private_key, 256)
+        priv_int = decode(self.priv.secret, 256)
         e_int = decode(self.e, 256)
         sig_int = (k_int + priv_int*e_int) % N
         self.s = encode(sig_int, 256, minlen=32)
@@ -139,8 +130,8 @@ class PoDLE(object):
         if not self.commitment:
             self.get_commitment()
         Phex, P2hex, shex, ehex, commit = [
-            safe_hexlify(x) for x in [self.P.serialize(),
-                                      self.P2.serialize(),
+            safe_hexlify(x) for x in [self.P.format(),
+                                      self.P2.format(),
                                       self.s, self.e, self.commitment]]
         return {'used': str(self.used), 'utxo': self.u, 'P': Phex, 'P2': P2hex,
                 'commit': commit, 'sig': shex, 'e': ehex}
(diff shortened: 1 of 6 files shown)


ofek:
@mb300sd I'm available for any assistance you may need in regards to Coincurve or EC in general. Thanks for moving this forward!

bitcoin/podle.py
```diff
@@ -9,7 +10,7 @@
 from secp256k1_main import ctx
 PODLE_COMMIT_FILE = None
 N = 115792089237316195423570985008687907852837564279074904382605163141518161494337
-dummy_pub = secp256k1.PublicKey(ctx=ctx)
+dummy_pub = coincurve.PublicKey(None)
```
ofek: Should be `dummy_pub = coincurve.PublicKey(os.urandom(32))`

mb300sd: That completely broke it, it was building with None :/

```
  File "/home/travis/build/JoinMarket-Org/joinmarket/bitcoin/podle.py", line 13, in <module>
    dummy_pub = coincurve.PublicKey(os.urandom(32))
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/coincurve/keys.py", line 194, in __init__
    raise ValueError('The public key could not be parsed or is '
ValueError: The public key could not be parsed or is invalid.
```

ofek: So very sorry, should be `coincurve.PrivateKey().public_key`

ofek: `coincurve.PublicKey(None)` will succeed, but any method calls with fail. This is because anything other than `bytes` is assumed to be a valid `ffi.new('secp256k1_pubkey *')`. This results in a speed improvement for derivation in PrivateKey.

mb300sd: It seems like dummy_pub is only ever used for combine() which never accesses self.public_key. Could be useful to have combine as a classmethod if it doesn't break anything?

mb300sd:
@ofek I made a PR on coincurve, should be the last piece of the puzzle to make this work. JM needs the custom nonce parameters on sign() that aren't currently exposed.

ofek:
@mb300sd Thanks! Also, https://github.com/ofek/coincurve#510

mb300sd:
Running this on my yield generator now. So far so good, will play with it some more tomorrow. Someone should test on Windows.

ofek:
@mb300sd @AdamISZ You may want to pin to 6.0.0. https://github.com/ofek/coincurve#600

Up to you though!

ofek:
@AdamISZ Does coverage have to increase to be able to merge?

AdamISZ:
@ofek No, the tests are not deterministic at the moment (and need work of course).

But let's get it working in a fork+branch first.

ofek:
@AdamISZ This appears to be ready (@mb300sd is that right?). Could you please review?

AdamISZ:
@ofek Will do, tomorrow. thanks.

bitcoin/secp256k1_main.py
```diff
@@ -239,23 +236,17 @@ def multiply(s, pub, usehex, rawpub=True):
...
     ('raw' options passed in)
     '''
-    newpub = secp256k1.PublicKey(pub, raw=rawpub, ctx=ctx)
-    #see note to "tweak_mul" function in podle.py
-    res = secp256k1._tweak_public(newpub,
-                                   secp256k1.lib.secp256k1_ec_pubkey_tweak_mul,
-                                   s)
-    return res.serialize()
+    res = coincurve.PublicKey(pub)
+    res.multiply(s, update=True)
```
AdamISZ: I think ...:

using update=True is going to multiply the point being called, and that's sometimes not what's wanted.

For example in the PoDLE construction, in the function `podle.PoDLE.generate_podle()` we have lines like:

    KJ = tweak_mul(J, k)

This will I believe update J to k*J. As it happens I don't think it'll matter in that specific function. `bitcoin.multiply` is also called in `configure.donation_address`, and at a quick glance, it looks like that does get screwed up (because `reusable_donation_pubkey` is referred to again after the update).

bitcoin/secp256k1_main.py
```diff
@@ -300,9 +291,9 @@ def ecdsa_raw_sign(msg,
         raise Exception("Invalid hash input to ECDSA raw sign.")
     if rawpriv:
         compressed, p = read_privkey(priv)
-        newpriv = secp256k1.PrivateKey(p, raw=True, ctx=ctx)
+        newpriv = coincurve.PrivateKey(p)
     else:
-        newpriv = secp256k1.PrivateKey(priv, raw=False, ctx=ctx)
+        newpriv = coincurve.PrivateKey.from_der(priv)
```
AdamISZ: We don't yet have a non-raw-privatekey usecase, this seems like a sensible choice (`from_der`) out of the given options.

bitcoin/secp256k1_main.py
```diff
@@ -311,16 +302,18 @@ def ecdsa_raw_sign(msg,
         ndata = ffi.new("char [32]", usenonce)
         usenonce = (nf, ndata)
     if formsg:
-        sig = newpriv.ecdsa_sign_recoverable(msg, raw=rawmsg)
-        s, rid = newpriv.ecdsa_recoverable_serialize(sig)
+        if rawmsg:
+            sig = newpriv.sign_recoverable(msg, hasher=None)
```
AdamISZ: I guess we don't really need this clause, because `formsg` is supposed to be a flag for message signing, and that will always be hashed.

bitcoin/secp256k1_main.py
```diff
@@ -311,16 +302,18 @@ def ecdsa_raw_sign(msg,
...
+        else:
+            sig = newpriv.sign_recoverable(msg)
+        s, rid = sig[:64], bytes_to_int(sig[64:])
```
AdamISZ: As well as other testing, please do test whether generated signatures still verify against a Core node's verification (this was added in develop recently), thanks.

AdamISZ:
OK I've done some review; first @mb300sd and @ofek thanks for this work, it's appreciated.

Perhaps add to the PR the edits to the readme/installation instructions (you can hotlink to the `pip install coincurve` instructions in coincurve's repo).

Re: testing. Obviously passing the build test is the first hurdle, and that's great. There is another kind of test that is very important though: you'll need to test counterparties-on-other-library:

For example, run `test/ygrunner.py` (there are some brief testing instructions for setting up regtest+miniircd+pytest on the [wiki page](https://github.com/JoinMarket-Org/joinmarket/wiki/Testing)) with some maker bots, *without* this PR, then run say a sendpayment or even better tumbler, talking to those bots, using this new code. I'm sure you can see what I'm saying here. It has to be both ways and it has to be a large-ish set of transactions before we can be confident that it's all OK. While incompatibilities aren't very likely (and funds loss type errors even less so), we have to be thorough on this.

bitcoin/secp256k1_main.py
ofek: Both versions here are functionally equivalent. The difference is the new version doesn't require a temporary variable.

bitcoin/secp256k1_main.py
ofek: Yes `hasher=None` assumes the message has already been hashed.

bitcoin/secp256k1_main.py
ofek: Can you please explain what exactly @mb300sd should do here?

bitcoin/secp256k1_main.py
AdamISZ: Yes you're right (in fact I think that second variable wasn't needed in the old version, but no matter). I'll have to double check what that donation address code is doing (I haven't looked at it in over a year), but that's independent of the PR.

bitcoin/secp256k1_main.py
AdamISZ: On reflection don't worry. The issue is: I added the message signing feature on user request about a month ago, but due to pressure of time haven't added it to the test suite (it's hardly critical anyway). Just wanted to sanity check that the algo still works correctly here (don't forget I am not coming from the position of knowing every line of the coincurve code). I or someone else can get round to checking it later.

bitcoin/secp256k1_main.py
mb300sd: I believe update only changes the PublicKey object 'res', and doesn't effect 'pub'. Changed it anyway to remove the confusion.

bitcoin/secp256k1_main.py
mb300sd: Tested, fixed, and tested :)

mb300sd:
Made some changes based on @AdamISZ 's comments. 

Will try to get the regtest working later on. For now, my yield generator is running using this version and I've sent a few payments. (I mainly did it because I have the latest libsecp256k1 installed on my system that broke the old binding) . So far so good.

AdamISZ:
@mb300sd re:

>  I have the latest libsecp256k1 installed on my system that broke the old binding

I took great pains to test this out, and again, I could not reproduce; are you sure you were running the latest version of the secp256k1-py binding? I'm guessing you did not do as I did in my last comment in #728 (`pip install --no-binary` etc.)? The lowest time waste thing here is if you can perhaps write down *exactly* how you installed both the lib and the binding on #728 ? including e.g. if you passed any flags to configure for the former?

This is not just academic; I want to know in case other people have similar problems. So far they don't; I think they are just using the binaries from the wheel. If in the end it's not resolved, OK, but it's not like switching to this new library is going to happen overnight. We've been using the former for a year and a half at least now.

(Also I don't know if you saw my latest reply here https://github.com/JoinMarket-Org/joinmarket/pull/749#discussion_r132772803 because it's in an "outdated diff" section).

mb300sd:
@AdamISZ 

I'm using the latest git of libsecp256k1.
```
git pull
./autogen.sh
./configure --enable-module-ecdh --enable-module-recovery --enable-experimental --enable-endomorphism --with-bignum=gmp
make -j32
make install
```

I'm running debian jessie, the no binary option does not seem to be available.
no such option: --no-binary

I'm using `pip install secp256k1`

I did see the comment, I removed the update=True use in the latest commit.

AdamISZ:
Ah OK great, thanks for getting back to me on that.

I didn't use any of the configure flags, that's a delta but I'm not sure the details there.

(i.e. I just did precisely [this](https://github.com/bitcoin-core/secp256k1#build-steps) including install).

What I discovered from my experiment, and what you said, is: it's pretty likely that nobody is using a system-installed libsecp256k1 ... because if you tried to follow the instructions on the secp256k1-py repo, it doesn't work (`pip install --no-binary secp256k1` gives a rather bizarre syntax error), plus you have to get it to dynamically link the lib you just built. And in your case having `--no-binary` not even be there must be due to a different pip version I guess (it's def. not there even after a `pip install --upgrade pip`?).

Anyway, I know it's a side issue in the context of this PR, but this is a good extra point of reference if it comes up again.

> I did see the comment, I removed the update=True use in the latest commit.

Ah gotcha I didn't yet review the latest updates to the commit.

Don't be surprised if this doesn't get merged for a while. I'll ping anyone else I can find who might be willing to look at it. Tests like I mentioned above should really be done although I don't see any urgency; live testing is OK if you're sufficiently confident and definitely counts, the only thing it lacks is tx volume.

ofek:
Is this closer to being merged?

ofek:
@AdamISZ Is this ready?

ofek:
@adlai @chris-belcher Can this be merged? Many have switched to Coincurve now https://github.com/ofek/coincurve#users

adlai:
@ofek I have no objection if @AdamISZ wants to merge this; however, I'd like to review it myself, including at least a cursory review of coincurve, so it doesn't have my official `ACK` for _at least_ "two weeks".

AdamISZ:
It's on the [TODO list](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/TODO.md) in the active repo. See the comments there (at the end of the doc).

adlai:
I'm gonna go ahead and give this a stACK (as in, slightly). Will merge after a bit more jiggling about, although I'm not gonna test this on Windows, nor am I going to exhaust the cross-compatibility scenarios suggested by @AdamISZ.

adlai:
> @mb300sd @AdamISZ You may want to pin to 6.0.0. https://github.com/ofek/coincurve#600

Having read the changelog for that release, I **_strongly object_** to that specific pinning. If you insist that a later version should be pinned, please do, but software patents can go fuck themselves (at least, as far as this repository is concerned).

ofek:
I would recommend the most recent `9.0.0` so mac users have an easier time!

